PR #12, opened 2021-08-02
update component library

mcrascal:
- [x] expands the component library, thanks @hughess 
- [x] component library is now an output of `svelte-kit package`


hughess:
🙌

PR #35, opened 2021-08-06
Fix histogram

### LINKED ISSUE #17: Histogram component cannot accept null values
By default, the chart should treat nulls as zero to match up with the behaviour of the other components in the library

### LINKED ISSUE #25: Histogram x axis labels are not formatted
Caused by the dataset that is outputted from the histogram component dropping the original column names. Change this output dataset to use the original x column name, which will pull through the format tag of the column.

### LINKED ISSUE #27: Wrong column name referenced in one of the histogram calculations
Quick fix - change "value" to "frequency"

hughess:
Fixes #17  
Fixes #25
Fixes #27

### DIFF
--- a/packages/components/viz/Column.svelte
+++ b/packages/components/viz/Column.svelte
@@ -55,21 +55,6 @@
     $: isBandwidth = typeof $xScale.bandwidth === 'function';
     $: isHist = binMax != null;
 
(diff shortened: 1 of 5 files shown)


mcrascal:
Branch created before docs site added. Ignoring vercel issue.

PR #70, opened 2021-08-14
Fix xAxisTitle text anchor by adding separate text tag

### LINKED ISSUE #65: Text anchor causes label overlap on reverse axis charts
Text anchor is set to middle, which causes overlapping labels when the units prop is a long string. Change text anchor to "start" to position the start of the string directly on the tick mark.

CleanShot 2021-08-14 at 10 29 06

hughess:
Fixes #65  

Adds a separate text tag for the xAxisTitle of reverse axis charts (e.g., the "units" prop of BarChart).

New text tag is positioned to the right of the tick label on the furthest right. Padding between tick label and new text tag is based on a rough string length approximation.

### DIFF
--- a/packages/components/viz/XAxis.svelte
+++ b/packages/components/viz/XAxis.svelte
@@ -1,7 +1,7 @@
 <script>
 	import { getContext } from 'svelte';
   import formatAxisLabel from '../modules/formatAxisLabel.js'
-	const { width, height, xScale, xDomain, yRange, padding, yDomain} = getContext('LayerCake');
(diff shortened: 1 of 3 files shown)


hughess:
**Before:**
CleanShot 2021-08-14 at 12 44 36

**After:**
CleanShot 2021-08-14 at 13 08 08

hughess:
This PR shows as closed, but it was successfully merged into the develop branch in this commit: ca4dce6c923d659fd325d9e0eea4aee36d3d7b9b

PR #120, opened 2021-09-02
Added another root cause to common error

leleu:
"offset is longer than source length" can also occur if the query identifier in the markdown block starts with a number.


hughess:
Thanks @leleu! 🎉

Issue #54, opened 2021-08-09
Add error handling to Value component


hughess:
Apply same checkInputs logic as in other components

PR #130, opened 2021-09-10
Add error handling and text placeholder option

hughess:
CleanShot 2021-09-10 at 10 26 32

CleanShot 2021-09-10 at 10 27 24


hughess:
Closes #54

PR #191, opened 2022-03-05
modify preprocess imports from example proj

hughess:


mcrascal:
'dev mode' is used in a few places -- I've renamed for clarity. 
Also, this change means there's no benefit to packaging the components when booting the dev server, so I've removed that.

PR #226, opened 2022-04-20
change location of sqlite db path

hughess:


edgarrmondragon:
@hughess does this mean absolute paths [as they're documented](https://docs.evidence.dev/getting-started/connect-data-warehouse#sqlite) will stop working and SQLite files are expected to be inside a specific directory within the project?

hughess:
@edgarrmondragon that's what we're testing at the moment, but this is only going into a pre-release branch, so it won't take effect in any projects yet.

We're testing having the SQLite files expected in the root of your Evidence project. If you had a script dropping in a db file, it should work by setting the path to be relative to the root of the project rather than absolute.

Would that work for your use case or create issues for you? I can also show you what we're working on with this if you have time and want to do a screenshare

edgarrmondragon:
> Would that work for your use case or create issues for you? I can also show you what we're working on with this if you have time and want to do a screenshare

@hughess  Yeah that works, I just wanted to confirm. I saw you're using it as part of a new tutorial so it makes perfect sense 😄. Thanks!

PR #234, opened 2022-04-27
Tutorial Update

archiewood:
Starting a PR to update the tutorial, will add from here.

**Aims:**
- Replace the Austin-311 dataset, which has become less usable as BigQuery has limited access to it.
- Have a tutorial to support newer features of Evidence.


archiewood:
Changes:
- Updated the tutorial to use the Needful Things dataset, which is a SQLite db, with smooth connection process.
- Reordered some of the content so that users deploy project before hitting advanced content.
- Added material which demonstrates more of Evidence's charting capability
- More 'Story driven' approach to the overall tutorial - should make motivation clearer, and content more interesting

PR #260, opened 2022-05-11
Persistent query show/hide states

archiewood:
As a user in dev mode, it's frustrating that query results collapse every time the page reloads.

Desired behaviour: 
- The query results are kept as shown / hidden based on what the state was before reload

Rough approach
- Store the state of each query in local storage, and read from there on reload

Questions
- This seems like a weird implementation of storage - I feel like you typically put svelte store variables in store.js?
- But since (I think) you want to control each query independently, you'd perhaps need to independently make variables for each query - and you don't know how many there are before the queries are compiled. So I put them in the component instead.

Next
- It would seem logical to also include the show/hide state of the query text itself, once we've worked out the best approach for this

### DIFF
--- a/sites/example-project/src/components/ui/QueryViewer.svelte
+++ b/sites/example-project/src/components/ui/QueryViewer.svelte
@@ -6,6 +6,8 @@
   import Prism from "./QueryViewerSupport/Prismjs.svelte";
   import {showQueries} from './stores.js'
   import CompilerToggle from './QueryViewerSupport/CompilerToggle.svelte';
+  import { writable } from 'svelte/store';
+  import { browser } from '$app/env';
 
   export let queryID; 
   export let pageQueries
@@ -30,7 +32,10 @@
   let error = queryResult[0]?.error_object?.error
   let nRecords = null
   let nProperties = null
-  let showResults = false
+  // Create a copy of the showResults variable in the local storage, for each query. Access this to determine state of each query dropdown.
+  let showResults = writable(browser && (localStorage.getItem('showResults_'.concat(queryID))==='true'  || false));
+  showResults.subscribe((value) => browser && (localStorage.setItem('showResults_'.concat(queryID),(value))));
+  
 
   if(!error){
     nRecords = queryResult.length
@@ -41,7 +46,7 @@
   
   const toggleResults = function() {
     if(!error && nRecords > 0){
-      showResults = !showResults
+      $showResults = !$showResults
     }
   }
 
@@ -72,7 +77,7 @@
           {/if}
       </div>
       <!-- Status -->
-      <div class = {"status-bar" + (error ? " error": " success") + (showResults ? " open": " closed")} on:click={toggleResults}>  
+      <div class = {"status-bar" + (error ? " error": " success") + ($showResults ? " open": " closed")} on:click={toggleResults}>  
         <span> 
           {#if error}
             {#if dev && error.message === "Missing database credentials"}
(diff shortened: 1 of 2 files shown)


hughess:
@ud3sh, is there any reason we should avoid using local storage as a solution for persistent settings like these?

archiewood:
Yeah @ud3sh I think that's probably desirable behaviour.

If you had two people working on / looking at a project they might want to have different stuff collapsed / expanded

ud3sh:
> Yeah @ud3sh I think that's probably desirable behaviour.
> 
> If you had two people working on / looking at a project they might want to have different stuff collapsed / expanded

yeah, that makes sense.
What I meant more was if Archie looks at the deployed site on his laptop vs his iPad things would look different.  Probably not a big issue in practice.

hughess:
That makes sense. 

@archiewood I think your approach here is a good way to manage that toggle. If we need to centralize the logic later we can think about it, but having the logic for the show/hide toggle in the component it's modifying seems like the right fit.

I'd say you're good to go to add this to the query text show/hide as well!

archiewood:
Updated to include showhide for:
- results table
- written SQL
- "master" switch - all queries on page

Question about the master switch:
Do we want this to change the state of all queries throughout the project, or just the page the user is on?

Currently this will change throughout the project. Not sure what is best?

mcrascal:
I like it persisted across pages. "I'm showing my work to someone non-technical, I hide queries, I'm not hitting that button over and over." And vice versa.

hughess:
I like across pages as well. Should it also do that when the project is deployed? 

For the "master" switch, as a user I think I'd want the same behaviour in a deployed site - whereas for the sql query text and results table being open/closed, I think I'd be okay with the page refreshing and the query viewers all being closed again.

archiewood:
@hughess When you say -"you'd be okay with", do you mean you'd want?

Currently I believe this will not happen. All the queries would remain closed.

I can alter the behaviour in deployed (I assume by using the 'dev' flag to determine which mode the site is in)

archiewood:
@hughess When you say -"you'd be okay with", do you mean you'd want?

Currently I believe this will not happen. All the queries would remain as the last time they were viewed on the device.

I can alter the behaviour in deployed (I assume by using the 'dev' flag to determine which mode the site is in)

mcrascal:
I think persisting "open" or "closed" between pages makes sense in both dev and prod. Whichever way you go on persistence between pages, I think it'd be preferable for that behaviour to be the same in dev and prod.

hughess:
I think the way it works now will be great. I don't have a strong opinion about whether the settings should persist in a deployed project, but we can always change the deployed behaviour in the future if we need to.

archiewood:
Cool - will merge.

hughess (commented):
Only thing I noticed here that should change is the default setting in dev mode for the global show/hide queries option.

By default, in dev mode queries should be showing, so this line should change so that it resolves to true if there isn't any local storage variable yet:

`export const showQueries = writable(dev && browser && (localStorage.getItem('showQueries')==='true'  || false ));`

archiewood:
ah so it maybe it should be:

```export const showQueries = writable(browser && (localStorage.getItem('showQueries')==='true' || dev ));```

since I think dev is 'true'?

want to reverse the PR and I'll change this?

I can then add changesets

hughess:
I'm not sure what the exact logic should be for the localStorage check part of that line, but I think with the most recent proposed line, you could have a situation where:

- The localStorage variable is set to `false`
- The check in your file will return `false` since it won't find "true" in localStorage
- `(false || true)` will resolve to `true`

What situation does the "||" cover?

archiewood:
Okay I've tested this locally:

`export const showQueries = writable(dev && browser && (localStorage.getItem('showQueries')!='false'));`

CleanShot 2022-05-13 at 10 34 25

I think this gives the desired behaviour?

hughess:
That's perfect! 🎉

Issue #129, opened 2021-09-09
PDF export


hughess:
Function to export report content into a nicely formatted PDF file

aaronsteers:
I'd like this too but might prefer a lighter weight and more portable export format over pdf. Is there any thought around html and/or "rendered markdown" exports? Some applications I'm thinking of are to post one or more rendered output pages to a GitHub PR, as an attachment in an automated Slack post, or similar automated pushes in the format of messages/posts.

Should I open as a new issue related to markdown and/or html, or is it okay to combine here?

hughess:
@aaronsteers that's a great idea! Sorry for the delay in responding - I seem to not be getting notifications for comments and will have to turn that on.

I think it makes sense to track that idea in a new issue (I'm assuming we would tackle them separately)

hughess:
Some mixed results from looking into the PDF export options. The good news is we have a lot of control over print formatting and should be able to deliver a nice print-to-PDF experience (it's almost there now, but some spacing things to fix).

The bad news is generating a PDF without going through the default browser print settings seems pretty difficult. I tried two packages for this and got them functioning, but they don't return ideal results:

1. Takes a screenshot of the page. It can be customized to look more how we'd like, but it's just an image - you can't select the text
![CleanShot 2022-05-19 at 17 05 42](https://user-images.githubusercontent.com/12602440/169406515-95803492-e24d-4cb0-b309-7bad0f7462cf.gif)

2. Builds a PDF using the HTML code on the page. This should in theory be the best approach, but it also means we need to redefine how we want everything to be rendered, which would be a ton of work.
![CleanShot 2022-05-19 at 17 07 44](https://user-images.githubusercontent.com/12602440/169406532-fd6a86bb-1682-43a1-bff1-fdf4590a84a5.gif)

One last option that sounds promising, but I don't know enough about, is using a tool like playwright or puppeteer in the background to run the print to PDF process and download the file. If that lets us use the browser PDF rendering, that could be pretty great.

Here's the current print-to-PDF through the browser:
![CleanShot 2022-05-19 at 17 23 48](https://user-images.githubusercontent.com/12602440/169407132-10d36556-91f2-40b0-8db6-097de175f5d6.gif)

hughess:
Closing this for now, since page formatting suitable for a PDF is supported in #273.

I'm going to add a new issue for rendered markdown support.

PR #273, opened 2022-05-19
Add Print to PDF Feature

hughess:
# Print to PDF
This PR adds Print to PDF for a full page in Evidence.

CSS print styles have been added to cover the following:
 - Avoid charts and chart errors being split over a page break
 - Avoid tables longer than 20 rows being split over a page break (limit to be adjusted as needed)
 - Avoid page break immediately after an h1, h2, or h3 header
 - Maintain colours as they are displayed in Evidence
 - Remove slider from tables

Print to PDF function is accessible from the top right button in the header

## Example
CleanShot 2022-05-20 at 16 27 45

## Issues
 - Icon should be replaced with one that has a thinner border
 - We should consider how tables should be printed if they are paginated (e.g., do we want to expand the full table, or leave a note that it's not showing all records, or make this configurable?)
 - Default browser header and footers cannot be removed programmatically without removing the page margins. These need to be removed within the user's print menu in their browser
 - Spacing is inconsistent across browsers, and Safari doesn't support CSS `@page` for setting margins
 - ECharts instances have their width set dynamically by the page - this can cause problems with alignment during printing. We may need to switch to generated images when printing to gain full control over their sizing


hughess:
Closes #129

Issue #229, opened 2022-04-20
VSCode Extension overrides other useful VSCode markdown features


archiewood:
**Bug Description**
When using VSCode for a non-Evidence project, I noticed that the Evidence markdown extension was preventing some of the VSCode functionality from working. For example [markdown preview mode](https://code.visualstudio.com/docs/languages/markdown#_markdown-preview) does not work with the extension installed

**Expected Behavior**
Default VSCode markdown features to be supported

**Steps To Reproduce**
1. Install Evidence extension from VSCode marketplace
2. Attempt to launch Markdown Preview (Cmd-Shift-V) 

**Workarounds**
- ~Uninstall the Evidence extension temporarily~
- See [docs](https://docs.evidence.dev/walkthroughs/vscode-extension#markdown-preview) for best solution right now

**Environment Information**
Operating System: MacOS, Windows 11
VS Code Version: 1.66.2 
Evidence VSCode Extension Version: v0.0.6

aaronsteers:
I had a similar experience of losing the 'normal' markdown preview experience. 

As a workaround, I was able to manually change the file type back to normal markdown. 

Ideally we could retain also the generic markdown preview capabilities.

archiewood:
Yes we want to fix this, I think @hughess had a look last week, but it seemed a bit tricky.

Currently our [best workaround](https://docs.evidence.dev/walkthroughs/vscode-extension#markdown-preview) in the docs - though perhaps that's what you were describing above?

archiewood:
Just had a play around with this, and have something working

Are the shortcuts you are missing:
- [ctrl/cmd]+shift+v: markdown preview
- [ctrl/cmd]+k v: markdown side panel view

Anything else?

![CleanShot 2022-05-19 at 21 58 53](https://user-images.githubusercontent.com/58074498/169433788-e68df986-2392-4916-816f-8b60c686e998.gif)

archiewood:
Fixed in #275

PR #275, opened 2022-05-20
Fix markdown preview shortcuts for VSCode

archiewood:
**Problem**
- Currently the Evidence VSCode Extension conflicts with standard shortcuts when using markdown files - see #229 

**Desired behaviour**
- Shortcuts work as expected: 
   - shift+cmd+v: showPreview
   - ctrl+k v: showPreviewToSide

**Solution**
- Add keybindings to the evidence package that provide the normal functionality.
- These follow exactly the same logic as in the standard shortcuts, but rather than checking for the 'markdown' editorlLanguageId, it checks for 'emd' (Evidence Markdown)
   - (This prevents the keys launching preview when in non markdown files)

### DIFF
--- a/packages/evidence-vscode/package.json
+++ b/packages/evidence-vscode/package.json
@@ -105,6 +105,20 @@
         "language": "javascript",
         "path": "./snippets/js.code-snippets"
       }
+    ],
+    "keybindings": [
+      {
+          "command": "markdown.showPreview",
+          "key": "shift+ctrl+v",
+          "mac": "shift+cmd+v",
+          "when": "!notebookEditorFocused && editorLangId == 'emd'"
+      },
+      { 
+        "command": "markdown.showPreviewToSide",
+        "key": "ctrl+k v",
+        "mac": "cmd+k v",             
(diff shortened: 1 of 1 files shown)


archiewood:
Also adding a SQL autocomplete for `SELECT * FROM table LIMIT 10` which is a frequent use for exploration

archiewood:
@hughess solid feedback. 

However, I think if it didn't pop up when you wrote 'select' then you would forget it existed / never encounter it. For that reason I think changing the shortcut description probably doesn't make sense. So two proposals:
1. Delete it as it's a pretty low value feature - it doesnt take that long to write ```select * from table limit 10```
2. Add an extra snippet, just for the SELECT statement. The idea being that this would be the first thing to pop up when you write select, and you'd have to key down to get to SELECT 10.
image

I'd lean towards deletion but let me know if you think different

hughess:
@archiewood deletion makes sense to me as well

archiewood:
@hughess I think you need to re-review this for me to merge it

hughess (approved):

PR #283, opened 2022-05-25
update defaults, and replace bool with yes/no

mcrascal:


hughess (commented):
It's working for me in `example-project`, but I wasn't able to test in `test-env` because I couldn't see the telemetry panel on the settings page in that project. I'm not sure if that's supposed to auto-update

mcrascal:
Thanks, let me confirm.

Adam

mcrascal:
Works as expected in test-env. Need to install to see the changes reflected.

PR #295, opened 2022-06-02
Exposed queries as their own variable and exposed native Postgres/BigQuery types to components

ud3sh:


packages/db-orchestrator/index.cjs
```diff
@@ -2,33 +2,48 @@ const { readdirSync, readJSONSync, writeJSONSync, pathExistsSync, emptyDirSync,
...
-    if (dev) {
-        if (!pathExistsSync("./.evidence-queries")) {
-            mkdirSync("./.evidence-queries")
+const updateCache = function (devMode, queryString, data, columnTypes, queryTime) {
+    if (devMode) {
+        const {cacheDirectory, resultsCacheFile, columnTypeCacheFile} = getQueryCachePaths(queryString, queryTime);
+        if (!pathExistsSync(cacheDirectory)) {
+            emptyDirSync(cacheDirectory); //TODO I think this invalidates the entire cache.  Check if this is by design.
+            mkdirSync(cacheDirectory, { recursive: true });
         }
```
ud3sh, line 42: RE comment ^^^.  I am pretty sure this is by design.

sites/example-project/src/pages/test-analysis.md
```diff
@@ -11,7 +11,7 @@
         {x: 17, y:18, y2: 26, y3: 16}
     ]
```
ud3sh, line 13: If users insert script tags and declare their own vars, we could get var name collisions as of today (there is a query by the same name on the page).  However, the error message is pretty clear in such cases.

packages/bigquery/test/test.js
```diff
@@ -6,10 +6,11 @@ let results;
 
 test('query runs', async () => {
     try{
-        results = await runQuery("select 100 as num")
-        assert.instance(results, Array)
-        assert.type(results[0], "object")
-        assert.equal(results[0].num, 100)
+        results = await runQuery("select 100 as num");
```
mcrascal, line 9: Would it make sense to expand our tests to cover all of the native types? E.g. 

```suggestion
        results = await runQuery("select 100 as number, current_date() as date, current_timestamp() as timestamp, 'Evidence' as string");
```

packages/db-orchestrator/index.cjs
mcrascal: Yes, this is by design, the cacheing strategy is pretty rudimentary right now. The name of the directory it's looking for is based on truncating the current time to the hour. If it does exist, then we are within the hour and we can add to the existing cache. If it doesn't exist, then whatever might exist is out of date and we invalidate it all, replacing it with our fresh directory for the hour. 

We'll want to move "clear cache and refresh" into the dev UI soon enough.

mcrascal:
@ud3sh 

Why not convert types at the beginning of the page when we're creating the query_id objects? 

Put another way, why not have the each of those bullets return true? 


----- 
```working_query
select 
    100 as number, 
    current_date() as date,
    current_timestamp() as time,
    2 > 1 as bool, 
    "Apple" as string
```

- Number: {typeof working_query[0].number === "number"}
- Date:  {working_query[0].date instanceof Date}
- Timestamp: {working_query[0].time instanceof Date}
- Bool: {typeof working_query[0].bool === "boolean"}
- String: {typeof working_query[0].string === "string"}

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
```diff
@@ -1,17 +1,56 @@
...
+        if (data) {
+          let columnTypes =  data[0]['_evidenceColumnTypes'];
+          
+          if (columnTypes) {
+            columnTypes.forEach(column => {
+                let columnDisplayName = column.name;
+                if (column.typeFidelity === 'precise') {
+                  switch (column.evidenceType) {
+                    case 'date':
+                      columnDisplayName = `${column.name} 📅`;
```
mcrascal, line 21:
```suggestion
                      columnDisplayName = `${column.name}`;
```

sites/example-project/src/pages/test-analysis.md
mcrascal: Agreed. The user bringing their own data here is pretty advanced as well.

sites/example-project/src/components/modules/getColumnSummary.js
```diff
@@ -42,6 +46,7 @@ export default function getColumnSummary(data, returnType="object") {
         colName = key;
         colFmtTag = getFormatTag(key);
         colType = getColumnType(data, colName, colFmtTag);
+        evidenceColumnType = getColumnEvidenceType(data, colName);
```
mcrascal, line 49: Why can't we just accept the result of `getColumnEvidenceType` as the colType that we pass on to components relying on this?

mcrascal: Or is that just the next iteration of this?

ud3sh:
> @ud3sh
> 
> Why not convert types at the beginning of the page when we're creating the query_id objects?
> 
> Put another way, why not have the each of those bullets return true?
> 
> ```
> select 
>     100 as number, 
>     current_date() as date,
>     current_timestamp() as time,
>     2 > 1 as bool, 
>     "Apple" as string
> ```
> 
> * Number: {typeof working_query[0].number === "number"}
> * Date:  {working_query[0].date instanceof Date}
> * Timestamp: {working_query[0].time instanceof Date}
> * Bool: {typeof working_query[0].bool === "boolean"}
> * String: {typeof working_query[0].string === "string"}

We did previously talk about this approach to help the components identify the type of the value in case we can't get it from object metadata (none-enumerable props).  Turns out we can do that using non-enumerable props.  To me object metadata allows for extensibility (beyond supporting the four core types we are targeting), so we should definitely keep the metadata. Also converting any DB type to JS Date won't be as reliable as what the DB tells us.

Then the question is, should we could do both an if so what would the converted type give us on top of what we have from metadata?
One benefit would be not having to re-parse dates, stringified numbers etc at formatting time (as you'd probably need to convert all dates to JS dates before formatting for instance).  I think we should considering doing this ... for numbers this would be trivial.  For Dates this would involve looking at each DB's date type, so that would be slightly more involved.

PR #296, opened 2022-06-02
Better image support

archiewood:
# Aims
- **Enable local image support:** a user can add their own image files to their project folder, to be accessed in .md pages
- **Better default image styling**

# Current implementation details
- **Copy images from a `./static` folder**
    - As of today, only files in the  user's `./pages` are copied over into the `./evidence/template/src/pages`
    - This PR edits adds a new filewatcher to also copy the contents of a `./static` folder to `./evidence/template/static`
    - I think the `./static` folder does not support HMR, so users would need to restart the server to access new files
- **Added default img styling to the global app.css file:**
    - centred
    - max-width 100% - (but if they are smaller than that they will not be resized)
    - rounded corners (4px radius)

image

# Questions
- `alt text` now works. But there are some other bit of custom functionality that could be useful, that currently you'd need to implement via css. **Is an Evidence `<Image/>`component that could support these a good idea?**
   - A caption styled to fit with the image
   - Edit eg the width, height, corner-radius etc without needing the css
- **Is using a `/static` folder intuitive here?** Or should we allow users to dump them in the root?
   - This also would mirror SvelteKit a bit more - SvelteKit's `/static` folder is in the same dir as their `/src` folder, so `/pages` is "one level deeper" than `/static`

### DIFF
--- a/packages/evidence/cli.js
+++ b/packages/evidence/cli.js
@@ -35,24 +35,23 @@ const runFileWatcher = function(sourceRelative,targetRelative) {
     "./pages/api/**", 
     "./pages/api.+(*)"
   ]
-  for(let i = 0; i < sourceRelative.length; i++) {
-    const watcher = chokidar.watch(path.join(sourceRelative[i],'**'), {ignored:ignoredFiles})
 
-    const sourcePath = p => path.join('./', p)
-    const targetPath = p => path.join(targetRelative[i], path.relative(sourceRelative[i], p))
+  const watcher = chokidar.watch(path.join(sourceRelative,'**'), {ignored:ignoredFiles})
 
-    watcher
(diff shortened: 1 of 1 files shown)


packages/evidence/cli.js
```diff
@@ -79,7 +79,8 @@ prog
   .describe("launch the local evidence development environment")
   .action((args) => {
     populateTemplate()
-    const watcher = runFileWatcher()
+    const watcher = runFileWatcher('./pages/','./.evidence/template/src/pages/')
+    const imgWatcher = runFileWatcher('./static/','./.evidence/template/static/')
```
mcrascal, line 83: @archiewood my concern here is that we're spinning up two separate chokidar "watcher" processes. Instead of doing this, I'd prefer we modify the one watcher to watch both the pages, and the static directory.

archiewood:
Closing in favour of #351

PR #297, opened 2022-06-06
Smooth copy paste

### LINKED ISSUE #254: Smooth copy paste experience
**Feature Description**
Copy/ pasting out of an evidence report should work roughly how you'd expect.

mcrascal:
Adds a smooth copy paste behaviour to charts. 

When a copy event is fired, the normal svg charts are hidden and replaced with an  tag depicting the chart, which plays nicely with other productivity applications. 

There are other quality of life items we should do on copy/ paste, but I'd say this closes #254. 

CleanShot 2022-06-05 at 22 41 57

### DIFF
--- a/.changeset/shaggy-beans-explain.md
+++ b/.changeset/shaggy-beans-explain.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
+Provide images of charts when users are attempting to copy and paste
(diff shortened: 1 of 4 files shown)


hughess:
This is awesome.

It looks like this works well in Google Docs and Word, but in Google Slides and PowerPoint there are some sizing issues with the pasted outputs. I'm not sure if there's anything we'll be able to do about that. 

I think it's worth considering copy/paste options at the individual item level as well (e.g., copy a chart to clipboard, copy data to clipboard). Maybe a separate project, but it would make it clearer that charts are copy-able and would make the sizing issues in slides easier to handle.

PR #298, opened 2022-06-06
Make the tag line a top-level markdown header

eleweek:
To me it looks better and more in line with the rest of headers. What do you think?

## New

image

## Old

image


hughess:
That looks a lot better - thanks for fixing that!

PR #301, opened 2022-06-08
Snowflake connector update

archiewood:
## Issue
- The snowflake connector does not have fields for warehouse and database name. 
- This means `select 10 as num` works - it does not need to find a database. This is what we use to test the connection.
- However if you `select * from database_name` you get the error:
image

## Proposed solution
- Add (mandatory) warehouse and database name fields to the snowflake connector form.
image

## Results
image

## Questions
- Are there any other files we need to update? (I think no, but can't tell)
- Is there any reason we would not want these fields to be mandatory?
- Where in the css can you change the grey colour of the placeholder text, I think users are finding it a bit dark and assuming it's filled in.

### DIFF
--- a/.changeset/curvy-frogs-check.md
+++ b/.changeset/curvy-frogs-check.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/snowflake': patch
+'@evidence-dev/components': patch
+---
+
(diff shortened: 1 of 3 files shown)


ud3sh:
Nice Archie.  I think we'd need bit more of a robust test suite before we do a big adoption push.

Issue #300, opened 2022-06-07
Extend type system to all databases


ud3sh:
Extend type system support to 

1. MySQL
2. Snowflake
3. SQLLite

The tutorial site should be working at the end of this work.

ud3sh:
As expected, all five SQL Lite types had to be inferred as we don't have access to native types via query results.  As was the case with BigQuery and Postgres, MySQL and Snowflake supported native types via query results.

PR #304, opened 2022-06-08
Issue-300 added type support for Snowflake, MySQL, and SQL lite.  Extractd common DB functionality to a shared module

ud3sh:


mcrascal:
Closes #300

PR #308, opened 2022-06-09
Project name settings, ship components with template

eleweek:
Implements editing project title.
1. Rolls @evidence-dev/components into the template
2. Creates project name store as well as api for changing it


archiewood:
What do we expect companies to use this feature? 
For example, if they were part of Acme Inc., would they call the project "Acme" or "Acme reports", or "Acme data" etc.

If I was making a personal project I'd know what to call it, eg Canada Covid 19 Dashboard. Or whatever the topic.

This also makes me think: should there be a character limit I'm on projectName to prevent overflow from Sidebar?

mcrascal:
@archiewood -- "Acme, Inc." perhaps a team name, possibly an internal project name "Acme Data Portal" etc. 

Philosophy is that teams should be able to brand their projects however they see fit. 

Excel doesn't put _Powered by Excel_ all over your model when you print it out 😆 

Good point about sizing/ some sort of validation there.

mcrascal:
I'd propose we close this for the moment, but I think we should revisit both the title feature, and the more substantial "ship components with the template" relatively soon, perhaps once we've made some more progress on the testing suite. 

To make this work outside of the monorepo, I think that we'll need to sync the dependencies from the component library up into @evidence-dev/evidence, otherwise @evidence-dev/evidence will still need to import the components package.

PR #309, opened 2022-06-10
Charts leverage contexts

mcrascal:


mcrascal:
@hughess would be great to get your blessing here. No rush -- this isn't blocking anything or causing any bugs, but I think it would be a nice quality of life improvement for future development on the charting library. 

This replaces the global stores for chart config (and props) with a store that is created in `<Chart />` and passed to its children using the context apis. This isolates each instance of the config store to each instance of <Chart /> and its descendants. 

## Example 

[In this commit I wrap the Echarts component in a conditional](https://github.com/evidence-dev/evidence/pull/309/commits/d9f618645e9074c13a993a00769457bd4553910e) that should destroy the chart instance, then bring it back.

**Before this PR - when its toggled back on, we get the last chart on the page** 
CleanShot 2022-06-09 at 21 13 42

**With this PR:**
CleanShot 2022-06-09 at 21 11 35

hughess:
It seems like this is covered, but just to confirm - the only piece I'm not 100% sure about is overriding the chart config from within primitive components.

E.g., in `<Bar/>`, there's a `config.update()` call which modifies the chart-level config used by `<Chart/>`.

As long as that works, there's no other concerns I have with this; seems like a much better solution!

hughess:
I've committed the flash on copy fix here. 

For the animation being turned off, there was one unnecessary line setting it to false, so I removed that here as well.

PR #319, opened 2022-06-14
303 leverage type system in all components

mcrascal:
@ud3sh a couple q's here. 

1. I am re-using your inference logic (from `inferColumnTypes.js`) as a fall back in `getColumnEvidenceType`. Does the usage inside of getColumnEvidenceType look OK to you? This is to cover the case of someone submitting data to a component that didn't originate from a query at all. 
2. I was sort of banging my head against the wall trying to import that inference function directly from db-commons, but there's some sort of cjs issue w/ vite. Any thoughts on how to do this? 

Still outstanding for me: 

- [ ] Some icons in the queryviewer 
- [x] Testing across dbs (only BQ so far)


ud3sh:
> @ud3sh a couple q's here.
> 
> 1. I am re-using your inference logic (from `inferColumnTypes.js`) as a fall back in `getColumnEvidenceType`. Does the usage inside of getColumnEvidenceType look OK to you? This is to cover the case of someone submitting data to a component that didn't originate from a query at all.

@mcrascal makes sense to me if we want formatting to work somewhat reasonably with values that are not from DB.

> 2. I was sort of banging my head against the wall trying to import that inference function directly from db-commons, but there's some sort of cjs issue w/ vite. Any thoughts on how to do this?

@mcrascal Can't think of something trivial/easy. Not being able to easily share code between the component package and the other packages *seems* to be due to us mixing a ES6 module with CJS modules. Not an expert on the topic yet, and will have to sit down with it, but perhaps we should consider moving all our ES6 or otherwise do some restructuring to the project to make code sharing easier.  Having a set of library functions that your entire codebase can reference shouldn't have to be this hard.

With your update to `getEvidenceType`, perhaps we don't really need the inference stuff in db-commons.  i.e we simply call the type inference logic whenever, the column doesn't have a known type.  

I wouldn't bother with this though for now since that will won't fixing having to share the `EvidenceType` enum, etc.  My preference would be that we address the code sharing issue, with this as the example separately.

ud3sh (approved):

mcrascal:
Merging this and opening two additional issues: 
1. Boolean false is being formatted as null in data tables (query viewer and dataTable)
2. Add icons / indications to the queryviewer -- I have some work in progress here, but want to close this PR

PR #326, opened 2022-06-19
313 remove value and fmt options from value component

mcrascal:


hughess:
@mcrascal nice! Only other thing needed on this one was the docs changes to remove the props

mcrascal:
See #330

PR #327, opened 2022-06-20
Extend formatting system to chart tooltips

hughess:
This PR adds value and column title formatting to chart tooltips.

It includes:
 - ECharts custom `formatter` function to pull in value and title formats as specified through column format tags
 - Specific tooltip formats for `<ScatterPlot/>`, `<BubbleChart/>`, and `<Histogram/>` which override the default chart tooltip

## Before
CleanShot 2022-06-20 at 10 09 13@2x
CleanShot 2022-06-20 at 10 09 57@2x
CleanShot 2022-06-20 at 10 10 59@2x

## After
CleanShot 2022-06-20 at 10 06 26@2x
CleanShot 2022-06-20 at 10 06 48@2x
CleanShot 2022-06-20 at 10 11 28@2x

# Future Improvements
 - Add more columns to the tooltips for `<ScatterPlot/>` and `<BubbleChart/>`
    - These should include at least one identifier column (e.g., Product Name)
    - The right behaviour might be to include all columns from the query result in the tooltip to start, then provide an option to exclude specific columns
    - If we add these options, we should probably also include a prop where you can specify exactly which column(s) you would like to appear in the tooltip
 - Add back the feature to sort tooltip series based on value (max on top of list, min at bottom)
    - This PR breaks the behaviour of multi-series line chart tooltips, where the tooltip would reorder the series shown in the tooltip to match the position of the series at that point in the chart (max on top, min on bottom)
    - To add this functionality back, we will need to add a sort function to the `formatter` definition
 - Make histogram tooltip support multiple series
    - Histogram tooltips work well for a single series, but when we add the ability to plot multi-series histograms, we will need to revisit the tooltip design and examine the data structure that ECharts returns to our component
 - Axis title override behaviour
    - You can change the label that appears for the x-axis or y-axis on your chart, but at the moment these labels do not replace the column titles shown in the tooltip
    - For example, if a column name is `gdp_usd`, it will be formatted as "Gdp ($)". 
       - You can override the axis label to show "GDP", but the tooltip will still show "Gdp ($)"

### DIFF
--- a/.changeset/lovely-items-kneel.md
+++ b/.changeset/lovely-items-kneel.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': minor
+---
+
+Adds value and column title formatting to chart tooltips, with revised tooltips for scatter, bubble, and histogram charts
--- a/sites/example-project/src/components/modules/formatValue.js
+++ b/sites/example-project/src/components/modules/formatValue.js
@@ -1,6 +1,10 @@
 
 export default function(value, columnFormat, columnUnits) {
 
+  if(value == undefined){
+      value = "-"
+  }
+  
   let suffix;
   switch(columnUnits){
         case "B":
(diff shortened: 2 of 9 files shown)


archiewood:
Amazing improvement!

PR #329, opened 2022-06-20
indicate types, fix boolean styling

mcrascal:
Happy to look at an icon in the future, but I think this is probably the most straightforward for now, and doesn't require a tooltip. I also think we can add a few additional "sub-header" rows like these to display a couple other things in the future: extents, distributions, central tendency, that type of thing. 

- also cleans up formatting of boolean (and treatment of "false") 
- ~~we should switch to a tabular font in here (compare the widths of the two gdp figures). I'll open an issue for this.~~ [edit: done]

CleanShot 2022-06-20 at 11 25 28@2x

### DIFF
--- a/.changeset/kind-steaks-laugh.md
+++ b/.changeset/kind-steaks-laugh.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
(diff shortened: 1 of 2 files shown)


mcrascal:
Tabular numbers

CleanShot 2022-06-20 at 11 47 28@2x

PR #330, opened 2022-06-20
update value component docs to remove row and col

mcrascal:


mcrascal:
@hughess good eye on the title 🤦‍♂️

hughess:
I noticed the props were in the tutorial a couple of times, so I just added a commit to remove those.

@archiewood here are the places that were changed:
1. `sites/docs/docs/tutorial/advanced-loops.md`
    - Removed `fmt=usd` for the `cpa` column. I didn't want to adjust the column name in case that would cause issues elsewhere, but would changing that column name to `cpa_usd` work?
2. `sites/docs/docs/tutorial/use-query-results-in-text.md`
    - I don't think this is an issue because it's a date..the type system should get this formatted correctly now

archiewood:
Agree these are both fine.

PR #336, opened 2022-06-27
Issue-333 Polished up formatting UI, added docs, removed featured flag

ud3sh:


sites/docs/docs/features/queries/number-formatting.md
```diff
@@ -123,7 +113,7 @@ When creating a table, Evidence formats column titles based on the name of the c
 </tr>
 <tr>
 <td>growth_pct</td>
-<td>Growth (%)</td>
```
ud3sh, line 126: There was some inconsistency with the docs. Before the current formatting changes we were not showing % for instance.

sites/example-project/src/components/modules/formats.js
```diff
@@ -6,142 +6,222 @@ import { getContext } from "svelte";
...
+    formatTag: "jpy",
+    formatCode: '"¥"#,##0.00',
+    axisFormatCode: '"¥"#,##0',
+    titleTagReplacement: " (¥)",
     valueType: "number",
     exampleInput: 101.1,
   },
   {
-    formatName: "gbp",
-    formatValue: '"£"#,##0.00',
```
ud3sh, line 48: Added ability to have a separate formatting for axis labels and titles (not exposed for custom formats). In theory, these could eventually be functions depending on where we want to take them.

sites/example-project/src/pages/api/customSettings.json.js
```diff
@@ -1,8 +1,7 @@
 import fs from 'fs';
 import path from 'path';
-import { dev } from '$app/env';
 
-const CUSTOM_SETTINGS_FILE = '.custom-settings.json';
```
ud3sh, line 5: @mcrascal @hughess In oder for .custom-settings.json to be source controlled, the path has to be like this (i.e so it ends up in the user's root project directory rather than in the template.  This also happens to work nicely in the example project although the the custom settings will be in the mono-repo root.
Having that this feels a bit smelly. If you see any red flags or dealt with a similar with an alternative solution, please share.

hughess: @ud3sh that path type shows up in a few other places as well (sqlite connector, git file checks) - unfortunately I don't know of an alternative for it

ud3sh:
> @ud3sh this new formatting system is amazing! I'm adding in some more built-in formats and the flexibility/customization of the formats is phenomenal

Looking forward to seeing them!   If we are levering alternate formatting for axis labeling widely we may want to think about how we should present that in the UI (if at all). 

PS: Custom functions (that don't use codes) can be done in easily in theory if we need them.  Tricker part would be presentation.

sites/example-project/src/components/ui/Formatting/CustomFormatsSection.svelte
```diff
@@ -127,20 +137,10 @@
         <input
           id="formatCode"
           type="text"
-          placeholder="$#,##0.0"
```
ud3sh: nice

sites/example-project/src/components/ui/Formatting/CurrencyFormatGrid.svelte
```diff
@@ -0,0 +1,109 @@
+<script>
+  import "./format-grid.css";
+  import { blur } from 'svelte/transition';
+  import { defaultExample, formatExample } from "$lib/modules/formats";
+  export let formats;
+  let selectedCurrency;
+</script>
+
```
ud3sh, line 8: I was actually thinking about adding an optional display name to the formats object. Not sure if it would have helped here.

sites/example-project/src/components/ui/Formatting/CurrencyFormatGrid.svelte
```diff
@@ -0,0 +1,109 @@
...
+  <option name=chf id=chf value=chf>CHF - Swiss Franc</option>
+  <option name=cny id=cny value=cny>CNY - Renminbi</option>
+  <option name=hkd id=hkd value=hkd>HKD - Hong Kong Dollar</option>
+  <option name=nzd id=nzd value=nzd>NZD - New Zealand Dollar</option>
+  <option name=sek id=sek value=sek>SEK - Swedish Krona</option>
+  <option name=krw id=krw value=krw>KRW - South Korean Won</option>
+  <option name=sgd id=sgd value=sgd>SGD - Singapore Dollar</option>
+  <option name=nok id=nok value=nok>NOK - Norwegian Krone</option>
+  <option name=mxn id=mxn value=mxn>MXN - Mexican Peso</option>
+  <option name=inr id=inr value=inr>INR - Indian Rupee</option>
```
ud3sh, line 25: Seems like the currencies without a match in pre-built formats will show an empty table when selected.  Should we exclude these?

Got me thinking, if we were to add all of these all, we could end up having a lot of unintended formatting.  No data to back this, but my gut feeling is column names ending with _nok, _rub, _try, _cop, _sar etc are less likely to be actual containing the currency values and more likely to be a random column suffix.

hughess: @ud3sh great point. 

I had left these ones blank until figuring out how we could add them all without creating a huge list of formats (it would be 35 currencies x 13 format variations = 455 formats).

I was trying to figure out if there's a way to dynamically create the 13 variations so we don't need to create a giant object containing all 455 (e.g., you click `MXN` in the drop-down and it swaps in the `mxn` prefix and currency symbol). The tricky part would be what happens when someone uses one of the variations in their code (e.g., `sales_mxn2b`).

On second thought though, your point on the column names is good. If people want to use a specific currency we don't have, they can also add a custom number format and use that.

I propose supporting the top 10 currencies in this round. Let me know if you have any ideas on creating the list of formats - otherwise I'll set up something quick to generate them just for this time.

PR #346, opened 2022-07-11
Turn on quote formatting in mdsvex

hughess:
Turns on styling for quotes, so when you save a quote in your `.md` file, it includes curly quotes on your webpage.

## Before
CleanShot 2022-07-11 at 17 03 44@2x
CleanShot 2022-07-11 at 17 03 49@2x

## After
CleanShot 2022-07-11 at 17 02 03@2x
CleanShot 2022-07-11 at 17 02 33@2x

Quotes included in SQL queries appear the same way (as straight quotes rather than curly).

### DIFF
--- a/.changeset/lovely-maps-love.md
+++ b/.changeset/lovely-maps-love.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/preprocess": patch
+---
+
+Turns on quote formatting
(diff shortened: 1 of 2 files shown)


hughess:
@mcrascal I saw that you had set quotes to `false` in the mdsvex settings - was there a reason they were turned off?

mcrascal:
I changed this in April 2021 to fix some pre-processor issue (that's all I can get from the old repo), but I doubt it's still applicable. I much prefer the curly quotes, so if it's working now, I'd go for it.

hughess:
Thanks!

PR #351, opened 2022-07-19
Static folder support

archiewood:
This generalizes #296.

Copies any content from a `/static` directory in the root of a project into svelte-kit's `/static` folder.
- Allows images to be hosted locally.
- Adds some default image styling

`image alt` will now work.


archiewood:
Image styling defaults:
- centred
- 4px radius corner rounding
- max width 100% 

image

PR #361, opened 2022-07-25
Mobile Hamburger configuration

archiewood:
**Why**
The mobile hamburger is currently in an unexpected location (bottom-left)
This makes it hard to find, and can also mean it is not clear when evidence is in "mobile" mode.

**What is best practice?**
When you look at most modern websites and apps, the mobile hamburger menu is either in the top left (eg Github - this page, https://www.adidas.ca/), or top right (eg https://getbootstrap.com/, https://www.nike.com).

**Potential solution**
- This PR includes a top right positioning.
  - This is a pretty easy implementation as it doesn't block the current nav items.
  - This can be done with the slide-in as is, from the left, or altering the slide-in to be from the right. See gifs below. 
  - Personally prefer the slide from the left as before, but keen to hear other views

**From Left** &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **From Right**
                 

**Other thoughts**
Most implementations have a full top navbar which *includes* a hamburger, rather than a floating hamburger button. This might be a better option, but does reduce screen space?
image

### DIFF
--- a/.changeset/rich-cows-rescue.md
+++ b/.changeset/rich-cows-rescue.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+moving mobile hamburger to top right
--- a/sites/example-project/src/app.css
+++ b/sites/example-project/src/app.css
@@ -156,7 +156,7 @@ code {
 
 :root {
 	/* Layout header height */
-	--header-height: 2.5rem; 
+	--header-height: 3.5rem; 
 	/* Font stacks */
 	--monospace-font-family:  Menlo, Monaco, "Lucida Console", "Liberation Mono", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", "Courier New", monospace;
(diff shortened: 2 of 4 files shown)


archiewood:
@mcrascal this is related to your comment re the Stackblitz setup. I think if we had something more like this then we could remove the "you are in mobile mode" text.

archiewood:
After a bit of css shuffling, the menu now **visually** appears on top of the top breadcrumbs / header bar.
However - should clean this up in the future to actually put the hamburger **inside** the header component.

5c8a-91a9-6f74-6572bf1fe7dd

PR #366, opened 2022-07-28
Better Feedback on Query Status

mcrascal:
This PR does two things: 

1. Trigger invalidate on the data object for each page whenever queries have changed, and update their contents client side
1. Report on the status of the queries with "toasts" in the bottom right 2. 

See below -- no hard refresh on SQL edits: 

CleanShot 2022-07-28 at 14 54 24

Still to do: 

- [x] Some of our built-in components are not reactive to client side updates to the data object 
- [x] The set context operation in preprocess is also not reactive


archiewood:
Cool - big improvement on the current "have to go back to my console to see if it's ready" experience.

**Could you make it more obvious the that query is "running" when it's grey?**
I'm imagining that if I had a query that was taking a longer time to run eg Snowflake seems to take a while, then it might not be super obvious what all the grey boxes mean?

For example you could make the text say: 
- working_query running
and then
- working_query complete / failed

Or you could animate the toast with a ... or a spinner or something?

archiewood:
Just had another thought while I was working in stackblitz. The server crashed and the only feedback was on the console.

The final part of the "feedback experience" that hasn't made it into the UI yet, is whether the server is running (or has crashed).
It might be out of scope for this, but a little card popping up saying "Server crashed" would address this.
_(Obviously it would be best if the server never crashed silently - ie with no visual feedback, but it still does in some edge cases)_

sites/example-project/src/pages/__layout.svelte
```diff
@@ -5,13 +5,16 @@
...
+	import {browser} from '$app/env'
 
 	import TableOfContents from "@evidence-dev/components/TableOfContents.svelte";
 	import Header from '@evidence-dev/components/ui/Header.svelte'
 	import Hamburger from '@evidence-dev/components/ui/Hamburger.svelte'
 	import Sidebar from '@evidence-dev/components/ui/Sidebar.svelte'
 	import LoadingIndicator from "@evidence-dev/components/ui/LoadingIndicator.svelte";
-
+	
+	import QueryStatus from "$lib/QueryStatus.svelte";
```
mcrascal, line 16: needs to reference the component library

mcrascal:
@eleweek 

Picking up the conversation here. Three items outstanding to get this into main: 

- [ ] preprocess creates local variables for each query `{query_name}`  in a way that is not reactive 
- [ ] tests aren't passing on this PR, in part because of [this](https://github.com/evidence-dev/evidence/pull/366/files?file-filters%5B%5D=.svelte&show-viewed-files=true#r943010415)
- [ ] merge conflict in __layout

mcrascal:
@archiewood @ud3sh @hughess -- this is ready for review. 

Would be helpful to get some additional tire kicking on this before we merge.

Thanks,

PR #372, opened 2022-08-04
Settings formatting nits

mcrascal:
Some further refinements on the styling and copy in the formats section of the settings. 

- Data tables aligned to header, increased size, removed border box 
- More obvious interaction and working transitions in the currencies section 
- Cleaned up description + example 
- Remove layers on the custom formats section. Form is visible from the beginning, only one action. 
- Collapsible section in the custom formats area serves the same purpose as collapsible sections above it. 
- Collapse all sections to ensure that the custom formats input is visible from the docs section at the top.

**After** 
CleanShot 2022-08-04 at 11 19 44

### DIFF
--- a/.changeset/cold-timers-swim.md
+++ b/.changeset/cold-timers-swim.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Updates to settings page layout
--- a/sites/example-project/src/components/ui/Formatting/BuiltInFormatGrid.svelte
+++ b/sites/example-project/src/components/ui/Formatting/BuiltInFormatGrid.svelte
@@ -4,8 +4,7 @@
   export let formats;
 </script>
 
-<div class="tableContainer">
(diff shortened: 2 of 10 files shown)


archiewood:
## Possible builds:

- Like the new code block a lot - much clearer!
  - The colouring of the background of the code block is a bit aggressive vs the rest of Evidence currently, which is pretty neutral. 
  - Perhaps something closer to the [PrismJS default](https://prismjs.com/) theme?
- Move Value formatting above telemetry?

mcrascal:
@archiewood 

- [x] Updated to the prism theme we use for SQL blocks elsewhere. +1 for consistency -1 for rock and roll.  
- [x] Move Value formatting above telemetry?
- [ ] Open the currency section by default. Currencies is sort of a special case. I tried it with percentages (the next smallest) and it still didn't let you see the whole scope of the section. 

CleanShot 2022-08-10 at 20 57 38@2x

PR #375, opened 2022-08-08
Postgres Connector Case Normalization Fix

archiewood:
## Context:
- User reported difficult connecting to Cube.js via their [Postgres SQL API](https://cube.dev/docs/backend/sql)
- Cube.js compiles queries, and then passes the results back to the query engine.
- Cube.js by default uses camel case for the names of the fields that are passed back

## Problem
- The postgres connector lowercases all of the field names: `myColumn` -> `mycolumn`
- But the `evidenceType` column name is not lowercased
- This means that if when Evidence looks for the type of `mycolumn` it cannot find it.

Result:
image

## Proposed fix
- Lowercase the field name for the column types object also

### DIFF
--- a/.changeset/popular-bats-tell.md
+++ b/.changeset/popular-bats-tell.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/postgres": patch
+---
+
+Fix postgres connector to handle cases consistently between query object and types object
--- a/packages/postgres/index.cjs
+++ b/packages/postgres/index.cjs
@@ -61,7 +61,7 @@ const mapResultsToEvidenceColumnTypes = function (results) {
         }
         return (
           {
-            'name': field.name,
+            'name': field.name.toLowerCase(),
             'evidenceType': evidenceType,
             'typeFidelity': typeFidelity,
           });
--- a/sites/example-project/src/components/modules/getColumnEvidenceType.js
+++ b/sites/example-project/src/components/modules/getColumnEvidenceType.js
@@ -10,11 +10,11 @@ export default function getColumnEvidenceType(data, column) {
     }
     if (item && item['_evidenceColumnTypes']) {
       let columnTypes = item['_evidenceColumnTypes'];
-      return columnTypes.find(item => item.name === column);
+      return columnTypes.find(item => item.name?.toLowerCase() === column?.toLowerCase());
     } else {
       // infer types as a fall-back (when someone is passing arbitrary data objects)
       let columnTypes = inferColumnTypes(data);
-      return columnTypes.find(item => item.name === column);
+      return columnTypes.find(item => item.name?.toLowerCase() === column?.toLowerCase());
     }
   }
 


ud3sh:
@archiewood Added a small change for this to work in the more general case (i.e normalize the column name for case whenever we look up the evidence type).  
On a side note, I don't know if I fully get the use case of cube.js.  
Question: "The postgres connector lowercases all of the field names: myColumn -> mycolumn".  Are you referring to the postgres connector library we are using?

archiewood:
Okay great!

> On a side note, I don't know if I fully get the use case of cube.js.

Cube.js allows you to build a **metrics layer.** 

- If you need to have one data warehouse, but multiple BI endpoints (eg Looker + Evidence), you don't want to have to define the logic for metrics in two different places. 
- Without it, if you use eg Looker, then you do the metric definition in LookML. But this is pretty annoying if you want to use the same metric (eg Daily active users) in another tool.

> "The postgres connector lowercases all of the field names: myColumn -> mycolumn". Are you referring to the postgres connector library we are using?

I was just looking at this function in index.cjs, which lowercases the field names when we save the data
```
const standardizeResult = async(result) => {
    var output = [];
    var output = [];
    result.forEach(row => {
    result.forEach(row => {
        const lowerCasedRow = {};
        const lowerCasedRow = {};
        for (const [key, value] of Object.entries(row)) {
        for (const [key, value] of Object.entries(row)) {
            lowerCasedRow[key.toLowerCase()] = value;
            lowerCasedRow[key.toLowerCase()] = value;
        }   
        }   
        output.push(lowerCasedRow);
        output.push(lowerCasedRow);
    });
    });
	return output;
	return output;
}
}
``` 

vs this function (also in index.cjs), which currently does not, when, we save the data types
```
const mapResultsToEvidenceColumnTypes = function (results) {
    return results?.fields?.map(field => {
        let typeFidelity = 'precise';
        let evidenceType = nativeTypeToEvidenceType(field.dataTypeID);
        if (!evidenceType) {
            typeFidelity = 'inferred';
            evidenceType = EvidenceType.STRING;
        }
        return (
          {
            'name': field.name,
            'evidenceType': evidenceType,
            'typeFidelity': typeFidelity,
          });
    });
};
```

archiewood:
Result:

**Data object we cache**
image

**Column types object we cache**
image

ud3sh:
Got it, great catch and thanks for the background.  Will checkout the background for lower casing the keys with Adam at some point.  My opinion is in general, we should display the columns in the article as they were set in the DB / BI tool (though snowflake always makes it ugly), but should normalize for casing for all other kinds of logic since most database column names are case insensitive.

Regardless, I think the changes we made should take care of the user issue at hand.

archiewood:
Okay excellent. I've tested this and it works for me.

Good to merge?

ud3sh (approved):

ud3sh:
> Okay excellent. I've tested this and it works for me.
> 
> Good to merge?

Yep!

PR #379, opened 2022-08-12
BigValue Component

mcrascal:
Adds a beloved BI feature: the **Big Number Viz**&#8482; to Evidence. 

Includes: 
* Number 
* Sparkline 
* Comparison
* Comparison is configurable for "down is good"
* Multiple Big Numbers stack into a row 
* Smart enough defaults that only the `data` prop is required 

BigValue Example


archiewood:
Love this feature!

## Defaults
1. Not sure the default behaviour is working perfectly here. Or maybe it is but the results are a bit puzzling:

**Eg1**
````
```orders
select * from orders limit 10
```

<BigValue data={orders}/>
````

image

**Eg2**
image

**Eg3**
image



## Visual Nits (for considerations)

1. It looks a bit like the sparkline is cropped at the bottom (but actually I think it is to do with the svg "brushstroke", rather than margins etc?)
image

3. For a BigValue, the Value could be a bit Bigger, perhaps closer to H1 than H2 size - cant remember exactly how _em_ vs _rem_ work but _1.5em_ looks like this:
image

4. Not sure this is an issue but the error message bleeds out rather than truncates. Also other components label what they are in this error message (eg LineChart)
image


## File structure
1. Any reason this is not in the Viz folder with other chart / value components?
image

mcrascal:
Thanks @archiewood, excellent feedback as always. Just consolidating here: 

- [ ] Drop the default system 
- [ ] Tidy up error messages 
- [ ] Move the component to viz, and add it to the preprocessor default imports 
- [ ] Take one more look at sizing and consider nudging up

mcrascal:
@archiewood updates in the branch, except for the size changes. Take a look, and feel free to merge if you're comfortable. 

Once you get the actual value close to H1, two things happen: 

1. They compete with the H1 Headers, so it's harder to understand where they necessarily fit. 
1. Just a lot less space for these.

I'd propose shipping at the current size and seeing how that feels in practice. 

Current Size: 
CleanShot 2022-08-15 at 14 52 16@2x

H1 Sized: 
CleanShot 2022-08-15 at 14 51 40@2x

archiewood:
Will have a play around with the new lack of defaults

Re sizing: happy to ship as is:
- I was probably thinking of making just the number sized H1, not the title of the <BigValue/>. Agree above definitely competing with actual H1s

mcrascal:
Ah got it 

Yeah looks pretty good IMO 

CleanShot 2022-08-15 at 15 58 28@2x

archiewood:
I noticed that if you are in a wide enough viewport to see it, the TableofContents shows "Big Value" every time there is one, as the logic in this displays all the H1 elements.

Solved this by using the `<ErrorChart/>` component for Big Value's error message, which styles the Error differently.

mcrascal:
Thank you @archiewood 🚢

PR #386, opened 2022-08-23
Fix tooltip column title formatting issue

### LINKED ISSUE #385: Currency column formats not replaced by currency symbol in tooltip
**Bug Description**
When using currency format codes, the format tag is not replaced by the symbol when shown in a chart tooltip.

**Expected Behavior**
Format tag should be replaced by the currency symbol (e.g., `sales_usd0` ---> `Sales ($)`)

CleanShot 2022-08-23 at 12 32 50@2x

hughess:
In `getSeriesConfig`, there was a scenario where we were passing echarts an unformatted column title - it only showed up in single series charts. This PR passes a formatted title instead, solving the tooltip formatting issue.

Closes #385

### DIFF
--- a/.changeset/purple-cougars-sip.md
+++ b/.changeset/purple-cougars-sip.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Fix for tooltip column name formatting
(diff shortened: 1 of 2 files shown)


ud3sh:
> In `getSeriesConfig`, there was a scenario where we were passing echarts an unformatted column title - it only showed up in single series charts. This PR passes a formatted title instead, solving the tooltip formatting issue.
> 
> Closes #385

Nice @hughess !

PR #393, opened 2022-08-31
Initial implementation of native markdown table formatting

hughess:
This PR adds styling to native markdown tables by applying styles to the global HTML table elements and tagging all other tables with a class (currently called `component-table`), which excludes them from this styling.

## Potential Issues
 - Having to add a class to any table that's not a markdown table feels strange
 - There may be a better way to do this through mdsvex's custom components (https://mdsvex.pngwn.io/docs#custom-components)
 - The `component-table` class is not descriptive enough - need a better name
 - The styling needs work, but as an interim option it's a big improvement over what markdown tables look like today

## Before
CleanShot 2022-08-31 at 14 51 35@2x

## After
CleanShot 2022-08-31 at 14 50 48@2x

### DIFF
--- a/.changeset/twelve-socks-decide.md
+++ b/.changeset/twelve-socks-decide.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Adds styling to native markdown tables
--- a/sites/example-project/src/app.css
+++ b/sites/example-project/src/app.css
@@ -154,6 +154,32 @@ code {
 	}
 }
 
+table:not(.component-table){
+    width:80%;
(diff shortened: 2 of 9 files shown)


mcrascal:
@hughess I like this -- a couple proposed changes in that commit: 

Rather than avoiding the tables in components with :not, I think we can select tables that are the direct children of the article (only the md tables are). That'll hopefully save us from having to remember this the next time we're doing something w/ a table in a component. 

A couple changes to just use vars for colors and fonts, and commented out the table borders, but that's just preference.

hughess:
@mcrascal nice, that's a lot better!

PR #417, opened 2022-09-20
Remove link from breadcrumbs if page does not exist

hughess:


archiewood:
Hey @hughess.

I'm now testing a project with parameterized pages.

This change seems to remove the ability to click back up onto parameterized pages. Is this inevitable / the only way we can fix this error?

Example folder structure
```
`-- pages/
    |-- index.md
    `-- [org]
        |-- index.md
        `-- [sub-org].md
```

Then if I am on a specific `sub-org` page  then the link to the specific `org` they are within is empty.

Home / specific_org / specific_sub_org
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Dead link

hughess:
@archiewood good example - I missed this case when fixing the bug.

One way we might be able to do it is by modifying the check we use to see if a file exists. We can include a check for square brackets in the .md filename and allow those to be included as links.

I'll play around with this to see if that's possible.

Issue #382, opened 2022-08-15
Value component treats dates differently to other components


archiewood:
**Bug Description**
Value component returns different dates to other components

**Expected Behavior**
Value component returns the same date as other components

**Error Messages and Screenshots**
<img width="798" alt="image" src="https://user-images.githubusercontent.com/58074498/184674940-eed8df7b-5fe5-4f7c-8152-e43dade3b4f7.png">

**Steps To Reproduce**

https://stackblitz.com/edit/evidence-k2rxhd?file=pages%2Findex.md

**Database:**
- [ ] SQLite (but I think to would affect all)

archiewood:
From @ud3sh: It seems that the Value component has some explicit code for dates that doesn’t exist in the other components.

```
columnSummary = getColumnSummary(data, 'array'); 
let dateCols = columnSummary.filter(d => d.type === "date")
dateCols = dateCols.map(d => d.id); 
if(dateCols.length > 0){ 
  for(let i = 0; i < dateCols.length; i++){ 
    data = getParsedDate(data, dateCols[i]); 
  } 
}
```

hughess:
@archiewood the stackblitz example page is awesome - such a great way to share this context.

I've almost got the reasons for this figured out. Only piece of info I still need is which date you expected to see as the first row in your example: `2021-12-31` or `2021-12-30`?

archiewood:
**I think** I expected to see `2021-12-31` 

the underlying data is in a SQLite db and therefore stored as a string.
- The string stored is '2021-12-31 HH:MM:SS' and in the SQL I have truncated that using:
```
select
substr(order_datetime,1,10) as date,
...
```
There is no timezone information associated. 

This is then converted into a date, and it assumes the time is 00:00:00
If you are not in UTC (and specifically if you are UTC -X) , then this will be translated back into the previous day

hughess:
I think in this case, the `Value` component is working as we'd expect - this is because of the explicit code you referenced above.

The `getParsedDate` function attempts to account for the timezone by appending a timestamp which causes javascript to leave the date as is (and avoid doing that conversion from UTC to local time which brings it back one day).

There are several areas in the project where javascript needs to interpret the dates for us. Here's that list with the current date parsing logic for each:
- `Value` - uses `getParsedDate`
- `DataTable` - uses `getParsedDate`
- `QueryDataTable` (in the query viewer) - uses javascript default
- Chart axes - uses ECharts default
- Chart tooltips - uses javascript default

[...]

PR #423, opened 2022-10-03
Apply consistent date handling across components

hughess:
# Summary
This PR applies consistent date and time treatment across components.

Prior to these changes, the following components used the `getParsedDate` function logic to handle dates:
 - `Value`
 - `DataTable`
 - `Chart` (axes)

The following components and functions were using default javascript logic, which in some cases led to different date results:
- `QueryDataTable`
- `formatValue`

These changes update `QueryDataTable` and `formatValue` to treat dates consistently to the other components mentioned above.

# Before & After
## Before:
- QueryDataTable showing dates inconsistent with query input
  CleanShot 2022-10-03 at 11 10 01@2x
- Chart tooltips not matching axis label values
  CleanShot 2022-10-03 at 11 10 13@2x

## After:
CleanShot 2022-10-03 at 11 13 10@2x
CleanShot 2022-10-03 at 11 21 46@2x

# Next Steps
Consistency is the first issue to solve for date handling in Evidence, but there are still two remaining issues related to dates:
1. Fix `getParsedDate` logic to be applicable across all browsers (currently causing issues in Safari and Firefox when used with Snowflake)
2. Centralize date logic alongside Evidence's data type inference (to avoid having to account for dates specifically in each new component we build)

### DIFF
--- a/.changeset/tame-pens-beam.md
+++ b/.changeset/tame-pens-beam.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Applies date and time handling consistently across components
--- a/sites/example-project/src/components/modules/formatting.js
+++ b/sites/example-project/src/components/modules/formatting.js
@@ -176,7 +176,7 @@ function applyFormatting(
       let typedValue;
       try {
         if (columnFormat.valueType === "date" && typeof value === "string") {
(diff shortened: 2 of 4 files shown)


hughess:
Closes #382

PR #425, opened 2022-10-09
Some minor fixes for docs

roveo:
While reading through docs, I noticed some minor problems with SQL queries and a typo, this PR fixes them.

- [Missing groupby](https://docs.evidence.dev/features/queries/sql-queries#query-blocks)
- [The last query is not valid for most DB engines, you need an alias for the subquery](https://docs.evidence.dev/features/queries/query-chaining#example)
- typo [componet](https://docs.evidence.dev/features/advanced/templating#text-expressions)

### DIFF
--- a/sites/docs/docs/features/queries/query-chaining.md
+++ b/sites/docs/docs/features/queries/query-chaining.md
@@ -43,7 +43,7 @@ from (
         sum(sales) as sales
     from production.daily_sales
     group by 1
-) as t
+)
 ```
 
 ## Presentation
@@ -53,6 +53,8 @@ You can choose whether you want to see the compiled or written SQL inside the qu
 ## Other
 
 The order that queries appear on the page doesn't matter to the SQL compiler. You can reference queries that appear before or after the query that you are authoring.
+
+Some SQL dialects require sub-queries to be aliased, including Postgres and MySQL. E.g. `from ${sales_by_region} as sales_by_region`.
 
 The SQL compiler detects circular and missing references. If a query includes either a circular reference or a missing reference, Evidence will display an error that looks like a syntax error in a normal SQL query. Queries with compiler errors are not sent to your database.
 
--- a/sites/docs/docs/features/queries/query-chaining.md
+++ b/sites/docs/docs/features/queries/query-chaining.md
@@ -2,7 +2,7 @@
 sidebar_position: 2
 hide_title: false
 hide_table_of_contents: false
-title: Query Chaining 
+title: Query Chaining
 ---
 
 Query chaining enables analysts to reference the results of a query from other queries.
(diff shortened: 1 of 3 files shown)


archiewood:
Hmm I think because this is a fork it doesn't have access to secrets, which is causing the tests to fail.

Since this is a docs-only change, it shouldn't actually need to run the tests. Will investigate

sites/docs/docs/features/queries/sql-queries.md
```diff
@@ -5,15 +5,16 @@ hide_table_of_contents: false
...
 
 ## Query Blocks
 You can include SQL queries in your page using a markdown code block (starting and ending with 3 backticks). Evidence requires a query name to be supplied directly after the first 3 backticks.
 
 ````markdown
 ```sales_by_country
 select country, sum(sales) as sales
-from international_transactions 
+from international_transactions
+group by 1
```
archiewood, line 17: Good catch. Some SQL dialects (I believe eg Snowflake?) would run this but many would choke

sites/docs/docs/features/queries/query-chaining.md
```diff
@@ -17,33 +17,33 @@ For example, if you want to reference a query named `daily_sales`, you would wri
...
-    select 
+    select
         region,
         sum(sales) as sales
-    from production.daily_sales 
-    group by 1 
-)
+    from production.daily_sales
+    group by 1
+) as t
```
archiewood, line 46: Agree that this would not run in many dialects. 
However, our compiler does not **_currently_** add the alias if the user does not (maybe it should). 

If you try this in Postgres for example, it will throw the error `Subquery in FROM must have an alias`

For clarity in the docs though, I think if we are going to add this alias, we need to add it to the query above also!

ie `from ${sales_by_region} as t`

sites/docs/docs/features/queries/query-chaining.md
```diff
@@ -17,33 +17,33 @@ For example, if you want to reference a query named `daily_sales`, you would wri
...
 group by 1
 ```
 
 Next, we have a query calculating average sales across regions, which references *sales_by_region* using the `${ }` syntax.
 
 ```sql
-select 
+select
     avg(sales) as average_sales
 from ${sales_by_region}
```
archiewood, line 32: ... `from ${sales_by_region} as t`

sites/docs/docs/features/queries/query-chaining.md
roveo: I'm projecting my dbt experience on this, so I'm expecting it to work the same way (`ephemeral` materialization).

If you're not using an alias, the compiler generates one for you:

```sql
select a, b, c
from ${some_table}
---
select a, b, c
from (select * from some_table) as generated_alias
```

And if you're using one, the compiler will use that:

```sql
select t.a, t.b, t.c
from ${some_table) as t
---
select t.a, t.b, t.c
from (select * from some_table) as t
```

So ideally it needs to work both ways.

archiewood: Okay great - I'm going to add an issue to track this

archiewood: Issue added [here](https://github.com/evidence-dev/evidence/issues/427). For now I'm just going to include a note in the docs about a sub-query alias being needed in some dialects.

PR #426, opened 2022-10-14
Consistent article width across breakpoint

archiewood:
@mcrascal this is the thing I was mentioning yesterday

## Situtation today:
- Weird "getting larger as you make width smaller" effect. Toggles between 70ch and 80ch width
- Especially odd when there is no Table of Contents

CleanShot 2022-10-14 at 18 08 02

## Proposed Change
- Always 70ch width

CleanShot 2022-10-14 at 18 07 11

Thoughts?

### DIFF
--- a/.changeset/fluffy-plants-shake.md
+++ b/.changeset/fluffy-plants-shake.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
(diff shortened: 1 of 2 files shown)


archiewood:
Including going across mobile breakpoint

CleanShot 2022-10-20 at 10 23 20

PR #429, opened 2022-10-17
Add pull_request_target and approval workflow

archiewood:
## Problem:
- Tests fail from PRs from non maintainers. 
- This is because forks by default do not have access to the secrets that are needed for the tests to run.

## Implementation:
- Set up `pull_request_target` to trigger the action. This can access secrets.
- However, this requires an authentication workflow to ensure there are not malicious attempts to use said secrets.
- This is set up, such that maintainers will have to "approve" PRs, allowing the tests to run
   - For existing maintainers / contributors, approval is automatic

This is well explained (and mainly copied from) [here](https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,8 +2,6 @@ name: Tests
 
 on:
   pull_request:
-    branches: 
-      - master
   
   Tests:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,7 @@
 name: Tests
 
 on:
-  pull_request_target:
+  pull_request:
     branches: 
       - main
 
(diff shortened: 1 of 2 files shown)


archiewood:
Okay so, testing changes to your CI is pretty hard. 

Because of the way `pull_request_target` works, adding it to your workflows will not cause the actions to run in the PR. 
(It has to be in the branch you are merging into in order to run as expected)

I've testing this on another repo I'm a maintainer of and it works as expected there.

However, would appreciate a review of this to check I haven't missed anything @ud3sh

.github/workflows/tests.yml
```diff
@@ -1,11 +1,22 @@
...
 jobs:
+  approve:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Approve
+        run: echo "For security reasons, all pull requests need to be approved first before running any automated CI."
+
   Tests:
     runs-on: ${{ matrix.os }}
+    needs: [ approve ]
```
ud3sh, line 17: Would this create the need for an approval for regular pull requests?

PR #449, opened 2022-11-03
Auto load database variables from `.env`

afzalsayed96:
This allows configuring database variables from `.env` at the root of an `evidence` project as follows:

```
SQLITE_FILENAME=needful_things.db
DATABASE=sqlite
SEND_ANONYMOUS_USAGE_STATS=no
```

### DIFF
--- a/packages/db-orchestrator/index.cjs
+++ b/packages/db-orchestrator/index.cjs
@@ -48,7 +48,7 @@ const validateQuery = function (query) {
 
 const importDBAdapter = async function(settings) {
     try {
-        databaseType = settings ? settings.database : process.env["DATABASE"] || process.env["database"]
+        databaseType = (settings && settings.database) || process.env["DATABASE"] || process.env["database"]
         const { default: runQuery } = await import('@evidence-dev/'+ databaseType);
         return runQuery
     }catch {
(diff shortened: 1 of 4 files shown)


afzalsayed96:
Interesting! No particular use case. I was just checking the project by using the template and noticed that `.env` wasn't supported out of the box. Reading the code, I found `.env` variables being read however. Which led me to creating this PR.

archiewood:
Understood!

The scenario where environment variables are currently used is for when users are deploying their project. 
For example, if you deploy to a static site hosting service like [Netlify](https://docs.netlify.com/environment-variables/overview/) or [Vercel](https://vercel.com/docs/concepts/projects/environment-variables), then you can set your secrets as environment variables.

When those hosting services build your project, node will read the environment variables stored there, and use them to run the queries in your project!

ud3sh:
If we were to support .env for the core evidence project, we'd probably want to document the variables that can be overriden and update the as needed read the values when available.  

On a side note, we currently we have .env support setup for running local tests (via a test dependency).  We haven't documented which vars can be overriden but they are essentially what's listed under `env:` in the github test workflow. https://github.com/evidence-dev/evidence/blob/main/.github/workflows/tests.yml

archiewood:
I'm going to close this for now. I don't think it makes sense to support this for now as I think it's slightly atypical usage. Happy to come back to this if people strongly disagree!

PR #450, opened 2022-11-03
Update chart size on render finished

afzalsayed96:
I was playing around and found an issue related to chart layout when performing client side navigation

https://user-images.githubusercontent.com/14029371/199779439-3b1c19a1-0d19-408e-980d-275415adebd1.mov

I'm new using echarts so not sure if this would be the most idiomatic solution to this problem. Happy to hear your thoughts

### DIFF
--- a/.changeset/strong-mugs-prove.md
+++ b/.changeset/strong-mugs-prove.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Fix for chart size in flex layout
--- a/sites/example-project/src/components/modules/echarts.js
+++ b/sites/example-project/src/components/modules/echarts.js
@@ -438,15 +438,18 @@ export default(node, option, renderer) => {
 
     const chart = echarts.init(node, 'evidence-light', {renderer: 'svg'});   
 
+    const resizeChart = () => { chart.resize();}
+    
+    window.addEventListener("resize", resizeChart);
+    chart.on('finished', () => { chart.resize();})
 	chart.setOption(option);
-
-    window.addEventListener("resize", () => { chart.resize();});
 
 	return {
 		// update(option){
 		// 	chart.update(option, true, true);
 		// },
 		destroy() {
+            chart.off('finished', resizeChart)
 			chart.dispose();
 		}
 	};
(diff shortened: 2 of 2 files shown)


hughess:
@afzalsayed96 thanks for looking into this!

I tried your changes and got the resizing working when I use client-side navigation, but ran into a couple of issues - do these also show up on your end, or are you seeing different behaviour?
 - The chart gets into the correct width eventually, but still starts in the extra-wide position
    - In the long-run, we might need to adjust how we are initializing and loading the charts, but I think this solution is better than what we have currently (where the only solution is to refresh the page, as your screen recording shows)

 - The "finished" event looks like it continues to fire after the chart has finished rendering
    - I added `console.log("resizing")` to the resize function to see how often it was firing

afzalsayed96:
@hughess It seems the root cause of the issue as pointed in https://github.com/apache/echarts/issues/11791 is related to flex layout of the container. And the solution is to observe the flex child for resize events.

While, this solution seem does better than the previous one, it relies on the assumption that class `'chart-container'` would be always applied. Let me know what you think of it.

hughess:
@afzalsayed96 I've tested with some charts I set up to be less than 100% width and it works well.

After taking a closer look, I have question about this line:
`const chartContainerElement = document.getElementsByClassName('chart-container')[0]`

If a page has multiple charts, this will only select the first chart on the page - is that okay in this case because we only need to observe one chart on the page to effectively trigger the resizing for all the charts?

afzalsayed96:
> If a page has multiple charts, this will only select the first chart on the page - is that okay in this case because we only need to observe one chart on the page to effectively trigger the resizing for all the charts?

Yes, incidentally this would have worked as when one `chart-container` would resize due to it's parent container it's other siblings would resize as well. But it wasn't that clean after all.

Hence, I've switched the logic to observe the parent `article` element which triggers resize of it's children `chart-container` elements.

hughess:
That's perfect. I'd say this is good to go! Going to merge into main.

Thanks again for this contribution @afzalsayed96!!

afzalsayed96:
Hey Archie,

Let me take a look at this later tonight and get back to you

archiewood:
Source code I used to generate this:
https://github.com/evidence-dev/demo/tree/add-customer-reporting-example

Issue #410, opened 2022-09-14
Top-level parameterized pages shown in sidebar


archiewood:
**Bug Description**
- If you have a markdown page [my-page].md in the root of your evidence project, then it acts as a parameterised page as expected. 
- But it also shows up in the sidebar which looks weird.
- For 1st and second level folders containing parameterised pages, this does not happen

<img width="375" alt="image" src="https://user-images.githubusercontent.com/58074498/190190452-59a97876-65f4-4342-9320-9f77e59e7979.png">

**Severity**
- [ ] Low - I'm able to keep using Evidence as normal, but flagging this for the team to fix

**Expected Behaviour**
- Not show is sidebar
- possibly crawl links in pages to determine which parameterized pages *should* be show in sidebar.

archiewood:
I think this is just a case of editing the logic in https://github.com/evidence-dev/evidence/blob/main/sites/example-project/src/components/ui/Sidebar.svelte to mirror what we do for 1st and 2nd level directories.

quoid:
I took a quick look at this and it *looks* the root Sidebar elements are populated by the `noFolders` array in [`Sidebar.svelte`](https://github.com/evidence-dev/evidence/blob/main/sites/example-project/src/components/ui/Sidebar.svelte). `noFolders` seems derived from the `menu` prop, which is an array passed by [`__layout.svelte`](https://github.com/evidence-dev/evidence/blob/449a29056a2e1a4759a58a39e4eb7b3acbadd411/sites/example-project/src/pages/__layout.svelte#L14).

That *seems* like the place to make the change. For the folder based sidebar elements it seems like you are simply checking for the existence of a `[`: [`if(!path.includes("/index.md") && !path.includes("["))`](https://github.com/evidence-dev/evidence/blob/449a29056a2e1a4759a58a39e4eb7b3acbadd411/sites/example-project/src/pages/__layout.svelte#L50). Adding a similar check for `rootMDFiles` (`menu` items) seems to prevent the parameterised sidebar elements from showing up at the root level.

```js
for(let path in rootMDFiles) {
    pathEnd = path.replace('/src/pages/', '').replace(/^\.\//, '')
    if (path.includes('[')) continue; // change here
    menu.push({
        filename: pathEnd,
        label: pathEnd.replace(/\.md$/, '').replace(/_/g, ' ').replace(/-/g, ' '),
        href: "/" + pathEnd.replace(/^index\.md/, '').replace(/\.md$/, ''),
[...]

Issue #416, opened 2022-09-20
Sidebar: Sorting, Hiding, Custom Naming


archiewood:
## Feature Description
Multiple users have asked for more control over the sidebar. The sidebar currently works as follows:

- **All first and second level pages/directories are displayed** on the sidebar, with the exception of parameterized pages. Except for this [bug](https://github.com/evidence-dev/evidence/issues/410)
- **The name is taken from the markdown/directory file name:** it the Proper Case version eg : `my-page.md` -> My Page
- **Directories are shown before Pages**
- **Sort is alphabetical**

<img width="278" alt="image" src="https://user-images.githubusercontent.com/58074498/191283487-578f9399-da4b-4970-b619-d62ed0643b86.png">

## Goals of Feature
Still have sensible defaults (as above) but the ability to override the following:
- **Custom page sorting:** Determine the order in which pages / directories are shown
- **Custom page naming:** Give the page a name different to the filename. Support anything in Unicode if possible
- **Hide from sidebar:** Remove a page from display in the sidebar
- **(Maybe) Ability to totally hide sidebar:** This may fall into a separate feature

## Current Solution

Limited sorting is possible by prefixing file names with alphabetically sorted prefixes eg:
`1.charlie.md`
`2.alice.md`
`3.bob.md`

But these prefixes appear in the sidebar, which is not ideal, and directories will still sort before pages, irrespective of their alphabetical rank.

## Examples/Screenshots
Possible implementation options:
- Some kind of YAML / json file where you configure this for the whole project
- Use markdown frontmatter for each individual page

mcrascal:
We should use frontmatter now that it has been implemented in #294 

* Sort priority 
* Custom naming 
* Hiding -- docs should include a warning that pages with no hrefs directed to them won't be included in the build. 
* Defaults based on current behaviour (file is shown, named after the file, and sorted alphabetically) 

Hiding the side bar is out of scope.

archiewood:
This is done see: https://docs.evidence.dev/themes-and-layouts/

PR #460, opened 2022-11-15
fix: hide root level parameterized pages in sidebar

quoid:
This addresses the bug described in #410. 

It specifically addresses:

> **Expected Behaviour**
> - Not show is sidebar

For manual testing I added `[something].md` at root, along with `anotherthing.md` and verified that `[something]` does *not* show in the sidebar, but `anotherthing.md` does.

This does **not** address:

> - possibly crawl links in pages to determine which parameterized pages should be show in sidebar.

If anyone could elaborate on what the "crawl" functionality looks like maybe the PR can be expanded.

### DIFF
--- a/.changeset/wild-rings-unite.md
+++ b/.changeset/wild-rings-unite.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Prevent parameterized pages from showing in sidebar at root level, addresses #410
--- a/sites/example-project/src/pages/__layout.svelte
+++ b/sites/example-project/src/pages/__layout.svelte
@@ -13,6 +13,7 @@
 
 	for(let path in rootMDFiles) {
 		pathEnd = path.replace('/src/pages/', '').replace(/^\.\//, '')
+		if (path.includes('[')) continue;
 		menu.push({
 			filename: pathEnd,
 			label: pathEnd.replace(/\.md$/, '').replace(/_/g, ' ').replace(/-/g, ' '),
@@ -61,7 +62,7 @@
 				folderNameError: pathSplit[0].includes(" ")
 			})
 		}
-	} 
+	}
 
 	for(let path in levelTwoIndexFiles) {
 		pathEnd = path.replace('/src/pages/', '').replace(/^\.\//, '')
@@ -94,7 +95,7 @@
 	let folderHrefUri;
 	let indexFileCount;
 	let folderNameError;
-	
+
 	for(let i = 0; i < folders.length; i++){
 		contents = menu.filter(d => d.folder === folders[i]);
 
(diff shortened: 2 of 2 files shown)


archiewood:
Hey @quoid,

Thanks for this contribution! 
We've been meaning to get around to this for a while and haven't managed yet.

Tested this, and works for me. Happy to merge as is

Re crawl functionality. 
This is not something I have an idea on how to implement, just a rough idea:
- When svelte kit builds parameterized pages for production, it [crawls](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/prerender/crawl.js) to find the links in the projects, and then builds those routes.
- We could perhaps (and probably only for build, not dev) find all the links, and use those to generate links to parameterized pages in the sidebar.

quoid:
@archiewood 

Thanks for checking it out. I was thinking about this after the PR, and it might align with what you are suggesting with here:

> We could perhaps (and probably only for build, not dev) find all the links, and use those to generate links to parameterized pages in the sidebar.

If there is possibly a scenario where users *might* want to see their param pages, it might make sense to approach this differently.

Since all `*.md` pages can get pushed to the [`menu](https://github.com/evidence-dev/evidence/blob/449a29056a2e1a4759a58a39e4eb7b3acbadd411/sites/example-project/src/pages/__layout.svelte#L10) array, instead of skipping pages where `path.includes("[")`, rather add a new `key` in the `menu` object.

In the case of a root level param page, it could look something like:

```js
{
    filename: "[something].md"
    folder: undefined
    folderHref: undefined
    folderHrefUri: undefined
    folderLabel: undefined
    folderNameError: undefined
    href: "/[something]"
    hrefUri: "/%5Bsomething%5D"
    label: "[something]"
    nameError: false
    paramPage: true // new
}

```

And then in [`Sidebar.svelte`] (https://github.com/evidence-dev/evidence/blob/main/sites/example-project/src/components/ui/Sidebar.svelte) house the logic for whether or not to hide the param page based on a user setting. This could be something else to add to #416 if it aligns with your vision.

Either way, to address the bug in the interim, merging is good with me if for you.

archiewood:
Having had this discussion, I think this "show parameterized pages in sidebar" behaviour is a distinct new feature, and should probably be treated separately because:
- Your point about users perhaps (or perhaps not) wanting this functionality is a good one, so we might want to expose it as an option. 
    - This could live in a couple of different places. One natural place I can think of would be in markdown Frontmatter for the page, which we have yet to implement.
- The implementation detail for obtaining the page names does not seem totally straightforward

I'm going to go ahead and merge this.

Thanks very much! 🙌

archiewood:
Closes #410

quoid:
> markdown Frontmatter for the page, which we have yet to implement.

Looking forward to this!

Issue #290, opened 2022-05-27
rename sites/example-project 


mcrascal:
Over time, `sites/example-project` has become the central development workspace for: 

1. the components package
1. the internal API routes, and most of the other config shipped with the cli 

For the sake of clarity, I propose we rename it to `development-workspace`

hughess:
Noting that we'll need to change the name in such a way that the current git history is not lost. It seems we might be able to do it using the `git mv` command, but TBC.

archiewood:
This might be hard as I'm not sure git does this by design: https://web.archive.org/web/20160304045715/http://permalink.gmane.org/gmane.comp.version-control.git/217

ud3sh:
> Noting that we'll need to change the name in such a way that the current git history is not lost. It seems we might be able to do it using the `git mv` command, but TBC.

Never dealt with git mv myself. Sounds promising!

yukseltron:
Hello! Not sure what progress has been made on this. Just wanted to add my thoughts.

I also do not like it being stored in `example-project`. I feel it's an unintuitive name, however, I think Archie is right. It seems that it's a [controversial](https://stackoverflow.com/questions/35618507/git-sure-fire-way-to-move-rename-files-while-keeping-the-history#:~:text=35-,tl%3Bdr%3B%20no,-Longer%20version%3A%20In) and [risky](https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history#:~:text=178-,No.,-The%20short%20answer) move to change the name.

There doesn't seem to be a guaranteed way to pull this off without a hacky solution that may cause future problems.

I wonder then, is there a case where renaming it is worth losing its git history?

archiewood:
Coming back to this, there looks to be a [possible but scary](https://gist.github.com/emiller/6769886) way to do this. It basically involves a total rewrite of the fit history by the sounds of it!

I think we should come back and reach a conclusion on this. One thing is sure: if we ARE risking losing the git history, better to do it sooner than later!

archiewood:
It looks like this is a [new GitHub feature](https://github.blog/changelog/2022-06-06-view-commit-history-across-file-renames-and-moves/).

I think we can just drive ahead and rename this and we will at least see the history in GitHub

PR #463, opened 2022-11-16
Rename example-project

archiewood:
This addresses #290 

Notes on git history:
- Git blame / Git lens still works:

CleanShot 2022-11-15 at 21 04 54

- Github allows you to see the version history of the file (with the click of a button)

CleanShot 2022-11-15 at 21 07 06

@mcrascal @hughess @ud3sh was there anything else we were concerned about?

## Naming
Naming being one of the two hard things in CS, is `development-workspace` the best name we can find?
To me it's a bit general / un-descriptive. 

**What is it?**
- a svelte kit app in a folder

**What's in the folder?**
- c.60% components-related, of which
    - c.35% are components that make up the UI in an evidence project   
         - highest % 
         - name driver
    - c.20% are viz components that users can add 
    - c.5% helper functions to render components correctly
- c.30% markdown files which form the "Living spec" of features Evidence supports
- c.2% internal API routes
- c.2% base layout / theme files shipped with template
- c.2% config files shipped with template
- c.2% random bits of data we use in testing
- c.2% tests

**What's *not* in the folder:**
- anything to do with connectors
- CLI 
- docs
- preprocessing and markdown compiling
- types

Potential names based on the above:
- `project-framework`
- `template-workspace`
- `component-library`
- `frontend-components`

(or play mix and match with the first and last words)

### DIFF
--- a/.changeset/five-dragons-drop.md
+++ b/.changeset/five-dragons-drop.md
@@ -1,0 +1,7 @@
+---
+"@evidence-dev/evidence": patch
+"@evidence-dev/components": patch
+"evidence-docs": patch
+---
+
+renamed example project


archiewood:
This PR is now one huge merge conflict. Should close this and redo. Ideally when there are not too many other PRs open as it will probably cause merge conflicts in those

PR #465, opened 2022-11-16
[Feature]: Adds support for scrolling on datatable

afzalsayed96:
This is a quick experiment to show how UX on QueryDataTable can be improved by listening to vertical wheel event on `table` and updating the rows rendered accordingly.

https://user-images.githubusercontent.com/14029371/202231411-9e3c85ba-924c-438c-9a69-21370a21bdbe.mov


sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
```diff
@@ -83,12 +83,21 @@
...
+      index= Math.min(Math.max(0, index + Math.floor(event.deltaY/Math.abs(event.deltaY))), max)
+      slice()
+    }
 </script>
 
 <div class="results-pane" transition:slide|local>
 <div class="container" >
-
-  <table in:blur>
+  <table in:blur on:wheel={handleWheel}>
```
archiewood, line 100: I think it should probably only scroll when the users is over the body of the table, not the header.
probably just a case of moving the on:wheel into the `<tbody>`

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
```diff
@@ -83,12 +83,21 @@
...
     }
 
+    function handleWheel(event){
+      // abort if scroll is in x-direction
+      if (Math.abs(event.deltaX) > Math.abs(event.deltaY)) {
+        return
+      }
+
+      event.preventDefault()
+      index= Math.min(Math.max(0, index + Math.floor(event.deltaY/Math.abs(event.deltaY))), max)
```
archiewood, line 93: The scroll speed itself feel a little off when there are either lots of rows or not many:
- When there are lots of rows (eg 10,000) then you dont make enough progress for a roll of the wheel.
- When there are few rows (eg 20) then you instantly max out and end up at the bottom, it's difficult to move one row at once.

Would be nice if the scroll kind of accelerated: when you move the wheel a bit, then it advances a row or two, but if you start to drag lots (or if you are on a trackpad, swipe down a long way) then it speeds up and advances more rows.

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
```diff
@@ -83,12 +83,21 @@
       dataPage = updatedSlice
     }
 
+    function handleWheel(event){
+      // abort if scroll is in x-direction
+      if (Math.abs(event.deltaX) > Math.abs(event.deltaY)) {
+        return
+      }
+
+      event.preventDefault()
```
archiewood, line 92: Not a huge issue, but this abruptly stops a page scrolling when you are scrolling past a queryviewer table. This feels a bit unexpected as my page scroll is interrupted.
CleanShot 2022-12-13 at 09 41 36

It it possible to only prevent the page scroll when the scroll is started over the queryviewer table?

archiewood (commented):
This is a great feature, really wanted something like this for a while.

Got a couple of thoughts on implementation which I've added inline.

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
afzalsayed96: Does it feel off or does it feel constant? For one roll of a wheel `x` number of rows will be scrolled no matter what number of total rows are present in DataTable.

We could try to add certain logic to detect acceleration but is it really warranted, considering theres always the bottom slider to paginate more dramatically?

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
afzalsayed96: Have added a check to resume higher level scroll once first or last row is reached based on direct of scroll

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
afzalsayed96: Addressed

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
afzalsayed96: @archiewood slowing down the scroll speed is doable but I would like to _point_ out that system scroll speed is controlled by the OS settings. So, even if 50% reduced scroll speed feels normal to you, it would be too slow or too fast for someone else with different tracking settings.

Screenshot 2022-12-14 at 7 42 44 PM


You can test this out yourself by comparing the scroll speed with a scroll wheel of a mouse and the trackpad on the same device.

Screenshot 2022-12-14 at 7 45 48 PM



On the other hand, while tying a roll of the wheel to a percentage sounds a better solution, there is no way in my knowledge to get scroll delta in terms of scroll wheel units. So, I'm not sure how this could be implemented.

Do let me know, if I should proceed with 50% reduction of scroll speed.

afzalsayed96:
Note: This only adds scroll to `QueryDataTable`. Would need to apply the same changes to `DataTable` component

archiewood:
I think we could hold fire on that - we are currently redoing the DataTable component completely so I'm not sure if we'll even have a scroll bar in the new implementation

sites/example-project/src/components/ui/QueryViewerSupport/QueryDataTable.svelte
archiewood: I see. Sorry for my ignorance here. The thing that's a bit puzzling UX wise is that the scroll in the table is so much faster than the scroll up and down the page.

Perhaps this illustrates it better, I'm scrolling roughly the same amount here.

CleanShot 2022-12-14 at 09 34 17

If there's a way to make these feel more consistent, I'm all for it. I think 50% would probably be closer.

If you have other thoughts, let me know!

archiewood:
@afzalsayed96 what do you think of this?

Works pretty well with my mouse. Anything I should be watching out for here?

afzalsayed96:
This makes scrolling on my device way too dramatic with an external mouse:-

https://user-images.githubusercontent.com/14029371/207923198-18353f6a-2d5f-42b0-a297-90d8ff2a5e80.mov

So, I'm guessing it will vary from users to users.

archiewood:
Thanks, I'm going to revert and merge as before then

🙏 for all your help on this!

afzalsayed96:
Thanks @Archie for all your inputs. Enjoyed brainstorming on this.

PR #481, opened 2022-11-29
Improving Visual Hierarchy

archiewood:
## Why?
- We've had feedback that there isn't enough distinction between some of our H1 - H6 styles. 
- H3 and below are not very useful as they are not very visualy distinct from `<p>` text size wise.
- Sub-lists contain gaps after final bullet that breaks list flow

## What?
This contains some **reasonably major stylistic updates** to the default styling of Evidence projects:
- **More visual contrast between H1 - H6 using sizes**: This brings Evidence in line with [more typical web standards](https://html.spec.whatwg.org/multipage/rendering.html#sections-and-headings) for reading
- Removing margin after sub-lists
- Adding slight margin separating list items

## Questions

- This will change how H1 tags should be used. It will now probably not be suitable to use aside from the top of a page. (and instead use H2s) Does this matter?
- The overall change will increase the vertical space occupied by headings (especially H1, H2), which somewhat reduces the information density. Is this trade off worth it?

### Headings

Before:

CleanShot 2022-11-29 at 10 18 19@2x

After:

CleanShot 2022-11-29 at 10 19 13@2x

### Lists: 

Before: 

image

After:

image

### DIFF
--- a/sites/example-project/src/components/ui/Sidebar.svelte
+++ b/sites/example-project/src/components/ui/Sidebar.svelte
@@ -93,8 +93,9 @@ aside.sidebar {
 }
 
 .project-title{
-		margin: 0 0 0.3em 0; 
-		padding-top: 0.2em;
+	font-size: 1.2em;	
+	margin: 0 0 0.3em 0; 
+	padding-top: 0.2em;
 	}
 .sticky {
     position: sticky;
--- a/sites/example-project/src/components/ui/Sidebar.svelte
+++ b/sites/example-project/src/components/ui/Sidebar.svelte
@@ -95,7 +95,7 @@ aside.sidebar {
 .project-title{
 	font-size: 1.2em;	
 	margin: 0 0 0.3em 0; 
-	padding-top: 0.2em;
+	padding-top: 0.3em;
 	}
 .sticky {
     position: sticky;
--- a/sites/example-project/src/components/ui/Sidebar.svelte
+++ b/sites/example-project/src/components/ui/Sidebar.svelte
@@ -26,7 +26,7 @@
 <aside class="sidebar" class:open>
     <div class="sticky">
         <div class=nav-header>
-            <a href='/' on:click={() => open = !open}><h1 class=project-title>Evidence</h1></a>
+            <a href='/' on:click={() => open = !open}><div class=project-title>Evidence</div></a>
         </div>
         <nav>
 			{#if folderList}
@@ -92,6 +92,13 @@ aside.sidebar {
     border-right: 1px solid var(--grey-300);
 }
 
+.project-title{
+		font-size: 1.2em;
+		font-weight: bold;
+		color: var(--grey-800);
+		margin: 0 0 0.3em 0; 
+		padding-top: 0.3em;
+	}
 .sticky {
     position: sticky;
     top: 0;
@@ -309,6 +316,7 @@ span.alert-icon {
 	aside.open {
 		left: 0;
 	}
+
 
 	div.nav-footer {
 		display: none ;
--- a/sites/example-project/src/components/ui/Sidebar.svelte
+++ b/sites/example-project/src/components/ui/Sidebar.svelte
@@ -26,7 +26,7 @@
 <aside class="sidebar" class:open>
     <div class="sticky">
         <div class=nav-header>
-            <a href='/' on:click={() => open = !open}><div class=project-title>Evidence</div></a>
+            <a href='/' on:click={() => open = !open}><h2 class=project-title>Evidence</h2></a>
         </div>
         <nav>
 			{#if folderList}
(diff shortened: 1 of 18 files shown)


sites/example-project/src/components/ui/Sidebar.svelte
```diff
@@ -92,6 +92,13 @@ aside.sidebar {
     border-right: 1px solid var(--grey-300);
 }
 
+.project-title{
+		font-size: 1.2em;
+		font-weight: bold;
+		color: var(--grey-800);
+		margin: 0 0 0.3em 0; 
+		padding-top: 0.3em;
+	}
```
thomsebastin, line 101: Hi, the indentation seems slightly off. There needs to be a small space after `.project-title`. Also, I am nor sure if this decision is intentional, but it's better to stick with `rem` units as `em`'s are relative unit and very delicate to work with on a group setting. But if this was intentional and everyone understands, then cool.

thomsebastin (commented):
Added few observations. I just happen to stumble here. Hope you appreciate my help.

sites/example-project/src/components/ui/Sidebar.svelte
```diff
@@ -26,7 +26,7 @@
 <aside class="sidebar" class:open>
     <div class="sticky">
         <div class=nav-header>
-            <a href='/' on:click={() => open = !open}><h1 class=project-title>Evidence</h1></a>
+            <a href='/' on:click={() => open = !open}><div class=project-title>Evidence</div></a>
```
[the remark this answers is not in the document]
archiewood: Good idea. This is effectively the Logo / Home button.

I removed the H1 styling because with the change, H1 is now a lot larger and doesn't fit.

I'll mess around with H2 and see where I get

hughess:
Thoughts on the changes in this PR / answers to your questions:
- The bullet spacing changes are great
- I think the new headers take up too much space (reducing the information density as you mentioned). A few ideas: 
   - I think we can use font weight a bit more to distinguish the headers - e.g., we can have more bold sans-serif options in our headers (make H3 a smaller version of H2)
   - Use the old H2 as our new H3, then scale up H2 and H1 somewhere between where they were before and where they are in this PR

I took a look at our current headings in comparison to our visual elements (screenshots below) and have these thoughts:
- I agree that H3 and below are not visually distinct enough to be useable in most cases
- The boldness of the `BigValue` contents makes it hard for headers to take visual priority over them
   - I think the solution here is to make the BigValue text a bit thinner (maybe `font-weight: 500`) 
- We use a lot of grey shades throughout Evidence and sometimes it causes elements to visually blend together on the page
   - One way to make components and other elements stand out more is to make our header and base text colour darker (e.g., `color: var(--grey-999)`) 

## Charts
CleanShot 2022-12-19 at 10 58 05@2x

## Tables
CleanShot 2022-12-19 at 11 07 51@2x

## Big Values
CleanShot 2022-12-19 at 10 57 39@2x

## Text
CleanShot 2022-12-19 at 10 58 17@2x

## Blockquotes
CleanShot 2022-12-19 at 10 58 25@2x

hughess:
I don't think this makes sense to add now, but in the future if we need more options to visually separate elements on a page, we can use colours to create distinct sections of content. For example:

CleanShot 2022-12-19 at 11 57 25@2x

hughess:
Here are some comparisons after changing the BigValue thickness and making the base font colour darker:

CleanShot 2022-12-19 at 12 07 15@2x
CleanShot 2022-12-19 at 12 08 00@2x
CleanShot 2022-12-19 at 12 08 13@2x

archiewood:
I dig the BigValue thickness changes I think they look more elegant.
They are still big enough to grab they eye I think

Re the font colours for the headings. I suspect keeping them at slightly lighter grey is the right call. The Flawless Typography Checklist has a good piece on this:

image

It's subtle but effective:

image

archiewood:
# Changes
- @hughess I've sized down the h1 and h2s a bit. 
   - When I did that I worried the h3 wasn't quite distinct enough from the h2 on size alone. So I've italicized the h3s.
- Settings panels: rely on h1-h3 styles so have tweaked these to remain looking good.
- BigValue components: reduced font weight to 500

@mcrascal would be great to get your blessing on this as I know you did a lot of the original thinking!

## Before

image

## After

image

### All the headers
image

sites/example-project/src/components/ui/Sidebar.svelte
archiewood: We do have a more general need to standardize our approach on UI variables / UI design, but for the moment we're predominantly using `em` units.

mcrascal:
I really like it, definitely has my blessing. Thanks @thomsebastin also! 

The only issue I have is with the italics in the H3s. I know the H3s are really tricky, but I don't dig italics on that UI font. 

I'd propose taking off the italics and taking the size down by 0.1em. After that the H3 is still very heavy relative to the serif body and smaller headers, but it's also a lot smaller than the H2s. 

I'll open a PR with the change -- your call if you want to include it.

CleanShot 2023-01-03 at 16 43 55@2x

mcrascal (approved):

archiewood:
That looks great. Feel free to just commit to this branch, or I can rather than opening a new PR!

mcrascal:
PR against this branch is already open! #552

archiewood:
Seems fine to merge. Failing check is because we have changed the build command on `evidence-development-workspace`

PR #500, opened 2022-12-09
add max-width to BigValue titles

### LINKED ISSUE #494: Add a min-width to BigValue component
**Feature Description**
- Set a suitable min-width for the BigValue component

**Goal of Feature**
- currently when all bigvalue text or numbers occupy the same width, they render cleanly in the UI:
image

- But if the titles are different lengths, they can look quite messy as they are different widths
   - This could be fixed with a suitable min-width. This would also allow the BigValue components to get wider if they needed

yukseltron:
Resolves #494 

- Add a max-width to BigValue comparison titles
- Left align title to end of Value rather than start

Before:
Screen Shot 2022-12-09 at 12 41 16 AM

After:
Screen Shot 2022-12-09 at 12 40 40 AM

### DIFF
--- a/.changeset/purple-cherries-begin.md
+++ b/.changeset/purple-cherries-begin.md
@@ -1,0 +1,6 @@
+---
+"evidence-docs": minor
+"@evidence-dev/components": minor
+---
+
+Add maxWidth and minWidth style props to BigValue component. Default set for minWidth to give smaller BigValues more room.
--- a/sites/docs/docs/features/charts/big-value.md
+++ b/sites/docs/docs/features/charts/big-value.md
@@ -16,7 +16,7 @@ hide_table_of_contents: false
     comparison='monthly_growth' 
     sparkline='date'
     comparisonTitle="Month over Month"
-    --max-width='1em'
+    maxWidth='10em'
 /> 
 ```
 
@@ -37,8 +37,8 @@ Multiple cards will align themselves into a row.
 * **title** - (Optional) title of the card. Defaults to the title of the value column.
 * **comparisonTitle** - (Optional) text to the right of the comparison. Defaults to the title of the comparison column.
 * **downIsGood** - (Optional) if present, negative comparison values appear in green, and positive values appear in red. 
-* **--min-width** - (Optional) overrides min-width of component, 25% by default. 
-* **--max-width** - (Optional) adds a max-width to the component, none by default. 
+* **minWidth** - (Optional) overrides min-width of component, 18% by default. 
+* **maxWidth** - (Optional) adds a max-width to the component, none by default. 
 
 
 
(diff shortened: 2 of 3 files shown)


archiewood:
Hey, sorry if this is unclear in the issue. 
The issue we're most worried about here is when the text is "too short" not "too long".

This stackblitz hopefully is a better explanation: https://stackblitz.com/edit/evidence-patpt5?file=pages%2Findex.md

yukseltron:
Thanks for the stackblitz, that helped clear things up more.

This time, I applied the min-width to the container instead of just the comparison title.

Before:
Screen Shot 2022-12-09 at 10 40 31 AM

After:
Screen Shot 2022-12-09 at 10 42 00 AM

However, I wonder if adding a max-width would still be ideal for longer titles/numbers? 

Example of long title with min-width and max-width:

Screen Shot 2022-12-09 at 10 45 09 AM

archiewood:
I think I dislike the titles spilling across lines as it ruins the horizontal alignment of the BigValue numbers.
I think in that case we should probably just let the thing run. 

From a user perspective they then have a choice:
- Either shorten the title so that the components are evenly spaced
- Or, if you need the longer title, you keep it

We could expose the prop to the user to allow them to override the min-width with a larger value in cases where they want to have generally longer titles, but still keep alignment. 

WDYT?

yukseltron:
I like the idea of exposing min-width as a prop. I think it gives more control to the user. I can make the min-width I set as a default for it.

I also think adding a max-width as an optional prop would be nice too. I think it's better to let the user decide how they want to format their titles and worry less about how short the title should be.

archiewood:
^ Sounds perfect 👍🏼

yukseltron:
Ok, I added style props for both min-width and max-with. The defaults are 25% and none respectively. As well, I updated the documentation for BigValue.

archiewood:
Excellent

**I've done:**
- Reduced the min-width default to 18%: allows you to stack 4 in wider viewports, given the default padding.
CleanShot 2022-12-09 at 13 06 12

**Can you:**
- Add a changeset: https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset

**Two related questions about the props:**
- Defining the min-width and max-width props at ::root - I'm not a css expert here, but does this mean they are global styles that would be accessible to anything in the page?
- I see to edit the min and max width, you have to add them as css style props `--max-width`, rather than js style props `maxWidth` - for the rest of our components, when we expose props we are using js type props, even if they are for styling purposes. For example see the [exported props in the DataTable](https://github.com/evidence-dev/evidence/blob/main/sites/example-project/src/components/viz/DataTable.svelte#L16)

yukseltron:
Ok, I've added a changeset and switched it to use JS props. Docs also updated accordingly.

Good catch about the use of `:root`! I incorrectly thought Svelte also scopes it to just that component.

By changing it to JS style props,  the `:root` reference is no longer needed.

PR #502, opened 2022-12-09
Fix chart rendering issue

afzalsayed96:
Fixes https://github.com/evidence-dev/evidence/pull/450#issuecomment-1340363141

### DIFF
--- a/.changeset/metal-mangos-judge.md
+++ b/.changeset/metal-mangos-judge.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Fix charts rendering issue in webkit browsers
--- a/sites/example-project/src/components/modules/echarts.js
+++ b/sites/example-project/src/components/modules/echarts.js
@@ -442,7 +442,13 @@ export default(node, option, renderer) => {
     
     let resizeObserver
     const containerElement = document.querySelector('div.content > article')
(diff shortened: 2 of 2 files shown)


archiewood:
Hey @Afzal Sayed - thanks so much for the PR 

I'm not sure how to test if this will solve this issue as we don't have a clean repro. Were you able to replicate the issue?
- TBH this is an issue of ours - would be solved by a better testing framework or alike.

Could you?
- Provide a gif or something you can post as a before and after? 
- Or repro the issue somewhere in the dev workspace ./sites/example-project
- Or failing the above options, perhaps you can explain why you expect it to work?

afzalsayed96:
Hey @archiewood,

Yes, I was able to replicate the issue and verify that this fixes it -

**Before**

https://user-images.githubusercontent.com/14029371/206851369-974600be-d021-4a55-be63-e29565aab709.MP4

**After**

https://user-images.githubusercontent.com/14029371/206851362-08542cca-689d-4d7c-921d-e1bafeddc55c.MP4

However, I was not able to reason the root cause behind this behaviour as it only seems to occur in webkit based browsers

archiewood:
Okay, great! Thanks

archiewood:
For some reason our changeset bot is not picking up this new changeset. Let me try closing and reopening

PR #531, opened 2022-12-15
add separate download data option to charts

### LINKED ISSUE #528: Download data from charts
**Feature Description**
- Ability to quickly export data from charts
- Should preserve data structure eg series' are pivoted

**Goal of Feature**
- Data workflows in other tools: excel, google sheets, email etc

**Current Solution**
- Download from queries. But queries are not always near charts, may have too many columns, and may not be pivoted correctly

**Alternative Ideas**
- Also nice if you could copy the data to clipboard in a click

yukseltron:
### Description
Resolves #528 

Hopefully, this is what was requested.

- DownloadData now has text as prop ("Download Data", "Save Image", etc). Might be good to rename (DownloadButton)
- Data added as prop to ECharts component
- Echarts now has download option for Chart PNG and CSV data

Before:
 
Screen Shot 2022-12-14 at 7 47 36 PM

After:

https://user-images.githubusercontent.com/14843458/207746387-3914617d-e623-474d-91bd-a65443d3fd4f.mp4

### DIFF
--- a/sites/example-project/src/components/viz/ECharts.svelte
+++ b/sites/example-project/src/components/viz/ECharts.svelte
@@ -36,7 +36,9 @@
 
 <EchartsCopyTarget {config} {height} {width} {copying}/> 
 
+{#if data}
 <DownloadData text="Download data" {data} class=download-button />
+{/if}
 <DownloadData text="Save image" class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}}>
   <svg xmlns="http://www.w3.org/2000/svg" width="13" height="13" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="3" y="3" width="18" height="18" rx="2"></rect><circle cx="8.5" cy="8.5" r="1.5"></circle><path d="M20.4 14.5L16 10 4 20"></path></svg>
 </DownloadData>
--- a/sites/example-project/src/components/viz/ECharts.svelte
+++ b/sites/example-project/src/components/viz/ECharts.svelte
@@ -37,7 +37,9 @@
 <EchartsCopyTarget {config} {height} {width} {copying}/> 
 
 <DownloadData text="Download data" {data} class=download-button />
-<DownloadData text="Save Chart" class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}} />
+<DownloadData text="Save image" class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}}>
+  <svg xmlns="http://www.w3.org/2000/svg" width="13" height="13" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="3" y="3" width="18" height="18" rx="2"></rect><circle cx="8.5" cy="8.5" r="1.5"></circle><path d="M20.4 14.5L16 10 4 20"></path></svg>
+</DownloadData>
 
 </div>
 
--- a/sites/example-project/src/components/viz/ECharts.svelte
+++ b/sites/example-project/src/components/viz/ECharts.svelte
@@ -9,6 +9,8 @@
     export let height = '291px'
     export let width = '100%'
 
+    export let data;
+
     let downloadChart = false;
     let copying = false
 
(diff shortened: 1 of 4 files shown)


archiewood:
Thanks for contributing on this! 

There are actually some relatively specific customer requirements here which I didn't put into the issue, apologies:

1. The data needs to be just the columns used in the chart, not the entire contents of the query.

2. The data needs to retain the "shape", ie pivots it has in the graph, (if relevant),

For example if you had a ˋ<BarChart data={myquery} x=year y=sales series=category/>ˋ from a query that has columns, year, country, category, sales, the download would have data structure:

| Year | Category A | Category B |
| -- | -- | -- |
| 2010 | $10,000 | $20,000 |
| 2011 | $11,000 | $21,000 |

3. The download format needs to be csv

sites/example-project/src/components/viz/ECharts.svelte
```diff
@@ -34,7 +36,8 @@
 
 <EchartsCopyTarget {config} {height} {width} {copying}/> 
 
-<DownloadData class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}} />
+<DownloadData text="Download data" {data} class=download-button />
+<DownloadData text="Save Chart" class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}} />
```
archiewood, line 40: Let's be consistant here with capitalization vs "Download data". 

Also suggest we say "Save image", which I think makes it 100% explicit what the format is. 

I wonder also if there is a better icon we can find here. Perhaps something like this: image 

There are probably some options in the same icon library as the download icon

mcrascal:
@archiewood is maintaining the shape of the data a hard requirement? 

As a user, I would have a strong preference for the "long" output (as it is now), which is suitable as an input for excel pivot table and/or notebook tool.

archiewood:
Fair point. I'm not 100% sure but I guess it would depend on the background of the user. 

I think "data people" would typically prefer "long" data, but would guess that most other users would expect it pre pivoted?

sites/example-project/src/components/viz/ECharts.svelte
yukseltron: Changed to 'Save image' and applied the image icon.

https://user-images.githubusercontent.com/14843458/207907624-29a166bc-79b4-4ebf-93c7-6536a7da83e4.mov

archiewood:
In terms of the UI, I think this works well.

In terms of the data returned in the csv, I worry this is not predictable enough for users. 

Examples:
1. Two charts that contain the same chart type return data in different formats:
   - [Multi series line](https://evidence-development-workspace-qfbf2306l-evidence-dev.vercel.app/charts/line-chart#Multi-Series%20Line1)
   - [Multiple column line](https://evidence-development-workspace-qfbf2306l-evidence-dev.vercel.app/charts/line-chart#Muliple%20y%20Column%20Line2)
2. Data includes columns that are not shown in the charts, but are in the query
- eg BarCharts on [this page](https://evidence-development-workspace-qfbf2306l-evidence-dev.vercel.app/charts/stacked-charts)

I think we might need a way to make the data a more predictable structure for users. Not fully sure on the approach yet given the data queries return can be in different shapes.

Rough ideas on an approaches:

1. Start from the {data} object returned from a query
- Get the data object from the query
- Filter to only grab columns shown in chart. This might be tricky as sometimes columns are specified in params, and sometimes they are inferred.
- (Possibly) do some data transformations to make the data always "long" (one record per row), or always wide (multiple columns with data, when multi series)

2. Start from the data passed to the chart
- Not sure what this looks like, but logically the transformations and filtering must have already been applied!
- But possible that the structure is unhelpful!

sites/example-project/src/components/viz/ECharts.svelte
```diff
@@ -34,7 +36,10 @@
 
 <EchartsCopyTarget {config} {height} {width} {copying}/> 
 
-<DownloadData class=download-button downloadData={() => {downloadChart = true; setTimeout(() => { downloadChart = false}, 0);}} />
+<DownloadData text="Download data" {data} class=download-button />
```
archiewood, line 39: There is probably one adjustment to make here, which is to wrap the DownloadData in an if block, checking if there is in fact a data object is passed to ECharts.

The reason for this is that it is possible to create a chart by passing a config object directly to `<ECharts/>` rather than passing a `data` object. In this case there is no data object, and so the button does nothing.

Example [here](https://evidence-development-workspace-qfbf2306l-evidence-dev.vercel.app/charts/custom-charts) which is a bit contrived, but any of the [custom charts in the docs](https://docs.evidence.dev/features/charts/custom-charts) would also have this issue

archiewood (approved):

archiewood:
Once again TY @yukseltron!

PR #532, opened 2022-12-15
Include how to change host in docs

visch:
### Description
Needed a way to update my host as I run in a VM and this is an easy way for me to host my code locally. No idea if this is valuable or not for you all or if it's how you want to format this. Needed to figure these steps out myself and wanted them documented :D 

### Checklist
- [na ] For UI or styling changes, I have added a screenshot or gif showing before & after

### DIFF
--- a/.changeset/bright-crews-move.md
+++ b/.changeset/bright-crews-move.md
@@ -1,0 +1,5 @@
+---
+"evidence-docs": patch
+---
+
+Hosts installation instructions added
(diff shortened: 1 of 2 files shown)


archiewood:
Good point.

I think it would probably be simpler to give users command line options, rather than them changing their package.json? Eg a section like this.

# Development Server Options

You can change how the dev server runs by adding "arguments". To do this you add `--` and then the arguments you need:
```
npm run dev -- --[setting] [option]
```

## Run from an alternative host
```
npm run dev -- --host 0.0.0.0
```

## Run using an alternative port
```
npm run dev -- --port 4000
```

Would that have helped? If so, happy to make the edits later!

archiewood:
Going to be easier to just do this from a fresh branch, so going to close this one!

Issue #266, opened 2022-05-17
Redundant queries executed when building parameterized pages 


mcrascal:
**Bug Description**
When running `npm build` on a project that includes a parameterized page, queries run from the against the db for each instance of the parameterized page, causing long build-times and wasted db resources for some projects. 

**Expected Behavior**
Run the query once, and serve the remaining queries from the cache for each page that's being rendered.

archiewood:
Just ran across this again today. Was going to open an issue for it!

PR #536, opened 2022-12-17
Use query cache in build mode

archiewood:
### Description
Fixes: #266 

### Checklist


archiewood:
@ud3sh I have a question on Evidence Cloud here:
- This implementation basically allows caching of queries during build, which previously did not happen: every query was run fresh in build mode.
- I'm slightly unclear on what build processes would do in Evidence Cloud in an example like this:
    - Someone sets up their cloud project to build every 10 minutes. 
    - In principle, cached queries would not be invalidated on this timeframe, so if they existed still then then queries would use previously cached results. (This is not the desired behaviour.)
- This would only be a problem if the caches are persisted between builds

So the question is: are files persisted between builds, or do they run "clean" each time?

I there is some persistence, I can probably edit this to purge the cache at the start of each build, but just wanted to check@

archiewood:
Also if you had any insight on how Netlify / Vercel etc do the above here I guess that would also be relevant

PR #550, opened 2023-01-03
Make clickable elements buttons to improve accessibility

archiewood:
### Checklist


archiewood:
This looks to be the postgres tests that are failing. This is odd since this PR does not touch the postgres connector (or any connector, for that matter)

It looks like supabase pauses their free databases after 1 week of activity: https://supabase.com/pricing

I've restarted it which I think will fix the above, but we probably need a better solution!

PR #553, opened 2023-01-04
Table image and url support

archiewood:
### Description
- Possible way of adding URL and link support
- But would be good to be able to allow adding arbitrary <slot> which would be more powerful than this

### Checklist

After:
image

### DIFF
--- a/sites/example-project/src/components/viz/Column.svelte
+++ b/sites/example-project/src/components/viz/Column.svelte
@@ -17,6 +17,7 @@
     export let title = undefined;
     export let align = undefined;
     export let img = false;
+    export let link;
     export let height;
     export let width;
     if(align === "centre"){ align = "center"};
@@ -28,6 +29,7 @@
         img: img,
         height: height,
         width: width,
+        link: link,
     }
 
     props.update(d => {
--- a/sites/example-project/src/components/viz/Column.svelte
+++ b/sites/example-project/src/components/viz/Column.svelte
@@ -16,12 +16,18 @@
 
     export let title = undefined;
     export let align = undefined;
+    export let img = false;
+    export let height;
+    export let width;
     if(align === "centre"){ align = "center"};
 
     let options = {
         id: id,
         title: title,
         align: align,
+        img: img,
+        height: height,
+        width: width,
     }
 
     props.update(d => {
--- a/sites/example-project/src/components/viz/DataTable.svelte
+++ b/sites/example-project/src/components/viz/DataTable.svelte
@@ -342,6 +342,11 @@
                       width: {column.width};
                       "
                   />
+                  {:else if column.link}
+                    <a 
+                        href={row[column.link]}>
+                        {formatValue(row[column.id], columnSummary.filter(d => d.id === column.id)[0].format)}
+                    </a>
                   {:else}
                   {formatValue(row[column.id], columnSummary.filter(d => d.id === column.id)[0].format)}
                   {/if}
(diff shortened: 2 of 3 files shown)


archiewood:
Re arbitrary slot support, the closest I got was by adding a html prop

archiewood:
This has quite a few downsides:
1. It's syntax is kinda weird (though could work on that a bit)
    - What I'd expect:
        ```svelte
        <DataTable 
            data={tableq}>
            <Column>
                 <!-- flag is a column of tableq containing the url to an image -->
                 <img src={flag} height=50px/>
            </Column>
        </DataTable>
        ```

    - What this gives you:
        ```svelte
        <DataTable 
            data={tableq}>
            <Column
                 id=flag
                 html={`
                  <img src={flag} height=50px/>
                  `}
             />
        </DataTable>
        ```
2. Doesn't support putting in svelte components, only vanilla HTML

hughess:
This is cool! Making the slot work as expected would be tricky because of the way `Column` works - it's basically just updating the store used by `DataTable` rather than actually getting inserted into the table.

Here are a couple of ideas on syntax:

### contentType Prop
Similar to your working examples - add props for each type of content we want to support
```html
<DataTable data={my_query}>
   <Column id=country />
   <Column id=flag contentType=image someImageProp=true />
   <Column id=url contentType=link openInNewTab=true />
</DataTable>
```

### Content Type Components
We create a component for each content type we want to support within tables. These components would need to update a store that `Column` could access before passing it to `DataTable`
```html
<DataTable data={my_query}>
   <Column id=country />
   <Column id=flag>
      <Image someImageProp=true />
   </Column>
   <Column id=url>
      <Link openInNewTab=true />
   </Column>
</DataTable>
```
If we were to go down this path, we might want to name the components more specifically so they wouldn't be confused with components that could be used outside of a table cell. E.g., `CellImage`

hughess:
Another thought on this - should we support clickable images? 

In that case, we'd need to support using `img` and `link` at the same time.

archiewood:
Would this be:
1. Click the image and get a larger version of the image?
1. Click the image and go through to another Evidence page, or external URL?
1. Something else

If it was 1, you don't need a unique prop, as it's the same URL as the img?
If it was 2, you could solve this if you make the whole row clickable using the link field? (though maybe that needs to be a prop - `rowLink=True`)

If it's something else, maybe I'm thinking about this wrong?

hughess:
I was thinking of 2 - I think a row link would solve that! I have a prototype of row links working in another branch and I'll share that in a PR soon.

So that means we'll be good with treating images and links as separate items in table. The last thing I'm trying to figure out is how these will fit into the features we plan to add in the future within table cells, like heatmaps, bar charts, progress bars, etc.

Because these new features wouldn't overlap with images and links, I'm leaning towards having a single prop `contentType` to define the type within each column. For example:

```html
<DataTable data={countries} rowLinks=country_url >
   <Column id=flag contentType=image height=50px />
   <Column id=country_url contentType=link linkLabel=country_name openInNewTab=false />
   <Column id=sales contentType=bar />
   <Column id=orders contentType=heatmap />
   <Column id=yoy_growth contentType=progressBar />
</DataTable>
```

That feels a bit suboptimal for the link column, but convenient for the other column types.

We could also call that prop `type` or something similar to make it easier to remember, but we'll also need to avoid confusion with data types.

This structure might also allow us to support html in the future (`contentType=html`), but I'm not sure how that would work.

@archiewood what do you think of this syntax?

archiewood:
I like it. For the image we need to include the ability to add an alt tag I think. Perhaps we make it an additional optional prop? 

If users don't add the alt tags then svelte will complain, at which point they can choose to add them if they want.

hughess:
Nevermind, I see you have a default for alt tag in your implementation already!

mcrascal:
Moving discussion/ tracking to #570 for the meantime.

Issue #540, opened 2022-12-20
Upgrade to SvelteKit 1.0


mcrascal:
**Feature Description**
Evidence is currently using a pre 1.0 version of Svelte Kit (SK). The incredible Svelte team recently [released SK 1.0](https://github.com/sveltejs/kit/pull/8158), and we're super excited to start taking advantage of the many new features and improvements therein. 

**Benefits of upgrading** 
We won't realize all of these benefits "out of the box" but they all get unblocked by doing this. 

1. Ease of maintenance and development for ourselves and our contributors.
2. Performance improvements a la #490
3. Enable us to remove a large amount of data processing logic from components and our markdown preprocessor by returning a javascript object from our data endpoints, instead of serialized JSON. 
4. Enable us to use the latest vite plugins, many of which implement functionality we are currently maintaining ourselves. 

We are going to aim to implement the upgrade initially with as little impact as possible, but this is the first step in a broader refactoring of the code base to make Evidence much easier to contribute to. 

**On Routing**

Svelte Kit has changed their routing system since we first started using it. Now in Svelte Kit, each route corresponds to a directory containing a `+page.svelte` file `about/+page.svelte`, not `about.svelte`. Evidence is not going to to adopt this convention, at least not as part of this update. One markdown file will continue to correspond to one route: `about.md`, or `about/index.md`.  

While we think the convention makes a lot of sense for Svelte Kit, the benefits are much harder to see for our users, given that we write endpoints, and layouts on their behalf. Our thinking might change here, but it's pretty unlikely.

csjh:
Made a start at this in https://github.com/csjh/evidence , couldn't figure out some Vite externalization bits and pieces but I'll take another stab tomorrow

mcrascal:
@csjh this is Epic! I'm just getting back into the swing after the break, but this looks like a really strong starting point. I'll have time to dig in in the morning. Thanks!

csjh:
Redid it more gradually which helped stomp out some of the issues but still can't get svelte-tiny-linked-charts to play happy. I think it has something to do with a missing "type": "module" in it's package.json. Current fix is just putting it in src/components/viz/_LinkedChart. Should work fine otherwise though:
https://github.com/csjh/evidence/tree/gradual

mcrascal:
@csjh this looks fantastic -- I've taken a spin through it, and have opened a PR into the feature branch here. 

[...]

PR #564, opened 2023-01-10
Gradual

mcrascal:
### Description
Bringing the fantastic work on #540 from @csjh into our feature branch here to keep moving it forward.


csjh:
Looks like the test fails are because SvelteKit dropped Node 14 support. Maybe swap to testing 16, 18 instead of 14, 16?

PR #567, opened 2023-01-12
Add support for images, links, and row links in DataTable

hughess:
### Description
This PR adds:
- Support for content types within DataTable (starting with images and links) - thanks to @archiewood 
- Row links in DataTable

### Content Types
A new prop called `contentType` is now included in the `Column` component. This lets you specify how you want the information in your column to be displayed. We are starting with images and links, but will be adding viz options in the future as well.

#### Images
```html
<Column id=flag contentType=image height=30px />
```

Props specific to images:
- `height`: height of the image in pixels (e.g., `30px`)
- `width`: width of the image in pixels
- `alt`: overrides the alt tag of the image. By default, the alt tag is the name of the image file (excluding the file extension). To override the alt tag, `alt` accepts a column name in your table

**Example**
CleanShot 2023-01-13 at 12 09 06@2x

#### Links
```html
<Column id=country_url contentType=link linkLabel=country_name openInNewTab=true />
```

Props specific to links:
- `linkLabel`: the text to show for the link. Can be a string (e.g., `Link &rarr;`) or a column name in your table (e.g., `country_name`)
- `openInNewTab`: true | false - whether the link should open in a new tab when clicked (default is false; opens in current tab)

**Example of column linkLabel**
CleanShot 2023-01-13 at 12 13 23@2x

**Example of string linkLabel**
CleanShot 2023-01-13 at 12 14 10@2x

### Row Links
```html
<DataTable data={countries} rowLink=country_url />
```

A row link makes each row of your table a clickable link. The `rowLink` prop receives a column name in your dataset which contains the link to use for each row. When you add row links, rows will appear clickable and with a highlighted background on hover:

CleanShot 2023-01-13 at 12 06 51

### Checklist


sites/example-project/src/components/viz/DataTable.svelte
```diff
@@ -319,7 +332,34 @@
...
                       text-align: {column.align};
-                  ">{formatValue(row[column.id], columnSummary.filter(d => d.id === column.id)[0].format)}</td>
+                      height: {column.height};
+                      width: {column.width};
+                  ">
+                  {#if column.contentType === "image"}
+                    <img 
+                    src={row[column.id]} 
+                    alt={row[column.id]} 
+                    style="
```
archiewood, line 342: So this is always going to use the url as the alt tag for the image.

Do we want to either:
- process the regex from the URL, perhaps using anything after the last slash
- allow users to override the alt tag with another prop?

hughess: Good call - I've just added a prop for `alt`.

Regex seems like a pretty good idea for that. If we do that, should we include the file extension?

hughess: I've added regex to take the filename after the last slash and strip off the extension

hughess:
A few things left to figure out:
- There's some strange behaviour when an image and a long url are included in the same table - the image gets squished even if height and width are supplied. Going to investigate this
- Naming of props
- Row links - these require you to have a column in your dataset which contains the URL. For internal page navigation in an Evidence app, this means you would need to have your page link inside your SQL query, which may not be ideal. The upside of having it in SQL is that it's easier to concatenate the correct string (vs. figuring out how to set up props that can accommodate that type of concatenation)

### Prop Names
- `contentType` vs. `type` vs. other?
- `linkLabel` vs. `linkText` vs. other?
- `rowLink` vs. `rowLinks` vs. other?

mcrascal:
This is awesome. 

> Row links - these require you to have a column in your dataset which contains the URL. For internal page navigation in an Evidence app, this means you would need to have your page link inside your SQL query, which may not be ideal. The upside of having it in SQL is that it's easier to concatenate the correct string (vs. figuring out how to set up props that can accommodate that type of concatenation)

Agreed, constructing the link in the query seems like a good enough starting place, plus any future syntax we offer to concatenate query columns with text in a prop would likely still be compatible with that approach. 

> naming props 

Throw my votes in here: I like just `link` as the prop name for a row link. I like `linkLabel`, and `contentType` for the others. 

Other questions: 

* In the case of row links, do we exclude the column containing the url from the table? Given that users are going to construct row links in their SQL, I think they'd want to exclude the column containing the URL from the resulting table being displayed. If we support a prop driven approach to constructing URLs in the future, they might want to include those columns (e.g. an ID column). I think this might point toward offering an `exclude = [array, of, cols_to_exclude]` prop at some point, so that users don't have to add `<Column />` props for everything except that one URL column.     
* Do we need to url encode the provided string for users before turning it into a link?

archiewood:
I _think_ I disagree on hiding the rowLink column by default. 

I think excluding the column containing the URL from the table might make using the table for navigation much more verbose for simple cases.

See this stackblitz example: https://stackblitz.com/edit/evidence-chvlky?file=pages%2Findex.md

Perhaps we could implement alongside a showLinkColumn approach, (or a `exclude = [array, of, cols_to_exclude]`) which gives the user the choice of whether to hide the column or not?

hughess:
I've checked into the URL encoding and it does work without any changes, so I think that's good to go as is.

I think I'd want to hide the row link column in most cases. @archiewood I updated the stackblitz to show the URL after concatenating in SQL: 
https://stackblitz.com/edit/evidence-vxfiuw?file=pages%2Findex.md

I like the idea of having options for excluding columns. Maybe we can start with the single prop (show or hide link column) and expand to the excludeCols array approach afterwards. I'm leaning towards having that default to hiding the link column

PR #568, opened 2023-01-13
Fix leaking CSS from Settings page into components

### LINKED ISSUE #455: CSS from settings page leaking into components in client-side navigation
**Bug Description**
If you navigate from the Settings page to a regular markdown page using client-side navigation, some CSS from the settings page takes precedence over component-specific styling on the markdown page. The issue does not appear a page refresh or when using server-side navigation.

**Severity**

**Expected Behavior**
CSS from Settings page shouldn't impact components on other pages.

**Steps To Reproduce**
1. Start on a page that has a query on it - open the query viewer data table and look at the data scrubber
2. Navigate to the Settings page using client-side navigation
3. Navigate back to the previous page by clicking the page link in the sidebar
4. The scrubber in the query viewer will now appear with a grey border and it will be in a slightly lower position

CleanShot 2022-11-10 at 21 17 28

**Info for Potential Solution**
- The formatting applied (e.g., the grey border in the data scrubber example) is coming from the `format-grid.css` file in the `src/components/ui/Formatting` directory
- If I comment out the formatting in that file and repeat the steps to reproduce above, the issue disappears
- It's not clear to me how the styling is getting applied

hughess:
### Description
Fixes #455 

Before, if you navigated to the Settings page and then to another page in your project, some CSS from Settings would leak into the page, causing styling issues with components (see examples for Query Viewer and DataTable below).

This was caused by an issue with importing a css file within components, as described by @quoid [here](https://github.com/evidence-dev/evidence/issues/455#issuecomment-1317853394)

#### Before
CleanShot 2023-01-13 at 15 39 17@2x

#### After
CleanShot 2023-01-13 at 15 39 58@2x

#### Before
CleanShot 2023-01-13 at 15 39 24@2x

#### After
CleanShot 2023-01-13 at 15 40 04@2x

### Checklist


sites/example-project/src/components/ui/Formatting/BuiltInFormatGrid.svelte
```diff
@@ -1,5 +1,4 @@
 <script>
-  import "./format-grid.css";
   import { defaultExample, formatExample } from "$lib/modules/formatting";
```
ud3sh, line 2: I assumed importing css would result in svelte localizing the classes to the importing component.  Doesn't seem to work that way. I think copying is fine for now but we should figure out the correct way to do it at some point.  I suspect this won't work either  `<style src='./format-grid.css'>`

ud3sh (approved):

PR #589, opened 2023-01-23
Add "Fail on build error" to the project

Winterhart:
### Description

This is an attempt to add: "Fail on build error" without disrupting existing projects.

- New CLI command
- New `package.json` command
- New flag `VITE` to propagate the settings
- Rework on `pre-processor` to fail on error in `strict` mode
- Shared `strictBuild` flag

### Checklist

### Follow up Task

Make a PR in the `template` to bump the version and include the new CLI command

### DIFF
--- a/sites/example-project/src/components/viz/context.js
+++ b/sites/example-project/src/components/viz/context.js
@@ -1,4 +1,5 @@
 const configKey = Symbol()
 const propKey = Symbol()
 
+export const strictBuild = (import.meta.env.VITE_BUILD_STRICT === 'true')
 export {configKey, propKey}
--- a/.changeset/rare-donuts-hope.md
+++ b/.changeset/rare-donuts-hope.md
@@ -1,0 +1,8 @@
+---
+"@evidence-dev/db-orchestrator": minor
+"@evidence-dev/evidence": minor
+"@evidence-dev/preprocess": minor
+"@evidence-dev/components": minor
+---
+
+New capability to fail build if there are errors in components or queries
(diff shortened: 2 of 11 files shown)


mcrascal:
@Winterhart this is looking really nice. 

I know this is still WIP, but a couple points to take a look at: 

1. I think we should consider making the default mode strict, and offering a flag for "relaxed" or whatever we want to call "not strict". I would rather make "not strict" the exception for end users. 
2. There are a few other components which can also be in an error state: datatable, value, big value.  
3. [Chained queries](https://docs.evidence.dev/features/queries/query-chaining) (SQL queries that reference other queries on the page)  can also generate errors before their SQL reaches db orchestrator. For example, if there is a circular reference between queries. This is evaluated [here](https://github.com/evidence-dev/evidence/blob/7bc9e4c1660fb562252c15a774c7f5ba33244bfe/packages/preprocess/index.cjs#L168) in pre-process.

Winterhart:
Thanks, @mcrascal for the first mini-review

1. I will keep the `VITE...` build strict flag but will add some improvement story to move to `strict` by default on build
2. 💯 will add the strict check to the other components
3. Will have a look 👀 at the `chained` query

sites/example-project/src/components/viz/context.js
```diff
@@ -1,4 +1,5 @@
 const configKey = Symbol()
 const propKey = Symbol()
```
ud3sh, line 3: cool, never seen this way of getting env vars before

ud3sh:
In the longer term, it seems that we can do few things to generalize component  `$: ` handling (among others things in component initialization), so that things like error handling can be in one place.

PR #597, opened 2023-01-26
Add more error handling in DataTable

Winterhart:
### Description

Adding error if `<Column />` `id` is not found in the table or if it doesn't contain any `id`. 

- The `error` is also sent to the `console` in `dev` mode 
- The `error` will fail the build in `strict` mode.

580-column-unknown

### Checklist

### DIFF
--- a/.changeset/stupid-papayas-jam.md
+++ b/.changeset/stupid-papayas-jam.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": minor
+---
+
+Addition of error handling on <DataTable .../> with unknown or missing column.id fixing #580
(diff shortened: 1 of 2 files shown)


hughess:
This is awesome!

It seems to work for me, except for this image column for some reason:
CleanShot 2023-01-27 at 15 21 38

Winterhart:
Nice catch! I will add two `undefined` checking. My hope is that we can remove most of the `undefined` checking by introducing `Typescript` + `Type`. For now, it should work.

Basically, every time we do:

```js

// can be undefined if the index does not exist
someVar[someIndex]


```

hughess (approved):

PR #598, opened 2023-01-27
minimal venn diagram implementation

mcrascal:
### Description
This implements a simple Venn Diagram per a customer request. 

This is sufficient for their use case, but it is not ready for prime-time. Propose we leave this undocumented, and open an issue for a broader diagrams suite in the future. 

* Does not offer robust data handling. Overlap percentages, labels, and amounts are supplied directly through props which accept arrays. Live data would be supplied using `query[row].column` syntax. 
* No automatic calculation of overlap percentages. There are a variety of approaches to calculating these, and we're not enforcing any one. For example, sets expressed as a percentage of a total records, sets expressed as a percentage of a single category of records etc.  
* No format handling -- default percentages for overlaps, en-gb for record counts. 

## Syntax 

```
<VennDiagram
    labels={["Title A", "Title B", "Title C"]}
    amounts={["AMOUNT A", "AMOUNT B", "AMOUNT C"]}
    overlaps={["A", "AB", "BC", "AC", "ABC"]}
/>
```

## Screenshot 

CleanShot 2023-01-26 at 21 14 57@2x

### DIFF
--- a/.changeset/smart-adults-think.md
+++ b/.changeset/smart-adults-think.md
@@ -1,0 +1,6 @@
+---
+"@evidence-dev/preprocess": patch
+"@evidence-dev/components": patch
+---
+
+Adds a minimal venn diagram to available components.
--- a/packages/preprocess/index.cjs
+++ b/packages/preprocess/index.cjs
@@ -58,6 +58,7 @@ const createDefaultProps = function(filename, componentDevelopmentMode, fileQuer
         import { pageHasQueries, routeHash } from '@evidence-dev/components/ui/stores';
         import { setContext, getContext, beforeUpdate } from 'svelte';
         import BigLink from '${componentSource}/ui/BigLink.svelte';
+        import VennDiagram from '${componentSource}/diagrams/VennDiagram.svelte';
         import Value from '${componentSource}/viz/Value.svelte';
         import BigValue from '${componentSource}/viz/BigValue.svelte';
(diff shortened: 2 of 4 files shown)


mcrascal:
CleanShot 2023-01-26 at 21 13 30@2x

archiewood:
Visually this looks really good.

I'm a little puzzled on the labelling. Two items:

1. I'm assuming the customer would want to label the absolute values for all of the sets, not just the A-- -B- and --C values?

2. Given the terminology, it's slightly unclear what the numbers next to the labels represent: is the number next to First Malaysian a count of all values associated with first Malaysian? (A--, AB-, A-C, ABC) Or it is the values that are _only_ first Malaysian (ie A--)

archiewood (commented):
Aside from comments re labels, everything looks excellent

mcrascal:
> I'm assuming the customer would want to label the absolute values for all of the sets, not just the A-- -B- and --C values?

They don't have a case for that -- their set overlaps are calculated as a percentage of A, but for completeness that would make sense.  

> Given the props terminology, it's slightly unclear what the numbers next to the labels represent 

These accept arbitrary values. Users could slot in anything, or nothing, in each of the labels. In the customer's case they represent total records for each category (A--, B--, C--). 

I'm going to ship this and open a second issue for documenting the feature, and providing a friendlier interface. We can prioritize that when appropriate.

PR #518, opened 2022-12-13
Add port and ssl options to mysql form

hughess:
### Description
This PR adds options for `port` and `ssl` to the MySQL database connector form on the Settings page. These options are a part of our MySQL connector, but I missed them when I added options to the settings form the first time. Thanks @hooopo for pointing this out.

### Checklist
- [x] For UI or styling changes, I have added a screenshot or gif showing before & after
- [x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)

### Before
CleanShot 2022-12-13 at 16 49 11@2x

### After
CleanShot 2022-12-13 at 16 48 37@2x

PR #599, opened 2023-01-27
Fix mysql ssl option

hooopo:
### Description
The current mysql connection for evidence is not working, even if the ssl option is not set. This is because 'true' is not a valid option for mysql2, as seen in this example: https://github.com/sidorares/node-mysql2/tree/07a429d9765dcbb24af4264654e973847236e0de/examples/ssl

This PR addresses the following issues:

* Users can set ssl to `true`
* Users can set ssl to `Amazon RDS`
* Users can set ssl to `JSON` string in order to handle more complex situations, such as https://mariadb.com/kb/en/nodejs-connection-options/#two-way-ssl-authentication
More information can be found here: https://github.com/evidence-dev/evidence/pull/518#issuecomment-1399571132

### Checklist

### DIFF
--- a/.changeset/wise-hounds-watch.md
+++ b/.changeset/wise-hounds-watch.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/mysql": patch
+---
+
+Fix mysql SSL option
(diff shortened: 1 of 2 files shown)


hughess:
@hooopo thanks for this PR! I'm not able to test this as I don't have a MySQL db set up at the moment. Were you able to test that the various ssl options work? If so, I'm good to merge this

hooopo:
@hughess Yes, I tested it in local.

## test with tidb.cloud ssl -> true


image

## test with tidb.cloud ssl -> {}
image

## test with tidb.cloud ssl -> false
image

## test mysql server without ssl
image

## test with plantscale
image

## test with singlestore including ca file

image

hughess:
Fantastic! The only thing left to do is add a changeset and then this is good to go - details for that are here: https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset

hooopo:
> Fantastic! The only thing left to do is add a changeset and then this is good to go - details for that are here: https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset

Added 😄

PR #601, opened 2023-01-27
Reduce imports from echarts library

hughess:
### Description
Previously, we were importing the full echarts library wherever it was used in the project, even though we only use 2 functions each time we import it. This PR reduces the imports to only the required functions. Thanks @Winterhart for the suggestion!

### Checklist


Winterhart (commented):
you are quick! cool changes

PR #623, opened 2023-02-01
Integrate tailwindcss (WIP)

nidhi-kala:
### Description

TODO
- [x] Install and initialize Tailwindcss according to [their docs](https://tailwindcss.com/docs/guides/sveltekit) for svelte kit. I was not able to use `'@sveltejs/kit/vite'` because we are on an older version of sveltejs #540 which doesn't contain that sub package. 
- [] Update 1 component to use tailwindcss classes instead of vanilla css. 
### Checklist


mcrascal:
Hi @nidhi-kala thanks for jumping into this! 

Is Svelte Kit 1.0 a blocker? We have a working version of Evidence on 1.0 in [this branch ](https://github.com/evidence-dev/evidence/tree/540-upgrade-to-sveltekit-10), and would be happy to take a PR to that branch, but this is looking quite promising!

nidhi-kala:
Hey @mcrascal !
No, it's not a blocker. Instead of using `sveltejs preprocessor` I'm using `vite` for postcss, which was already installed.
I have added our color theme to tailwind config and now working on converting the `ErrorChart.svelte` component to tailwind.

sites/example-project/src/components/viz/ErrorChart.svelte
```diff
@@ -3,62 +3,13 @@
     export let chartType;
 </script>
 
-<div width=100% class="error">
-    <div class="wrapper">
-        <div class="chart-type">{chartType}</div>
+<div class="grid [grid-template-rows:auto] grid-cols-[100%] justify-center bg-red-50 text-grey-700 font-ui font-normal rounded border border-red-100 min-h-[150px] box-content py-5 px-[30px] my-5 print:break-inside-avoid">
+    <div class="m-auto w-full">
+        <div class="font-bold text-center text-[1.2rem]">{chartType}</div>
```
mcrascal, line 8: @nidhi-kala I know you're doing a 1:1 mapping of what we already have here, which is great, but I'd be interested in your thoughts: 

Could we get _close_ to what we already have by using the existing tailwind font scales (e.g. `sm`, `md` etc.)? Or would would we need to add some additional scales to our tailwind config to get close to what we're already doing?

sites/example-project/src/components/viz/ErrorChart.svelte
```diff
@@ -3,62 +3,13 @@
     export let chartType;
 </script>
 
-<div width=100% class="error">
-    <div class="wrapper">
-        <div class="chart-type">{chartType}</div>
+<div class="grid [grid-template-rows:auto] grid-cols-[100%] justify-center bg-red-50 text-grey-700 font-ui font-normal rounded border border-red-100 min-h-[150px] box-content py-5 px-[30px] my-5 print:break-inside-avoid">
```
mcrascal, line 6: I think there's a slight regression here. The error chart should respect the width of it's container. So for charts themselves it will be the full width of the article, but in the case of BigValue (and potentially other components in the future), it should only take up the default width of that component. Compare [this branch](https://evidence-development-workspace-mtkzjo7f2-evidence-dev.vercel.app/text-and-metrics/big-value) with the [current state](https://evidence-development-workspace-git-table-download-evidence-dev.vercel.app/text-and-metrics/big-value)

sites/example-project/src/components/viz/ErrorChart.svelte
nidhi-kala: Closing because this is resolved/updated.

sites/example-project/src/components/viz/ErrorChart.svelte
nidhi-kala: This has been updated in new commits. Marking as resolved.

mcrascal:
Hi @nidhi-kala apologies for the delay. #540 has been taking all of my attention! 

In my manual review, I noticed two small regressions vs. main. 

1. When opening and closing the menu items on the left hand sidebar (e.g. `Text and Metrics`), the svelte slide animation doesn't run. 
2. In development mode, the settings icon is quite small, and shifted up and right vs. main. 

Any thoughts on these?

mcrascal:
Moved to #702

PR #629, opened 2023-02-02
CSV Support

hughess:
### Description
- We already support CSV file querying through our DuckDB connector, but there are a few drawbacks for someone who wants to use only CSVs:
   - They need to include an empty DuckDB file along with their CSVs
   - The path to the file is not intuitive + it's not clear where to store the CSV files 
- This PR adds a new connector for CSVs, which uses DuckDB under the hood
- When the CSV connector is selected, it spins up an in-memory DuckDB instance and uses that to query the CSV files in the project
- This PR also includes a new `sources` directory, which will be the recommended place to store CSV files. E.g., 
```sql
select * from 'sources/myfile.csv'
```

### Checklist

### DIFF
--- a/.changeset/hip-deers-look.md
+++ b/.changeset/hip-deers-look.md
@@ -1,0 +1,8 @@
+---
+"@evidence-dev/csv": minor
+"@evidence-dev/db-orchestrator": minor
+"@evidence-dev/evidence": minor
+"@evidence-dev/components": minor
+---
+
+Add CSV connector
--- a/packages/csv/.gitignore
+++ b/packages/csv/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/packages/csv/index.cjs
+++ b/packages/csv/index.cjs
@@ -1,0 +1,4 @@
+const runQuery = require('@evidence-dev/duckdb')
+
+module.exports = runQuery
+
--- a/packages/csv/package.json
+++ b/packages/csv/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "@evidence-dev/csv",
+  "version": "0.0.1",
+  "description": "Redshift driver for Evidence projects",
+  "main": "index.cjs",
+  "author": "Sean Hughes",
+  "license": "MIT",
+  "scripts": {
+    "test": "uvu test test.js"
+  },
+  "dependencies": {
+    "@evidence-dev/duckdb": "workspace:^",
+    "@evidence-dev/db-commons": "workspace:^"
+  },
+  "type": "module"
+}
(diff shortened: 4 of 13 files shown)


archiewood:
Nice! This seems like a pretty sensible implementation.

## Documenting usage
I think we could do three things to help users understand how to use the csv connector:
- Explain where evidence looks for csv files, and include and example code block in the settings panel:
> Evidence can query CSV files stored in the `sources/` directory in the root of your project:   `select * from sources/example_file.csv` CSV querying is powered by DuckDB - see their docs for reference on [query syntax](https://duckdb.org/docs/sql/query_syntax/select)
- Add a usage of the csv connector in the template project, 
- I might also add an explicit note about how to use CSVs in the [supported databases page of the docs](https://docs.evidence.dev/getting-started/connect-database), as it is slightly different to the others.

## Other thoughts
It's a bit weird that the sources directory ends up in the root of the monorepo (though I understand why it does, given the relative path of example project
  
## Future steps:

I think a more flexible implementation would allow an arbitrary source file directory, eg to the seeds folder of a dbt repo that Evidence was installed within. Perhaps it does already if someone were to do `select * from ../../../seeds/my_data.csv`

hughess:
@archiewood great thoughts.

### File Directory

This implementation can accept a full file path, so you can throw anything in there. There are three places the CSV can be stored/accessed from:
1. In the `sources` directory: `select * from 'sources/myfile.csv'`
2. Anywhere on your computer, referenced with an absolute path: `select * from '/Users/myname/Downloads/myfile.csv'`
3. Anywhere on your computer/in your Evidence project, referenced with a relative path: `select * from '../../../folder/myfile.csv'`

Number 3 is not ideal because it requires a 2-folder hop to get into the root of the Evidence project, but numbers 1 and 2 work as you'd expect.

### Documentation
100% agree with these

I also think there is an argument to be made for making the template project use the CSV connector

### Sources Folder in the Monorepo
This should probably be removed (along with some other test pages I had set up in the monorepo)

hughess:
Here is the new settings menu option for CSV:
CleanShot 2023-02-08 at 13 13 22@2x

archiewood:
I think this looks great.

Out of interest, what does the **Test** button do here? Persumeably there is nothing to test?


## Aside
The only thing that isn't clear here is the relative path magic (ie we only copy csv files from that file) so CSVs in other places need the ../../ hop.

But I think we can perhaps leave that - it's not our "recommended" way of doing things

Winterhart (commented):
I have been able to connect a `csv` from the settings and append the `.gitignore` file...

I only have few recommandations for the `post` function.

hughess:
@archiewood the test button runs a query against the in-memory duckdb.

We might need to add a note about the relative path issue, but I think for now we can see how we do with the two recommended options

PR #635, opened 2023-02-09
Code Blocks (non-SQL)

archiewood:
### Description
Addresses #408 .
This implementation does not include code highlighting, as ~it conflicted with Query Viewer highlighting~ Vite was choking on the imports in prod

CleanShot 2023-02-09 at 22 25 43

### Checklist


sites/example-project/src/components/ui/CodeBlock.svelte
```diff
@@ -92,7 +86,7 @@
           // Dynamically load all required languages from the cdn
           // This cannot be loaded from node_modules because it is interpolated.
           import(
-            `https://cdnjs.cloudflare.com/ajax/libs/prism/1.29.0/components/prism-${rl}.min.js`
```
archiewood: This seems to not work during build

```
The above dynamic import cannot be analyzed by vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats.
```

sites/example-project/src/components/ui/CodeBlock.svelte
```diff
@@ -92,7 +86,7 @@
...
+            `../../../../node_modules/prismjs/components/prism-${rl}.min.js`
```
archiewood, line 89: This also doesn't work, as suggested by the comment...

archiewood: removed this functionality

sites/example-project/src/components/ui/CodeBlock.svelte
archiewood: functionality removed

sites/example-project/src/components/ui/CodeBlock.svelte
```diff
@@ -0,0 +1,121 @@
...
+      if (!copied) {
+        await navigator.clipboard.writeText(text);
+        copied = true;
+        setTimeout(toggleCopied, 1500);
+      }
+    } catch (e) {}
+  };
+</script>
+
+<pre><code>{#if source}{source}{:else}<slot />{/if}</code>
```
Winterhart, line 23: Why do we want to put a `<slot/>`  if `source` is `undefined` ? 

if you don't need the `slot` you can start the component with

```svelte

{#if source !== undefined} 

```

or 

```svelte

{#if source }

```

and you will not have the need to validate `source` is not `undefined` in the inner-components

sites/example-project/src/components/ui/CodeBlock.svelte
```diff
@@ -0,0 +1,121 @@
...
+{#if copyToClipboard}
+<button
+  type="button"
+  class="container"
+  class:copied
+  on:click={() => {
+    if (source !== undefined) {
+      copy(source);
+    }
+  }}>{#if copied}<Success color=var(--green-500)/>{:else}<Copy/>{/if}</button>
```
Winterhart, line 33: It's better to give  `if/else` indentations and spaces

```svelte
    {#if copied}
      <Success color=var(--green-500)/>
    {:else}
      <Copy/>
    {/if}

```

packages/preprocess/index.cjs
```diff
@@ -159,6 +178,8 @@ const updateExtractedQueriesDir = function(content, filename){
 
     visit(tree, 'code', function(node) {
         let id = node.lang ?? 'untitled'
+         // Prevent "real" code blocks from being interpreted as queries
+         if (prismLangs.has(id.toLowerCase())) return
```
Winterhart, line 182: The only trade-off I see here is if someone wants to name their query `Shell` it will not run it.

packages/preprocess/index.cjs
```diff
@@ -5,6 +5,24 @@ const visit = require('unist-util-visit')
 const md5 = require("blueimp-md5");
 const fs = require('fs')
 const fsExtra = require('fs-extra')
+const { supportedLangs } = require("./supportedLanguages.cjs");
+// This is future proofing for when we add support for Prism highlighting
+const prismLangs = new Set()
```
Winterhart, line 10: I would extract line 10 to 25 to a `function` returning a `Set`

Winterhart (commented):
Few minor comments/ideas and questions 🤔  Great job it's a nice feature! 👍🏼

sites/example-project/src/components/ui/CodeBlock.svelte
archiewood: The idea here would be to allow an alternative use of the codeblock like this:
```svelte
<CodeBlock language=html>
   <div>Hello world</div>
</CodeBlock>
```

as an alternative to the markdown syntax

sites/example-project/src/components/ui/CodeBlock.svelte
archiewood: I am not sure why, but when I included the indentations, it introduced whitespace into the component near the Copy Icon. This was the workaround i found to remove it.

packages/preprocess/index.cjs
archiewood: Yes this is true. These names are now "illegal"

sites/example-project/src/components/ui/CodeBlock.svelte
archiewood: This is some kind of svelte bug / weird behavior: https://github.com/sveltejs/svelte/issues/3204

I've edited to format it as readably as possible, but adding indents causes them to show up as whitespace in the codeblock.

PR #642, opened 2023-02-16
Fix for sources directory in template

hughess:
### Description
In the CSV PR, I included support for a `sources` directory. Unfortunately, I missed one place to reference the sources folder so that it would be populated in a project’s `.evidence` template. I added it to the file watcher process, but not the script that builds the template.

In most cases, this doesn’t matter and it has no effect. But if you add a sources directory to a fresh install, it causes a problem:
- The file watcher picks up that the folder is there, but is not able to find the corresponding folder in the `.evidence` template
- The error shuts off the dev server in the terminal
- Somehow, the dev server continues to run on port 3000 in the background - trying to restart the dev server results in “port 3000 is already in use”
- Going to localhost:3000 shows the header and sidebar of the site, but it isn’t able to load any of the content. In the console, it shows that the query status checker is still running constantly

### Solution
Adding `sources` to the template creation script solves this:
- If there's no `sources` directory in the project, nothing happens
- If you add a `sources` directory to a fresh install, it's able to match the two and copy over the files

I've tested this in a fresh install by editing the template inside the evidence package (from within `node_modules`)

### Checklist


hughess:
@mcrascal I missed adding `sources` in one place in my CSV PR. After this change, the template/file watcher appears to be working identically for both `static` and `sources`, but hoping to get a quick check that I haven't missed anything this time.

PR #644, opened 2023-02-17
support multiple statements (duckdb)

watsy0007:
### Description
support multiple statements
```sql
install httpfs;
load httpfs;

set s3_ak='ak';
set s3_sk='sk';

select count(1) as count from read_parquet("s3://bucket/example.parquet");
```

### Checklist


archiewood:
Hey @watsy0007, Thanks for the PR!

Would be great if you could give a bit more context about what you're trying to do here, and how this would help you achieve it.

A bit nervous about adding this as is, as we need to think about how we'd return the results of each statement to the UI.

watsy0007:
thanks for your relay @archiewood 

our data was saved on S3 with CSV and Parquet formats. 
To use DuckDB for analysis, we need to load `httpfs` extension and set S3 credentials.

i have 2 options to solve this:

1. Execute those codes in code block in markdown files
2. Add pre-execute codes when setting db config and execute them when connect button is pressed

In my opinoin, option 1 is simple and easy. but i think option 2 is more appropriate

archiewood:
I can see how httpfs is useful for duckDB, if you need to query CSV / parquet files that are stored in S3

I _think_ the best way to achieve it, rather than allowing multiple statements (which could be a trap in other cases, for example if someone tries to run multiple `SELECT` statements, is to allow HTTPFS to be configured as a setting.

Perhaps a checkbox in the settings UI where you connect to DuckDB ("Use HTTPFS" or similar), which then shows fields to enter of `s3_region` and `s3_access_key_id`, and `s3_secret_key_id`.

Then use the duckdb connector to conditionally run the required SQL statements, if HTTPFS is selected?

Not totally sure of the details - but what do you think @watsy0007 ? Would this serve your use case?

watsy0007:
> Perhaps a checkbox in the settings UI where you connect to  DuckDB ("Use HTTPFS" or similar), which then shows fields to enter of s3_region and s3_access_key_id, and s3_secret_key_id.
>
> Then use the duckdb connector to conditionally run the required SQL statements, if HTTPFS is selected?

I thinks this is a great solution. I very glad to see this feature in the future. And sorry for my lack of svelte experience. 

Should I close this issue?

archiewood:
I suggest I open it as a separate issue and close this PR

Issue #651, opened 2023-02-23
Upgrade DuckDB connector to 0.7.0


archiewood:
See: https://github.com/motherduckdb/duckdb-async/issues/2

archiewood:
Closed by #654 - need to bump the template project when we release this

PR #654, opened 2023-02-23
Bump duckdb to 0.7.1

gregwdata:
### Description
Update duckDB to 0.7.0, addressing https://github.com/evidence-dev/evidence/issues/651

Incremented the `duckdb-async` dependency version, and manually migrated the `needful_things.duckdb` file from `0.6.1` to `0.7.0`. That manual migration including updating the schema of each table to ensure timestamp and date columns are stored as the correct type.

This updated `needful_things.duckdb` file will need to be copied to the template repository to make the tutorial work correctly once this version is released. I have not done that, but would be happy to submit a PR there if it helps (will need guidance on how to coordinate version numbers.

Tested by creating the template repo and installing this version of Evidence to it. That highlighted the need to change db schema, otherwise the queries on the index threw errors, since it was running a `date_trunc` function on a column that was stored as a VARCHAR. Verified that query results rendered correctly on the template project once the schema was modified while migrating to `0.7.0` version of the file.

### Checklist
- [x] For UI or styling changes, I have added a screenshot or gif showing before & after (**NA**)


archiewood:
@gregwdata 

Epic PR, thank you 👌🏼

### Regarding the types
- I _think_ I upgraded the `template` project's duckdb file to use native datetime types [here](https://github.com/evidence-dev/template/commit/b95432dee0a3b025c3b883f68a6d68ed20b29512), which is why the template project queries currently work.
- However, I didn't upgrade the one in the `evidence` repo (I actually wasn't aware it was committed!)
- But I think we will need to upgrade the duckdb file in the template _anyway_ for 0.7.0 as the duckDB API is not backwards compatible with old

### Regarding version number co-ordination
We don't have an automatic solution yet, but once we release this change, then we will have a version number we can point the template to.

Side point: We've been thinking about locating the template inside the `evidence` monorepo longer term, which would solve this versioning issue!

packages/duckdb/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "uvu test test.js"
   },
   "dependencies": {
-    "duckdb-async": "0.6.1",
```
archiewood, line 12: Was your package-lock.json file also updated?
If so can you commit that too?

gregwdata: Can you point me to the package-lock.json file you're referring to? I don't see one in an obvious place in the main branch of evidence that I forked. (I'm an NPM noob, so may be missing something obvious here 😁 )

archiewood: (I suspect this is what is causing the tests to fail)
` ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/duckdb/package.json`

gregwdata: gotcha. Just committed that.

archiewood:
Hmm the macOS tests seem to be timing out during the install process.

We’ve had issues with the duckDB binaries before on macOS, could be a similar issue.

@antonycourtney have you seen any issues with MacOS?

archiewood:
It appears this is in fact the issue. The pre-built binaries for DuckDBs npm package failed to upload for 0.7.0.

As a result widows and Mac are attempting to build from source which is so slow that the tests are timing out and failing.

There's some discussion of this on their Discord Nodejs channel

DuckDB are looking to release 0.7.1 soon to address this.

archiewood:
Update: 0.7.1 released. Now just for a duckdb-async release and we can try again!

antonycourtney:
> Update: 0.7.1 released yesterday. Now just for a duckdb-async release and we can try again!

On it...

antonycourtney:
duckdb-async 0.7.1. now published on npm. Good luck with upgrade, please let me know if you encounter any issues.

PR #666, opened 2023-03-01
Add Evidence deployment option

archiewood:
### Checklist

### DIFF
--- a/.changeset/gentle-gifts-pull.md
+++ b/.changeset/gentle-gifts-pull.md
@@ -1,0 +1,5 @@
+---
+"@evidence-dev/components": patch
+---
+
+Add Evidence deploy option
--- a/sites/example-project/src/app.css
+++ b/sites/example-project/src/app.css
@@ -153,7 +153,7 @@ hr {
 
 code {
 	font-family: var(--monospace-font-family);
-	font-size: 0.7em;
+	font-size: 0.8em;
 	background-color: var(--grey-100);
 	padding: 0.2em 0.2em;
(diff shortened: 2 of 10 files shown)


hughess:
Nice, the deployment form is great!

Found some styling issues with the query viewer:
- Looks like the font-size in the data table is quite a bit larger than before
- Font size in the SQL pane is also a bit larger
- The scrubber looks a bit off (I assume due to the change in font-size/line height)
- The space below the query viewer when the table is open looks larger than before
CleanShot 2023-03-02 at 08 18 06@2x

hughess (approved):
Looks good to me! I think worth a quick check on Windows as well to make sure the styling lines up

hughess:
I took a look at the production preview on Windows and there don't seem to be any major differences between that and Mac. 

Last things:
- The font size in the data table is still a bit larger than it was before
- The spacing around the query viewer when closed is larger than before - you can see this on the Query Chaining page, where there are query viewers stacked up at the bottom of the page
- On mobile, the font size in the SQL section of the query viewer is very large

archiewood:
@hughess good spot on the mobile SQL font size and query viewer margin.

Re the data table font size -> its now corresponds to an exact point size rather than a fractional pt size: e.g for 11px vs 10.4px @ wide viewports, 9px vs 8.4px @ narrow viewports

I'm tempted to leave this unless you think it is materially worse?

Before -> After

image

hughess:
@archiewood looks good to me!

PR #674, opened 2023-03-09
Misc. Fix for SvelteKit Upgrade

Winterhart:
- Move to relative import instead of lib import for some components
- Add variable `fs` settings on `vite` config

### Checklist

### DIFF
--- a/packages/preprocess/index.cjs
+++ b/packages/preprocess/index.cjs
@@ -30,13 +30,12 @@ const getPrismLangs = function(){
 
 const getRouteHash = function(filename){
     let route = filename.split("/src/pages")[1] === "/+page.md" ? "/" : filename.split("/src/pages")[1].replace(".md","").replace(/\/\+page/g,"")
-    let routeHash = md5(route)
-    return routeHash
+    return md5(route)
 }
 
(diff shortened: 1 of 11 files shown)


packages/preprocess/index.cjs
```diff
@@ -30,13 +30,12 @@ const getPrismLangs = function(){
...
 const getRouteHash = function(filename){
     let route = filename.split("/src/pages")[1] === "/+page.md" ? "/" : filename.split("/src/pages")[1].replace(".md","").replace(/\/\+page/g,"")
-    let routeHash = md5(route)
-    return routeHash
+    return md5(route)
 }
 
 const createDefaultProps = function(filename, componentDevelopmentMode, fileQueryIds){
     let componentSource = componentDevelopmentMode ? '$lib' : '@evidence-dev/components';
-    let routeHash = getRouteHash(filename)
```
ud3sh, line 39: routeHash -> routeH?

ud3sh (approved):

packages/preprocess/index.cjs
Winterhart: Yes, the same name was used twice as we have an import on `routeHash`

PR #697, opened 2023-03-16
Fix table search bug

### LINKED ISSUE #586: Bugs in table search
**Bug Description**
While on a call with a user, I noticed that a number of search terms did not return correctly.
Some examples were:
Alphanumeric id strings: eg `kugOug42354hOU234Ug`
Dates (perhaps formatted as strings?): `2020-03-12` (it seemed to get them until the user typed the specific day. So `2020-03` would work but `2020-03-12` would then not return any rows)

**Severity**
- [x] Medium - I'm able to use Evidence, but a feature I need to use is not working

hughess:
### Description
Fixes bug where DataTable was returning no results. The table was not correctly handling null values across all data types.

Fixes #586 

### Example to Reproduce Error
Prior to these changes,  you could get the error to appear with the query below (in BigQuery) plus the DataTable component:
```sql
select 
  distinct country_code, 
  two_alpha_code,
  short_name,
  regexp_contains(two_alpha_code, '[0-9]') or regexp_contains(two_alpha_code, 'X') or
  (regexp_contains(two_alpha_code, 'Z') and not regexp_contains(two_alpha_code, 'NZ')) or
  two_alpha_code is null as category,
  special_notes
from `bigquery-public-data.world_bank_wdi.country_summary` 
```

```html
<DataTable data={countries} search=true />
```

Then, search for "NZ" and the table will return 0 results and will show an error in the browser console.

### Checklist


Winterhart (approved):
Looks good I have been able to test it with the Table on preview

CleanShot 2023-03-16 at 10 53 19

PR #705, opened 2023-03-20
Add Alert Component

ItsMeBrianD:
### Description

This adds a very simple Alert component; it is available in 5 styles (default, info, success, warning, danger). There are many ways we could expand this, but it would need a little more definition.

This is based on [Flowbyte's Alert Component](https://flowbite.com/docs/components/alerts/), which provides a good starting point.

### Checklist

image


sites/example-project/src/components/ui/Alert.svelte
```diff
@@ -0,0 +1,30 @@
...
+    &.danger {
+      @apply text-red-800 bg-red-50;
+    }
+    &.success {
+      @apply text-green-800 bg-green-50;
+    }
+    &.warning {
+      @apply text-yellow-800 bg-yellow-50;
+    }
+  }
```
mcrascal, line 29: I _really_ like this approach to applying the TW styles. Much more svelte-y

ItsMeBrianD: There are pro's and cons to this; typically I use this whenever there are dynamic styles (e.g. here we have the `status`), because it works well against tailwind's auto-pruning of styles.

One con is that this can really slow down the build time if you use it everywhere, as (if I recall correctly), it will process each of style tag separately. [There are some docs on the topic](https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction)

ItsMeBrianD:
https://user-images.githubusercontent.com/10779616/226670550-88341a4e-319b-4754-a39f-c40430c85ddf.mp4


Here is a "sticky" behavior; this prevents the alert from potentially blocking content at the top of the page.

Alternatively, we can add a [portal](https://github.com/romkor/svelte-portal) and place the sticky _inside_ the header, which would keep it fixed at the top of the page wherever the user is without blocking content.

mcrascal:
I like it -- I think when it's in that sticky mode, it would look better if it was full width. 

I think either that would mean: 

1. We transition sticky alerts from "in page" style to full width sticky when you scroll past them, or; 
2. Sticky alerts start that way stacked under the header in the order they appear on the page (which I think would necessitate breaking out of the layout/ shifting down all the other content.) 

Thoughts?

ItsMeBrianD:
image
By full width do we mean escaping the margins and taking the entire screen?

ItsMeBrianD:
Also; it seems that a deeper dive into why nested styles were breaking is going to be needed; the tabs component has a `hover:` and `active:` directive used in the styles which seems to be breaking the build as well.

https://github.com/evidence-dev/evidence/actions/runs/4481516803/jobs/7878329845?pr=707#step:7:268

mcrascal:
Yeah red arrows is what I'm thinking width wise.

ItsMeBrianD:
https://user-images.githubusercontent.com/10779616/226717500-ce033fdb-da76-4f7b-8a11-e788e5cbb061.mp4

This is how the sticky behavior now functions; I modified [svelte-portal](https://github.com/romkor/svelte-portal/tree/master) to accept a `null` target, which will move the portal'd element back to it's original place in the DOM. This means that the sticky state can be toggled on/off seamlessly.

Currently, the behavior previously shown no longer exists, as when the element is made sticky, it is moved outside the content and into the main container, but I believe this is functioning approximately as expected.

One thing that may need to be addressed in the future - this is somewhat brittle, relying on the `#evidence-container` element to exist in order to function properly; but going further than an id adds a lot of complexity for a relatively simple component.

ItsMeBrianD:
Got the build fixed up 🎉 

If the design looks good; this should be good to ship. Once this is merged into the tailwindcss branch I can send it off to the tab branch and fix the build there as well

Issue #294, opened 2022-06-02
Support for Markdown Frontmatter


archiewood:
##  Feature Description
The ability to add frontmatter to the top of a markdown file that is rendered by Evidence, and have it passed through to the <head> of the evidence pages.

Eg in YAML format
```markdown
---
title: A page title
description: Here is a detailed description of the page that is typically up to 160 characters
og:
    title: Title for social
    description: Description for social
    image: img_for_social.png
---

# Page h1
Here is the start of the page content
```
Would render in the final page HTML as
```
<head>
    <title>A page title</title>
    <meta name="description" content="Here is a detailed description of the page that is typically up to 160 characters"/>
    <meta property="og:title" content="Title for social">
    <meta property="og:description" content="Description for social">
    <meta property="og:image" content="https://www.website.com/path/to/img_for_social.png">
</head>
```

## Goal of Feature
- Better indexing of pages by search engines
- Ability to change how the page appears when shared on Slack, Twitter etc.

## Current Solution
Can directly put HTML into a .md page to be processed by svelte
```
<svelte:head>
    <title>A page title</title>
    <meta name="description" content="Here is a detailed description of the page that is typically up to 160 characters"/>
</svelte:head>
```

But the [__layout.svelte file](https://github.com/evidence-dev/evidence/blob/main/sites/example-project/src/pages/__layout.svelte) will override the title tag to be "Evidence" irrespective. This can (and should) be removed to allow greater control.

## Related Issues
- MDsveX [supports frontmatter](https://mdsvex.pngwn.io/docs#frontmatter-1), but requires using MDsveX layouts (rather than svelte layouts)
    - https://github.com/pngwn/MDsveX/issues/313
    - https://github.com/pngwn/MDsveX/issues/242
- An interesting [Svelte layouts + MDsveX implementation](https://github.com/mvasigh/sveltekit-mdsvex-blog)

archiewood:
We could also use this to support defining parameters to be used in the page. 

Eg filters that are applied to every SQL query.
```
---
filters:
  - year  # where year = {$page.params.year}
---
```

archiewood:
Also:
- Ability to name parameterized pages (ie what shows up in title and breadcrumbs)

mcrascal:
To start here, we're going to add support for a front-matter page title, and a frontmatter object. 

This page title should appear on the page, in the nav sidebar, and in the title in the browser tab. It should be optional, not including it on a page shouldn't change the existing behaviour. 

[...]

Issue #85, opened 2021-08-26
Support standalone .sql files 


mcrascal:
Would be great to be able to use a .sql file that is sitting outside of your markdown document.

mcrascal:
I can place a .SQL file in a `sources` directory, and include it in a page using front matter, from there, it is treated as if I had written it in code fences.

mcrascal:
Considerations:  
- What's MVP for presenting the query results on the page (e.g. stack them at the top/bottom, provide a standalone route) 
- Treatment on not-resolvable paths 
- Treatment of extremely large queries (e.g. warning -- might pull people down a path of having extremely large result sets) 
- Naming of the sources directory (e.g. data sources, queries etc.) 
- Path escaping the sources directory

ItsMeBrianD:
There have been some performance issues when loading large datasets (>50k rows), I think it's related to the movement of some items from preprocess to load functions. Previous behavior was slow initial load, but cache worked well after that; now it is just slow. 

While refactoring some preprocess steps it might be worthwhile to see if this is low hanging fruit.

PR #710, opened 2023-03-22
Improved support for frontmatter

ItsMeBrianD:
### Description

This PR addresses #294; and improves the support for Markdown Frontmatter.

MDSvex already provides any frontmatter to the component (in this case, `+page.md`), so this PR primarily addresses adding things like a page title, description, and opengraph information.

Because this touches preprocess, I took the opportunity to split it up into several files, functionality should not be impacted, but it makes things a little easier to reason with when looking at a specific part of the pipeline.

### Checklist
- [ ] ~For UI or styling changes, I have added a screenshot or gif showing before & after~


ItsMeBrianD:
Fixed the 2 listed issues;

With regard to #85, I think there will need to be a little more clarity on the goal and use-cases. In theory it could be handled either at build-time or at run-time, as the logic to extract the frontmatter will be basically the same (e.g. regex match on `---` and then parse the yaml content).

packages/preprocess/src/process-queries.cjs
```diff
@@ -0,0 +1,164 @@
...
+         }
+    });
+    return queryIds;
+}
+
+/**
+ * @type {(componentDevelopmentMode: boolean) => import("svelte-preprocess/dist/types").PreprocessorGroup}
+ */
+const processQueries = (componentDevelopmentMode) => {
+    let queryIdsByFile = {}
```
ud3sh, line 144: This `queryIdsByFile` was bit of a strange "workaround" as we needed remember state between different pre-process phases (without having to load things twice from what I recall).   I think this works neatly with way you refactored this, under the assumption that the order of the array returned by `evidencePreprocess` doesn't matter.   I think it doesn't but I am not sure.

sites/example-project/src/pages/frontmatter/+page.md
```diff
@@ -0,0 +1,37 @@
+---
```
ud3sh, line 1: +1 for example!

ud3sh (approved):

sites/example-project/src/pages/frontmatter/+page.md
```diff
@@ -0,0 +1,37 @@
...
+| `title`          | changes the name of the tab, and also adds a header to your page                                                             |
+| `description`    | is used for search engines                                                                                                   |
+| `og`             | changes how your link shows up when shared on things like Slack, Facebook, Twitter, Discord, etc                             |
+| `og.title`       | changes the title that appears in the embed; if this is not specified, but `title` is, then `title` is used (and vise versa) |
+| `og.description` | changes the body of the embed                                                                                                |
+| `og.image`       | will appear in the embed if specified, but it is not required.                                                               |
+
+
+You can access any of your frontmatter values by surrounding them with `{}`.
+For example, this page has `some_arbitrary_frontmatter` included in it's own frontmatter, which has a value of `{some_arbitrary_frontmatter}`
```
mcrascal, line 37: Can we update this example to show this value being printed into the page?

ItsMeBrianD: This should render as 
> For example, this page has `some_arbitrary_frontmatter` included in it's own frontmatter, which has a value of `This is arbitrary, and can be anything`

ItsMeBrianD: This is fixed on the latest deployment

mcrascal:
This looks really good. 

1. We'll need docs: I'd add this to the markdown reference page, and I'd include a sentence or two on the `core-concepts/pages` with a link over to the reference. 
2. I really like the open graph support. I think in a future iteration of that it'd be great to default back to an image of the first chart on the page. This is beyond the scope of this frontmatter issue though, so I'll open something separate for that.

mcrascal (approved):

archiewood:
Hey @ItsMeBrianD, while I've been testing the standalone SQL files PR, I've run into an issue (or at least a thing that worked unexpectedly)

A pretty common pattern in templated pages is to add a title named after the parameter.

ie if you have 
```
|-- index.md
`-- [categories] /
       `-- index.md
```
Then you title it like this

```markdown
# {$page.params.category}

Some other content
```

I assumed this would work in the frontmatter like this
```
---
title: {$page.params.category}
---

Some other content
```

which actually gives: 
image

I assume the js object is being converted into a string before it reaches the page somehow?

archiewood:
This also brings up a more general question about js support in frontmatter. (We do not have to sovle this now, just thoughts)

It would be great if we can eliminate as much of the need for people to add `<script>` tags as possible.

**Should we support arbitrary js in frontmatter?**

I quite often end up getting an associated field from a table like this.
```
<script>
  let company_id = companies.filter(d=>d.company===$page.params.company)[0].company_id
<script>
```
```
---
company_id: companies
- filter: company
- equals: {$page.params.company}
- column: company_id
- row: 0
---
```

or just

```
---
company_id:  {companies.filter(d=>d.company===$page.params.company)[0].company_id}
---
```

This could be a bad idea for reasons I have not thought of!

ItsMeBrianD:
Regarding the title; that might work with this syntax:
```
---
title: $page.params.category
---
```
If it doesn't then this probably isn't supported right now (but could be), worth exploring for sure.

Adding variables / filters in the frontmatter could be nice; but might get a little wonky - we'll have to explore. This probably deserves it's own issue to make sure that we can have the discussion and nail it down before writing the code.

PR #734, opened 2023-04-04
External SQL File Support

ItsMeBrianD:
### Description

This pull request adds support for `.sql` files in the `sources/` directory. These files will behave as if they were inlined, meaning that `file -> file`, `inline -> file` and `file -> inline` query dependencies function as intended.

Query views for external files are placed at the bottom of the page.

### Checklist


sites/example-project/src/pages/queries/external-queries/+page.md
```diff
@@ -0,0 +1,108 @@
...
+<CodeBlock source={`my_project/
+  pages/
+    my_page.md
+  sources/
+    my_query.sql`}/>
+
+<br/>
+You can now access `my_query.sql` with frontmatter:
+
+<CodeBlock source={`---
```
archiewood, line 40: As a bit of an aside, I think you should be able to do:

````
```shell
---    
sources:
  - my_query.sql
---
```
````

and it should render as a codeblock

File: packages/preprocess/README.md [hunk 2]
```diff
@@ -0,0 +1,40 @@
+# Evidence Preprocess
+
+This package is responsible for transforming Evidence Markdown files into Svelte Components / Pages.
+
+## Overview
+
+The basic pipeline looks something like this:
+
+```mermaid
```
Winterhart, line 9: 👍🏼 cool tool

packages/preprocess/src/extract-queries/extract-queries.cjs
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * @param {string} content File content
+ * @param {string} filename File name
+ * @returns {Query[]}
+ */
+const extractExternalQueries = (content, filename) => {
+  const frontmatter = parseFrontmatter(content);
+  if (!frontmatter) return [];
+  if (!frontmatter.sources) return [];
+  if (!Array.isArray(frontmatter.sources)) {
```
Winterhart, line 31: You don't want to return? here

packages/preprocess/src/extract-queries/extract-queries.cjs
```diff
@@ -0,0 +1,107 @@
...
+    );
+  }
+
+  /**
+   * @type Query[]
+   */
+  return frontmatter.sources.map((source) => {
+    if (typeof source === "string") {
+      const id = source.split(".sql")[0].replace("/", "_").replace("\\", "_");
+      const content = fs.readFileSync(`./sources/${source}`).toString().trim();
```
Winterhart, line 43: do you have a `catch` if file doesn't exist?

packages/preprocess/src/process-frontmatter/handle-og.svelte
```diff
@@ -1,33 +1,33 @@
 <!-- 
-        MDSvex comes in handy here because it takes frontmatter and shoves it into the metadata object.
-        This means that all we need to do is build out the expected page metadata
-    -->
+    MDSvex comes in handy here because it takes frontmatter and shoves it into the metadata object.
+    This means that all we need to do is build out the expected page metadata
+-->
 {#if typeof metadata !== "undefined" && (metadata.title || metadata.og?.title)}
```
Winterhart, line 5: I am curious why not if `metadata` !== `undefined`

packages/preprocess/src/process-frontmatter/parse-frontmatter.cjs
```diff
@@ -0,0 +1,30 @@
...
+    // if (!stat.isFile()) throw new Error(`${filepath} is not a file`)
+
+    // // Read File
+    // const fileContent = fs.readFileSync(filepath).toString()
+
+    // Run against regex
+    const frontmatterMatches = frontmatterRegex.exec(content)
+
+    // We have a match
+    if (frontmatterMatches?.[1]) {
```
Winterhart, line 22: Should we have a check on `array` `length`

packages/preprocess/src/utils/get-route-hash.cjs
```diff
@@ -0,0 +1,19 @@
+const md5 = require("blueimp-md5");
```
Winterhart, line 1: I was wondering if the default `crypto` module had a `md5`? 🤔

packages/preprocess/src/utils/get-route-hash.spec.cjs
```diff
@@ -0,0 +1,20 @@
...
+    })
+
+    it("should hash /src/pages/some-route/+page.md to equal md5('/some-route')", () => {
+        const result = getRouteHash("/src/pages/some-route/+page.md")
+        expect(result).toEqual(md5("/some-route"))
+    })
+
+    it("should hash /src/pages/random-string/+page.md the same way twice", () => {
+        const route = `/src/pages/${faker.company.bsAdjective()}-${faker.company.bsBuzz()}-${faker.company.bsNoun()}/+page.md`
+        expect(getRouteHash(route)).toEqual(getRouteHash(route))
```
Winterhart, line 18: As `md5` should always be the same, it would be okay to have getRouteHash(route) and a hard-coded string
e.g. `5d41402abc4b2a76b9719d911017c592`

sites/docs/docs/core-concepts/queries/index.md
```diff
@@ -81,4 +81,68 @@ Some SQL dialects require sub-queries to be aliased, including Postgres and MySQ
 
 The SQL compiler detects circular and missing references. If a query includes either a circular reference or a missing reference, Evidence will display an error that looks like a syntax error in a normal SQL query. Queries with compiler errors are not sent to your database.
 
-![circular-error-single](/img/circular-error-single.png)
\ No newline at end of file
+![circular-error-single](/img/circular-error-single.png)
```
Winterhart, line 84: 👍🏼

Winterhart (commented):
Few questions...

mcrascal:
This is fantastic. 

1. For the frontmatter syntax. Can we set this up so that you set a name for the query when you reference it? I want to decouple the folder structure in sources from the names people have to use inside their pages. See probably incorrect yaml below. 
2. Can we not allow standalone files to reference page queries. I think ideally they should be ignored by the query compiler. This is really cool that it works, but I think it'll take people down a pretty chaotic path if they try to use it. 
3. For docs and examples, I think we should call this "standalone queries" or something along those lines, not external. To me, External reads like these queries are being governed in some wholly different system. 

```
sources: 
   - query_one: path/to/query.sql 
```

packages/preprocess/src/extract-queries/extract-queries.cjs
ItsMeBrianD: Ah yes this should return

packages/preprocess/src/utils/get-route-hash.cjs
ItsMeBrianD: It likely does; this is behavior that already existed, so I left it as is. We can attempt to change it if we think it's in scope, but I would lean towards filing an issue for it.

packages/preprocess/src/utils/get-route-hash.spec.cjs
ItsMeBrianD: This is more about ensuring that the route is processed exactly the same way twice, but adding a test case for a hardcoded string isn't a bad idea.

packages/preprocess/src/process-frontmatter/handle-og.svelte
ItsMeBrianD: In some cases; if metadata is _never_ assigned (e.g. not even set to undefined), then referencing it at all causes an error, doing it this way ensures that it will not break even if metadata doesn't exist at all.

packages/preprocess/src/process-frontmatter/parse-frontmatter.cjs
ItsMeBrianD: If frontmatter is defined; then we know it is addressable as an array; if you go out of bounds in an array, it comes back as undefined; which fails this conditional:

```
Welcome to Node.js v16.19.1.
Type ".help" for more information.
> let x = []
undefined
> x[1]
undefined
```

packages/preprocess/src/utils/get-route-hash.cjs
Winterhart: Let's keep it for now, and have a code-debt for that

sites/docs/docs/core-concepts/queries/index.md
```diff
@@ -81,4 +81,42 @@ Some SQL dialects require sub-queries to be aliased, including Postgres and MySQ
...
+
+## SQL File Queries
```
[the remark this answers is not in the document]
archiewood: Or maybe we can just circumvent this whole thing by calling this feature "SQL Files". Stream of consciousness stuff here!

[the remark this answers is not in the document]
ItsMeBrianD: imo SQL Files leaves the least room for ambiguity, but inline may need to have a more explicit name, maybe `Markdown Queries` and `SQL File Queries`?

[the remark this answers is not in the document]
archiewood: Yeah makes sense to me. 

`Markdown Queries` is a bit confusing. Suggests the syntax for the query is markdown.

Maybe it's all fine as is, and we just call this section SQL files. (or .sql files)

mcrascal:
- [ ] @archiewood to review
- [ ] @ItsMeBrianD to confirm vercel deploy

PR #735, opened 2023-04-04
Prerelease Version Changes

### LINKED ISSUE #685: Pre-release unable to resolve peers automatically
**Bug Description**
Pre release doesn't use sequential version numbers, and as a result a package installing `evidence: next` won't reliably get the correct peer evidence packages, requiring a manual work-around to test pre-releases.

ItsMeBrianD:
### Description

Resolves #685 

#### Previous Behavior

Prerelease command leads to version `0.0.0-${GIT_HASH}`

#### New Behavior

Prerelease command leads to `${package version}.${prerelease number}`
This includes the current semvar version; which should make it more reliable, and also uses a numbered prerelease id. The only issue here is that it will not have the git hash associated with it, and it is unclear if we can have multiple npm tags (e.g. have `next` and `${GIT_HASH}` as tags.


Winterhart (commented):
So if I understand, we'll have a prerelease versioning similar to: `15.0.1.1` for a components on `15.0.1`?

ItsMeBrianD:
> So if I understand, we'll have a prerelease versioning similar to: 15.0.1.1 for a components on 15.0.1?

Correct

> Also, have you tried the changeset prerelease feature?

I have not

Winterhart:
The potential issue with this versioning. (if I remember correctly) With `npm`if someone uses `^15.0.1` they will be on `next` since `^` will auto-upgrade them

ItsMeBrianD:
I've adjusted this to use the format `x.y.z-next.a`, which should resolve the issue with `^` selecting the `next` version. I looked at how sveltekit did their prerelease versioning and it looks like they did `1.0.0-next.x`; even with new prereleases, despite being on `1.15.x` right now. It may be worth considering handling `1.0.0-next.x` instead of this system, but it's a little unclear how that would be automated with ci

Winterhart (approved):

archiewood:
What's the plan with this?

Issue #739, opened 2023-04-06
Migrate default layout and settings page to tailwind


mcrascal:
Bring spacing, sizing, colors into accordance with tailwind's spacing system

Clean up regressions: 

- [x] Sticky header should be full width in mobile #749 
- [x] Kebab (three dot) menu: shadows toned down, click outside to dismiss, constrained to article width on wide screen, remove "get help on slack" in production builds #749 
- [x] Settings #736 

Related: 
- #533 
- #610

nidhi-kala:
@mcrascal If this is open for contributions, I'm keen to pick this up and also look into #736

mcrascal:
Thanks @nidhi-kala -- that'd be awesome. 

If you're open to picking up #736, I think that would be a better place to start, since there is some work already going on re: the core layout on #749.

nidhi-kala:
Great! I will start working on #736

PR #749, opened 2023-04-10
Dropdown menu UI fixes

archiewood:
### Description
This addresses some of the regressions in https://github.com/evidence-dev/evidence/issues/739
- Click outside the menu to close
- Expand header bar to 100% on narrow viewports
- Tone down shadow
- Remove get help in production

### Checklist

#### Click outside to dismiss
CleanShot 2023-04-10 at 19 25 38


archiewood:
Todo:
- [ ] Make all drowdown menu items same height (currently `<a>` and `<button>` elements are slightly different heights )

archiewood:
The click outside to dismiss solution feels a bit gnarly. Would take advice on this if there are other thoughts!

ItsMeBrianD:
Regarding click outside, here's a library action: https://www.svelteui.org/composables/use- -click-outside, and here's one that I've worked on previously https://github.com/SprocketBot/sprocket/blob/staging/clients/web/src/lib/utils/actions/clickOutsideAction.ts

mcrascal:
To do: 

- [ ] Look at constraining to article width (e.g. the kebab should be vertically aligned with the right margin of the article, and potentially constraining breadcrumbs alongside it.

archiewood:
I have a version of the header which constrains the menu and breadcrumbs to article width.

### Before
CleanShot 2023-04-12 at 14 57 45

### After
CleanShot 2023-04-12 at 14 54 25


I'm not totally convinced this is better. Some examples:
## Notion
image

## Streamlit
image

I think the problem here we're trying to solve is making the menu more obvious. Open to suggestions, but some ideas:
- Make the icon bigger
- Make the icon darker grey
- Change the icon for something more visually eye-catching
- Move some menu items out of the menu
    - Perhaps show/ hide queries as separate icon
    - A generic share icon, that currently only has one option, Export PDF.

ItsMeBrianD:
@archiewood this may be a helpful tailwind utility for maintaining this behavior:
https://tailwindcss.com/docs/container

mcrascal:
I'm persuaded for full width. 

Is the preview deploy current? Click outside not working for me in safari.

archiewood:
ah yeah, me neither. shoulda tested cross browser.

Feels like @ItsMeBrianD's clickoutside action it probably a more robust way to go

sites/example-project/src/components/ui/PageMenu.svelte
```diff
@@ -27,14 +32,38 @@
...
+		}
+		update({ enabled: initialEnabled });
+		return {
+			update,
+			destroy() {
+				window.removeEventListener('click', handleOutsideClick);
+			}
+		};
+	}
 </script>
```
ItsMeBrianD, line 59: This should get moved to it's own file; it's something that we are probably going to want to use in other places.

ItsMeBrianD (changes requested):
Looks great, just need to extract clickOutside to it's own file and this should be good to go

PR #760, opened 2023-04-13
lint: ignore changeset directory

ItsMeBrianD:
### Description

Linting the changeset directory doesn't really add value; and just creates friction - so we should not do that. 

### Checklist

- ~~[ ] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)~~

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -20,3 +20,4 @@ yarn.lock
 /packages/components/*
 **/sites/docs/*
 **/dist/*
+.changeset/*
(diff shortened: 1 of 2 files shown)


archiewood:
Yeah this makes sense. 

It complained about that, and I wasn't entirely sure if it would break the changesets automation, so i ignored it

archiewood (approved):
It's also complaining about some indenting and spacing in a markdown file.

Do we want to lint markdown? Probably yes?

PR #762, opened 2023-04-14
Move Show/Hide SQL to own button

archiewood:
### Description

Move back out of menu - users have reported it's hard to find inside the page menu.

CleanShot 2023-04-13 at 22 06 26

### Checklist

### DIFF
--- a/.changeset/silly-comics-prove.md
+++ b/.changeset/silly-comics-prove.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
+move Show/Hide queries button out of page menu into own button
--- a/sites/example-project/src/components/icons/SQLHideIcon.svelte
+++ b/sites/example-project/src/components/icons/SQLHideIcon.svelte
@@ -1,0 +1,44 @@
+<script>
+	export let height = 24;
+	export let width = 24;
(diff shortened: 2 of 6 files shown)


mcrascal:
Thanks @archiewood, this is cool, but I'd propose we hold off on this for now.

1. I think it is possible that it is too hard to find the option in the kebab, but I'd rather wait for some more feedback to roll in before reverting. 
2. If we do pull it back out, I think a less abstract "show / hide queries" button or link is probably preferable to having "SQL" badged on every report.

archiewood:
1. Sounds good. Makes sense to me
2. I played around with a couple of things. Like an eye icon, button with text etc.

This felt like a pretty good trade off on:
- compactness
- not too in your face
- obvious what it does

archiewood:
I think it's also possible that the right place for this _in production_ is:
- query level control of show hide.
- buttons on components "show source" that allow end users to show the data behind a specific component

mcrascal:
Makes perfect sense, I think we'll end up revisiting this. My concern with just badging it "SQL" is more for the completely non technical stakeholder. If was sending a report to someone senior or a sales person or whoever, I would not want it to say SQL or PYTHON or SCALA in the top right.

Issue #603, opened 2023-01-27
Override URIs to non-root path - svelte.config.js


BenJoyenConseil:
**Feature Description**
I would like to override sveltekit's `base` URI and others settings like `outDir` when I run `evidence build`. The [documentation of sveltekit ](https://kit.svelte.dev/docs/configuration#paths) suggests to use `svelte.config.js,`

**Goal of Feature**
I would like to deploy a static evidence's build on Github Pages or something equivalent. Github Pages requires the urls to be relative or following the URI of the repository, so I need to set the sveltekig --base parameter. But it is not suffisant because the URIs in the Evidence.dev html (ex `/favicon.ico`) are all hard coded to point to the root "/".

**Current Solution**
The current source code in cli.js does not allow to override `svelte.config.js`, it overrides the `svelte.config.js `everytime we run `evidence build`

So my current solution is using unix's find to track all urls
```
mv -R ./build ./public
find ./public/* -type f -not -path '*.ico' -exec sed -i '' "s,=\"/,=\"/my_github_uri/,g" {} +
```

aaronsteers:
@BenJoyenConseil - I came here looking for help on the same. My issue logged below seems to describe the same issue you describe above.

- https://github.com/MeltanoLabs/jaffle-shop-template/issues/6

aaronsteers:
With help from @archiewood [here](https://github.com/MeltanoLabs/jaffle-shop-template/issues/6), I resolved my 404s by rending subpage links with a preceeding `'./'` instead of preceeding `'/'`. Subpages still don't render correctly.

mcrascal:
Hi all, 

We took a look at supporting this, see: #764, and are not going to move ahead for the time being. 

While exposing the base path configuration option is straightforward, supporting it across our entire builtin component library, and maintaining it is too much of a headache relative to the other things we'd like to be focusing on at the moment. 

We'll likely revisit this, since there are use cases for this beyond just GH pages, but in the meantime netlify and vercel are both great deployment targets for public projects.

svrooij:
And I guess you can just map your repo website to a subdomain, while still being hosted at GitHub Pages? Then it only requires you to setup a subdomain, which requires 1 additional file in the repo and some dns changes

jdimmerman:
For what its worth, I was able to get this working by doing the following, but it is a cat and mouse game each time we upgrade Evidence/svelte/vite so maybe not worth it.
[...]

PR #764, opened 2023-04-17
#603 Root URL Configuration

ItsMeBrianD:
### Description

This pull request includes experimental support for `EVIDENCE_BASE_URL`; which enables hosting with Github pages.

There are a lot of moving pieces here, so this needs a pretty thorough run-through. 

### Checklist


sites/example-project/src/components/ui/Sidebar.svelte
```diff
@@ -9,12 +10,14 @@
 
 	// children of the index page
 	let firstLevelFiles = fileTree?.children;
+
+	$: console.log(fileTree.children);
```
Winterhart, line 14: A forgotten `console.log` ?

sites/example-project/src/components/ui/Sidebar.svelte
```diff
@@ -38,7 +43,7 @@
 				<a
 					href="/settings"
 					class="settings-link"
-					class:selected={$page.url.pathname === '/settings'}
+					class:selected={$page.url.pathname === `$/settings`}
```
Winterhart, line 46: Are we sure we want that `$`?

sites/example-project/src/pages/text-and-metrics/markdown/+page.md
```diff
@@ -1,3 +1,5 @@
+<script> import { base } from "$app/paths"; </script>
```
Winterhart, line 1: So if I understand, if people wants `links` in their reports they will have to import `base`?

Winterhart (commented):
Few questions.. wow it's a lot of changes!

sites/docs/docs/deployment/github-pages.md
```diff
@@ -0,0 +1,24 @@
+---
+    hide_table_of_contents: false
+    title: Github Pages
+---
+
+:::caution
+Deploying to Github pages requires more javascript knowledge than alternatives like [Evidence Cloud](/deployment/evidence-cloud) [Netlify](/deployment/netlify) or [Vercel](/deployment/vercel), if you are unsure, consider using another method.
```
archiewood, line 7: I would say "...requires some JavaScript, unlike alternatives, such as..." as many simple evidence apps need no js.

sites/docs/docs/deployment/github-pages.md
```diff
@@ -0,0 +1,24 @@
...
+:::
+
+## Deploying to Github Pages
+
```
archiewood, line 11: Is it worth a short para on why?

GitHub pages are typically deployed with a base path that is not `/` eg ..., which is a non standard configuration, and requires some working around.

sites/docs/docs/deployment/github-pages.md
```diff
@@ -0,0 +1,24 @@
...
+Deploying to Github Pages requires a few additional steps and considerations compared to other providers.
+
+First, [configure githug pages for your repository](https://docs.github.com/en/pages/quickstart).
+
+Once Github pages has been set up; build your project with the environment variable `EVIDENCE_BASE_PATH` set to your repositories name; note that it must start, but not end, with a `/`. This will configure evidence to point built-in links to `/your-base-path` instead of `/`. (`EVIDENCE_BASE_PATH={your repository} npm run build`)
+
+Finally, whenever you create a relative link (e.g. one that starts with `/` and navigates to a page in your evidence project), you must prefix it with `{base}`. This will ensure that the link functions properly when deployed to github pages.
```
archiewood, line 18: Is this the definition of a relative link? Sounds like an absolute link to me.

What about links like ../../my-cool-analysis/

packages/evidence/scripts/build-template.js
```diff
@@ -50,6 +50,9 @@ fs.outputFileSync(
             files: {
                 routes: 'src/pages',
                 lib: 'src/components'
+            },
+            paths: {
+                base: process.env.EVIDENCE_BASE_PATH ?? undefined
```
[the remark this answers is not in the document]
ItsMeBrianD: No if it is unset the behavior remains the same as before

sites/example-project/src/pages/text-and-metrics/markdown/+page.md
ItsMeBrianD: It was added to preprocess so it shouldn't be required, just needed in the example project

mcrascal:
@ItsMeBrianD thanks -- are there any measures we can take to ensure we don't hard code relative urls into our components in the future, and break GH deployments?

Winterhart (changes requested):
Code looks great, few minor `console.log` to remove.

 I've just tested in `localhost` I got one issue.
Here's how I can reproduce it

- I click on a normal page `Linear Chart`
- I click on `hierarchical Nav` on the side menu
- Then if you navigate back to `Linear Chart` it's not working

In the terminal:

```
3:03:12 PM [vite-plugin-svelte] /src/pages/hierarchical nav/+page.md:123:9 'base' is not defined
```

On the browser:

Screenshot 2023-04-27 at 3 08 00 PM

archiewood:
This isn't so much a code review as more of a philosophical point

- This is a pretty far reaching change, and impacts how we'll develop components and features in the future.
- I worry it could be somewhat fragile, and increase the surface area of our product for edge cases.
- Github pages is a pretty common SSG, but it's by no means the only one, nor the most powerful.
  - Even when we do support it, the GH pages experience will be worse than others - you'll have to append the base path manually to each to get it to work.
  - Evidence becomes less "deploy anywhere" - the same source code will not do the same thing on all the platforms we support.

TLDR svelteKit decided not to support this out the box, and I wonder if we should be taking a similar approach.

I'd love to hear other opinions on this, but I wonder if this is a reasonable increase in complexity for not enough benefit

ItsMeBrianD:
> @ItsMeBrianD thanks -- are there any measures we can take to ensure we don't hard code relative urls into our components in the future, and break GH deployments?

We can add an additional check to run a build with the `EVIDENCE_BASE_PATH` environment variable set; the build will likely break if the base path isn't respected.

> This is a pretty far reaching change, and impacts how we'll develop components and features in the future.

> I worry it could be somewhat fragile, and increase the surface area of our product for edge cases.

This definitely creates a fair amount of development overhead, so I could see this going either way

> Evidence becomes less "deploy anywhere" - the same source code will not do the same thing on all the platforms we support.

If you include `{base}`, and then build without the `EVIDENCE_BASE_PATH`, it will behave as if `{base]` isn't there; so in theory this shouldn't impact where you can deploy a project.

---

As to the overall viability of this feature; I think it really depends on how large the use-case / demand for Github Pages feature is. It does remain one of the easiest places to stick a demo / project.

ItsMeBrianD:
@Winterhart I can't reproduce the issue that you are describing, did you rebuild the `preprocess` package after checking out the branch?

Winterhart (commented):
I just re-run the branch and re-build all the components. 

The error seems gone 🥳  I get a `undefined` alert on the page is it normal?

CleanShot 2023-04-27 at 17 52 17

mcrascal:
@ItsMeBrianD 

I would propose we put this on the shelf for now. After reviewing: 

1. Even in the nav and UI components here, we've got a handful of issues cropping up. All look like they'd be quick to fix, but this is such an annoying gotcha to leave behind for ourselves and contributors.  
2. It seems like this is breaking svelte kit's pre-load on-hover behaviour. 

In terms of actual user impact, I think this is pretty small. GH pages are a nice-to-have. 

Next step would just be to document the lack of GH pages support in a separate PR. 

We will likely need to tackle this for some trickier customer-embedded cases, so I doubt this will all be for nothing. I had originally preferred this solution over some sort of middleware / preprocessor approach, but I think when we do decide to revisit this, it'll be worth a more detailed look at one of those approaches.

ItsMeBrianD:
@mcrascal Sounds good; I'll open another PR with the documentation for that.

sites/example-project/src/components/ui/Databases/DatabaseSettingsPanel.svelte
```diff
@@ -43,7 +44,7 @@
 	async function runTest() {
 		const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
 		await sleep(1000);
-		const res = await fetch('/api/testConnection.json', {
+		const res = await fetch(`${base}/api/testConnection.json`, {
```
jtshafer, line 47: why not just use `./`

PR #812, opened 2023-04-25
#724 Fix Environment Variable Collisions

### LINKED ISSUE #724: Environment variable name collisions
**Bug Description**
A customer found that there is a name collision between:
- defining what data warehouse / DBMS you are using (environment variable `database` or `DATABASE`). See [here](https://github.com/evidence-dev/evidence/blob/332b48b3bbb291c2a1a9569de8307eb9cc168de2/packages/db-orchestrator/index.cjs#L49)
- defining which database is being referenced in the DW. (environment variable options include `database`, `DATABASE`, `SNOWFLAKE_DATABASE`. See an example for snowflake [here](https://github.com/evidence-dev/evidence/blob/332b48b3bbb291c2a1a9569de8307eb9cc168de2/packages/snowflake/index.cjs#L113)

**Severity**
- [x] Medium - I'm able to use Evidence, but a feature I need to use is not working

**Potential solutions**
These are both breaking changes for users. Any user that upgrades and is using different env vars would need to update their settings. We could do one or both:

1. **Remove the redundancy in environment variables.** Only allow one for each setting. 
   - If so, the only options for defining the db in the warehouse should probably be `SNOWFLAKE_DATABASE`, `POSTGRES_DATABASE` etc, as they are explicit
2. **Rename the variable that we use for the connector DBMS.**
   - Change the setting env var (and probably name of variable too), from `DATABASE` to eg `DATASOURCE`

**Workarounds**
Use one of the other environment variable options that does not collide.
For example:
- DBMS: `DATABASE: snowflake`
- Database in the warehouse: `SNOWFLAKE_DATABASE: production_data_db`

**Database:**

Any DBMS that has a concept of a `database`
- [x] Postgres
- [ ] BigQuery
- [x] Snowflake
- [x] Redshift
- [x] MySQL
- [ ] SQLite
- [ ] DuckDB
- [ ] Other (please specify)

ItsMeBrianD:
### Description

Resolves #724 

### Checklist


packages/bigquery/index.cjs
```diff
@@ -1,5 +1,29 @@
...
+			{ key: 'CLIENT_EMAIL', deprecated: true }
+		],
+		private_key: [
+			{ key: 'EVIDENCE_BIGQUERY_PRIVATE_KEY', deprecated: false },
+			{ key: 'BIGQUERY_PRIVATE_KEY', deprecated: false },
+			{ key: 'private_key', deprecated: true },
+			{ key: 'PRIVATE_KEY', deprecated: true }
+		]
+	}
+};
```
ud3sh, line 27: RE: `private_key`, `client_email` etc.  
It seems to me that we are using a mix of camel or snake for our var names. Should we settle on one?  (Not specific to this change).

packages/db-commons/src/getEnv.spec.cjs
```diff
@@ -0,0 +1,68 @@
+const { getEnv } = require('./getEnv.cjs');
```
ud3sh, line 1: +1 for tests

packages/db-commons/src/getEnv.cjs
```diff
@@ -0,0 +1,32 @@
...
+					);
+				return process.env[k.key];
+			}
+		}
+	} else {
+		throw new Error(`Could not find ${key} in env map!`);
+	}
+};
+
+exports.getEnv = getEnv;
```
ud3sh, line 32: Looking at the tests, it seems to me that the `key` parameter is suppose to represent a hierarchy.  e.g for the following `envMap`, if we want `2nd`, we ought to call `getEnv(envMap, ['1st', '2nd'])`.  I think some of the vars/params in getEnv can be renamed to make this a bit more self documenting.

		const map = {
			'1st': {
				'2nd': [{ key: 'my-key', deprecated: false }]
			}
		};

ItsMeBrianD: In one or two places nesting was used; down for that rename though

packages/bigquery/index.cjs
ItsMeBrianD: Ah yeah I hadn't noticed this; should probably be updated to be camel in this case.

ItsMeBrianD:
@archiewood I think one more pass; but everything should be good to go here

archiewood:
I can test if you need, what's the ideal protocol?

archiewood:
Okay so using:
image

I'm getting:

image

Should I also be getting warnings about?
- DUCKDB_FILENAME
- DUCKDB_GITIGNORE_DUCKDB

I notice that DUCKDB_FILENAME is **not** marked as deprecated, so perhaps this is as you intended?

ItsMeBrianD:
> `DUCKDB_GITIGNORE_DUCKDB`

I don't know if this environment variable is used anywhere; I searched through the project and didn't find anything.

> `DUCKDB_FILENAME`

This was not flagged as deprecated; we can change that if we'd like to

ItsMeBrianD:
> there are all only used in production though, so I assume I need to test a deployment with different permutations of the above variables them?

It could be done this way, or you can build locally;

If you create a `.env` file with `export SOME_ENV_VAR=SOME_ENV_VALUE`, then run `. .env`, that will load the environment variables, you should be able to use `npm run build` and it will behave like a production build

archiewood:
> This was not flagged as deprecated; we can change that if we'd like to
Not sure it's that important

I'm done on this from my side

archiewood (approved):

Issue #815, opened 2023-04-26
Waterfall Chart


hughess:
A waterfall chart is great for showing decomposition of changes.

I've had a start at creating one (which I will open in a draft PR for those who are interested in taking a look).
<img width="699" alt="CleanShot 2023-04-26 at 10 42 45@2x" src="https://user-images.githubusercontent.com/12602440/234636102-dbd43f3d-8d11-4d5b-950d-32f745dcc805.png">

<img width="597" alt="CleanShot 2023-04-26 at 10 57 15@2x" src="https://user-images.githubusercontent.com/12602440/234636122-e94cbd04-926e-40b9-a21b-2c6a66de0c44.png">

## Considerations for implementation
- Axis labels are important on these charts and will often be long strings - the chart needs to be able to handle multi-line labels without cutting off characters or breaking words apart
   - I think it probably needs "breaking" behaviour by word, followed by either truncation (`...`) or expansion of the chart size
   - This will also need to work well responsively and on mobile
- We don't currently have multi-line axis labels in our other charts - this will need some thinking to make sure the chart area is the right size in all cases. Currently, adding multi-line axis labels shrinks the main chart area
- The initial attempt I've made requires a specific data format: 
   - `category` - category to use on x-axis
   - `value` - change to show on chart
   - `total` - boolean - should this be treated as a total/subtotal on the chart
- The SQL required to get data into this format can require a lot of work. We might want to offer other alternatives for the structure of the data getting passed in
   - I don't think this would change the implementation within echarts - we could just add a layer of logic above this to get it into the format echarts requires
- Data labels should be included - positioning these may be difficult   
- I haven't thought about tooltips yet. I'm thinking they will be straightforward though - probably a config override sent from `Waterfall` to `Chart`
- Subtotals - the logic for generating the chart series will need to be updated to accomodate subtotals - currently, having a total in the middle of the data throws off the cumulative sum
- We may want to offer sorting - e.g, sort by magnitude of change. This shouldn't be hard to do with this data format since we can easily identify the totals and exclude them from sorting

matsonj:
> The SQL required to get data into this format can require a lot of work. We might want to offer other alternatives for the structure of the data getting passed in

I think including a sample SQL query in the documentation would go a long way. It is indeed fairly complex but you can simplify it massively with clever windowing (LAG / LEAD or cumulative sums). I think having a clean example of the data model on this one is key because it always takes me like an hour in power BI to get these right.

archiewood:
[...]

PR #816, opened 2023-04-26
Waterfall chart

hughess:
### Description
Draft of waterfall chart approach. Feel free to play around with this and discuss here or in the issue #815 
CleanShot 2023-04-26 at 10 42 45@2x

CleanShot 2023-04-26 at 10 57 15@2x

### To Do
- Axis label overflow behaviour + behaviour on mobile
- Ensure chart area is right size when multi-line axis labels are included
- Consider alternative data structures that the component can accept
- Data labels with reasonable positioning and overflow behaviour
- Tooltip support
- ~~Subtotals - the logic for generating the chart series will need to be updated to accomodate subtotals - currently, having a total in the middle of the data throws off the cumulative sum~~
- Sort feature, by magnitude of delta?
- In a lot of cases, people need to start the y-axis at a higher number than 0 to make the changes visible enough on a waterfall chart
   - We could offer a `yMin` prop to allow someone to set it manually
   - I could also imagine some dynamic logic to check the change range (max of cumulative sum - min of cumulative sum) as a % of axis range and adjust the `yMin` so that the change range takes up something like 50% of the vertical chart area

### Checklist

### DIFF
--- a/packages/preprocess/src/process-queries.cjs
+++ b/packages/preprocess/src/process-queries.cjs
@@ -45,6 +45,7 @@ const createDefaultProps = function (filename, componentDevelopmentMode, fileQue
         import FunnelChart from "$lib/viz/FunnelChart.svelte";
         import SankeyChart from "$lib/viz/SankeyChart.svelte";
         import ScatterPlot from '$lib/viz/ScatterPlot.svelte';
+        import WaterfallChart from '$lib/viz/WaterfallChart.svelte';
         import Histogram from '$lib/viz/Histogram.svelte';
         import ECharts from '$lib/viz/ECharts.svelte';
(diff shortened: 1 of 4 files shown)


archiewood:
@hughess can I close this for now

hughess:
@archiewood I'm going to take one more look at it this week to see if there's a minimal version that can become shippable quickly. Either I'll make those changes so it can be shipped this week, or I'll close it after taking a look.

Interested in any feedback on which of these things should be required for a minimal shippable version:
- Data labels
- Tooltips
- Axis label size management (overflow onto new lines, truncate at a certain point)
- Support for negative waterfalls (I think relevant only for percentage bridges)

hughess:
Moving to labs. It will come back when it's ready!

PR #820, opened 2023-04-27
Make download buttons non-copyable

### LINKED ISSUE #819: Download buttons copy and paste with content
**Bug Description**
When copying and pasting content from Evidence into other programs (like Google Docs), text from download buttons (download data, save image) will get pasted with the content.

Would expect text in UI buttons to not be selectable.

**Severity**

hughess:
### Description
Fixes #819 

### Before
CleanShot 2023-04-27 at 13 41 43

### After
CleanShot 2023-04-27 at 13 42 48

### Checklist

### DIFF
--- a/.changeset/mean-baboons-punch.md
+++ b/.changeset/mean-baboons-punch.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
+Make download buttons non-copyable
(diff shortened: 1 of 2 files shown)


mcrascal (approved):
lgtm

hughess:
I've just realized that the content can still be copied, just not selected. So if you finish your copy motion while hovered over a chart, it will copy the button contents. 

I think that should be a pretty rare occurrence though, so I don't think it's worth making another change.

PR #826, opened 2023-05-01
fixed chart width when printing to pdf

### LINKED ISSUE #482: Chart width does not match page width when printing to PDF
**Bug Description**
When you click to print a page to PDF, charts on the page will often not line up exactly with 100% of the page width. Most often, it seems that they extend beyond the width of the page (see example screenshot below).

Depending on the width of the page when the PDF button is clicked, the chart may appear as narrower or wider than 100% of page width. 

**Severity**
- Low - I'm able to keep using Evidence as normal, but flagging this for the team to fix

**Expected Behavior**
Charts should take up exactly 100% width.

Chart in screenshot below extends wider than 100% page width:
CleanShot 2022-12-01 at 19 09 21@2x

csjh:
### Description
Fixes #482 

### Before
image

### After
image

### Checklist

### DIFF
--- a/.changeset/curly-apes-greet.md
+++ b/.changeset/curly-apes-greet.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
+Fixed chart width not adjusting properly when printing to PDF
--- a/sites/example-project/src/components/modules/echartsCopy.js
+++ b/sites/example-project/src/components/modules/echartsCopy.js
@@ -421,7 +421,7 @@ export default (node, option) => {
 		}
 	});
 
-    const { config, ratio } = option;
+	const { config, ratio } = option;
 
 	const chart = init(node, 'evidence-light', { renderer: 'canvas' });
 	config.animation = false; // disable animation
--- a/sites/example-project/src/components/modules/echartsCopy.js
+++ b/sites/example-project/src/components/modules/echartsCopy.js
@@ -421,14 +421,16 @@ export default (node, option) => {
 		}
 	});
 
-	const chart = init(node, 'evidence-light', { renderer: 'canvas' });
-	option.animation = false; // disable animation
+    const { config, ratio } = option;
 
-	chart.setOption(option);
+	const chart = init(node, 'evidence-light', { renderer: 'canvas' });
+	config.animation = false; // disable animation
+
+	chart.setOption(config);
 
 	let src = chart.getConnectedDataURL({
 		type: 'jpeg',
-		pixelRatio: 2,
+		pixelRatio: ratio,
 		backgroundColor: '#fff',
 		excludeComponents: ['toolbox']
 	});
--- a/sites/example-project/src/components/modules/echartsCopy.js
+++ b/sites/example-project/src/components/modules/echartsCopy.js
@@ -436,5 +436,5 @@ export default (node, option) => {
 	});
 
 	// Replace the contents with an img tag
-	node.innerHTML = `<img src = ${src} width=100%/>`;
+	node.innerHTML = `<img src=${src} width="100%" style="position: absolute; top: 0" />`;
 };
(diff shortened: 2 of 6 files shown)


hughess:
@csjh this is great!

I noticed something interesting - I get the same results as you when I use Ctrl + P in Chrome, but when I click the "Export PDF" button, I get the overflow like before. I assume that button isn't getting identified as "printing"?

ItsMeBrianD (approved):
🚀

csjh:
> @csjh this is great!
> 
> I noticed something interesting - I get the same results as you when I use Ctrl + P in Chrome, but when I click the "Export PDF" button, I get the overflow like before. I assume that button isn't getting identified as "printing"?

Seems like calling `window.print()` doesn't fire the events, and `window.matchMedia("print")` fires too late for it to affect the PDF. For now I guess changing the Extract PDF functionality to
```js
function print() {
    window.dispatchEvent(new Event("beforeprint"));
    setTimeout(() => window.print(), 0);
    setTimeout(() => window.dispatchEvent(new Event("afterprint")), 0);
}
```
should work albeit a bit hacky. The setTimeouts are needed because the print window was beating `beforeprint` to the punch.

ItsMeBrianD:
I would consider firing a custom named event in the window to avoid any issues with other event handlers expecting event data; should just be another handler with the others

hughess:
Looks really close. There's 2 issues left as far as I can tell - 1 that is related to this, and 1 that may be a separate issue.
- The resolution of the charts in the PDF is quite low after the change to the image, rather than using the SVG as before
   - Is there any way to use an SVG with the same method?
   - I recall we ran into issues with SVG and PNG for the copy/paste feature, which is why the echarts copy target uses JPEG
   - Since this is a different case than copy/paste, we may need to create another echarts target to support PDFs to get a high enough resolution image 

The possibly unrelated issue:
- I wasn't able to realize this until I saw the charts at the same width as the query viewers on a PDF - the article seems to be squished a bit and I believe it's not taking up 100% of the printable width of the page
  - One way to notice this is to compare the aspect ratio of the chart on an exported PDF vs. on a webpage

csjh:
> * Is there any way to use an SVG with the same method?

I've looked around a bit and it doesn't seem like it's possible to get the SVG to play nice. Seems better to just up the `pixelRatio` on the rendered image.

As for the squishing, it seems like the `article` in `+layout.svelte` isn't taking up as much width as it should, but I do think that's a separate issue.

hughess:
Nice, the new resolution looks great. It also seemed to fix the aspect ratio issue somehow.

One new issue I noticed. On the Misc Analyses > US Bank Stats page, if you print to PDF, some of the elements will start to overlap further down the document, which didn't happen previously:
https://evidence-development-workspace-git-482-char-1250e1-evidence-dev.vercel.app/misc-analyses/us-bank-stats/

CleanShot 2023-05-01 at 19 18 38@2x

I assume this has something to do with the media queries on the Chart, which allow the horizontal bar chart to get printed across multiple pages.

csjh:
I think I finally cracked it, the `img` was refusing to stay inside its div, so a `position: absolute; top: 0` will hopefully stop it from fighting back.

There appears to be a slight over-adjustment for the height, so the div has a bit of extra empty space on the bottom. Could probably do some calculations based on the aspect ratio via something like this:
```js
node.innerHTML = `<img src=${src} width="100%" style="position: absolute; top: 0" />`;
const rect = node.getBoundingClientRect();
node.style.height = `${rect.height / rect.width * 512}px`;
```
but I think the 512 would have to change for every paper size, which I'm not 100% is available to Javascript... Realistically the difference isn't noticeable on graphs that aren't big (i.e. don't span multiple pages)
image

ItsMeBrianD (approved):
Looks good!

PR #827, opened 2023-05-01
docs: note sql file aliases

### LINKED ISSUE #769: Document named standalone .sql in frontmatter

ItsMeBrianD:
Adds description of aliased queries to the documentation.

Resolves #769


sites/docs/docs/core-concepts/queries/index.md
```diff
@@ -118,3 +118,14 @@ In your evidence file, you can now reference `my_query` and `some_category_my_ca
 ### Advanced Usage
 
 SQL file queries can [depend on other query files](/core-concepts/queries/#query-chaining), but they will all need to be referenced in the files you use them in. For example, if `my_query` depends on `some_category_my_category_query`, then you will have to have them both in your [frontmatter](/markdown/#frontmatter), as shown above.
```
archiewood, line 120: I'd add a subtitle here: ## File Query Chaining

sites/docs/docs/core-concepts/queries/index.md
```diff
@@ -118,3 +118,14 @@ In your evidence file, you can now reference `my_query` and `some_category_my_ca
...
+
```
archiewood, line 121: subtitle here ## Aliasing File Queries

sites/docs/docs/core-concepts/queries/index.md
```diff
@@ -118,3 +118,14 @@ In your evidence file, you can now reference `my_query` and `some_category_my_ca
...
+If the name of your query file is long, or doesn't match how it is being used on the page; it is easy to change how it is refernced on the page. Instead of importing your query as ` - query.sql`, you can import is as ` - renamed: query.sql`. Now instead of the data being `{query}`, it is `{renamed}`.
+
```
archiewood, line 123: Using the name `renamed` and `alias` here slightly confused me.
Initially I thought that you had to use the key `renamed`.

Maybe we can use a more realistic example in the code snippet below, using `orders_per_month_exc_b2b_us_canada.sql` and `monthly_orders` or similar

archiewood (commented):
Added a coupla comments

mcrascal:
+1 on @archiewood's comments. 

I would also not bundle the naming behaviour under advanced usage, I would introduce it as the main example, and just note that by default if you import a query without a name, it will be named using the 

Named import is the path we should be taking folks down -- it'll be much easier to maintain their projects with this behaviour (e.g. re-organizing standalone sql queries will only result in one required edit to the import statements, not touches to every query reference).

archiewood (approved):

PR #847, opened 2023-05-04
747 component copy paste not working firefox

csjh:
### Description

Firefox was copying to the clipboard before Svelte had time to update the content to match, so I made the `copy` event force Svelte to update synchronously via `flush()`.

Note that Firefox can't seem to keep the shape of the `BigValue` as well as other browsers, not sure why.

### Before
(content pasted into gmail)
image

### After
image

### Checklist


archiewood:
Had a look at firefox vs chrome here.


Not sure we need to get these exactly right. There are quite a few differences in the outputs that may be worth investigating in the future, but feels like a separate issue.


| Firefox + Google Docs | Chrome + Google Docs |
|--|--|
| Clean Shot 2023-05-04 at 14 35 11 | CleanShot 2023-05-04 at 14 34 42 |

Issue #788, opened 2023-04-21
Value component error icon


hughess:
**Bug Description**
The value component error icon is currently black, but should be white to match the text in the error.

The icon should also be vertically centered

![value-error-icon](https://user-images.githubusercontent.com/12602440/233674778-d9c97a52-e2aa-4772-a948-9dde8a4cbc82.PNG)

**Severity**
- [ ] High - I'm not able to use Evidence until this bug is fixed
- [ ] Medium - I'm able to use Evidence, but a feature I need to use is not working
- [x] Low - I'm able to keep using Evidence as normal, but flagging this for the team to fix

nidhi-kala:
fixed this in #873

PR #873, opened 2023-05-15
Value component error icon

nidhi-kala:
### Description

Addressing #788
Value component error icon is white and positioned in the center.

### Checklist

Before: 
image

After:
Screenshot 2023-05-14 at 20 59 59

### DIFF
--- a/sites/example-project/src/components/viz/Value.svelte
+++ b/sites/example-project/src/components/viz/Value.svelte
@@ -85,7 +85,7 @@
 	<span class="error">
 		<span class="error-label">Error</span>
 		<span class="additional-info-icon">
-			<HelpCircleIcon height="18" width="18" verticalOffset="2" color="--grey-000" />
+			<HelpCircleIcon height="18" width="18" verticalOffset="2" color="--grey-100" />
 		</span>
 		<span class="error-msg">{error}</span>
 	</span>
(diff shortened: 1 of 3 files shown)


sites/example-project/src/components/viz/Value.svelte
```diff
@@ -85,7 +85,7 @@
 	<span class="error">
 		<span class="error-label">Error</span>
 		<span class="additional-info-icon">
-			<HelpCircleIcon height="18" width="18" verticalOffset="2" color="--grey-000" />
+			<HelpCircleIcon height="18" width="18" verticalOffset="2" color="--grey-100" />
```
archiewood, line 88: Not 100% sure what's going on here, but I'm not seeing the icon in Safari

## Chrome
CleanShot 2023-05-15 at 08 22 50@2x


## Safari
CleanShot 2023-05-15 at 08 20 21@2x

archiewood (changes requested):

sites/example-project/src/components/viz/Value.svelte
nidhi-kala: Thanks for the review! :)
The cause might be `display: inline`. Without explicitly setting the dimensions, it was probably misrendering SVG in Safari. Fixed that.

archiewood (approved):

PR #874, opened 2023-05-15
Add secondary axis

hughess:
### Description
This draft PR adds a secondary axis to the charting library, called `y2`.

I was playing around with what would be possible here and wanted to add this PR to share the information. 

CleanShot 2023-05-13 at 18 03 40@2x

This almost works already, but there are some hairy issues left to solve before we could implement this:
- Get working in composable charts
   - Toughest part here is working around the logic in Chart.svelte which automatically assigns columns to `y` when no `y` prop is supplied
   - This might require somehow passing `y` and `y2` values from child components up to the parent Chart.svelte
- yAxis overrides within child components
   - Most child components adjust the yAxis depending on the series information in the chart (e.g., `Bar.svelte` updates the echarts config by updating a store)
   - These axis adjustments all assume the y axis in the chart config is in one object, but with this change, the y axis could either be a single object or an array containing 2 objects
   - One solution here is to always include a secondary axis, but have it set to hidden - then the override within each child component becomes `show: true` rather than having to trigger a secondary axis to get added to the config (this is how it is set up in this PR at the moment)
- Tooltip formatting
   - It's not straightforward to retrieve the formatting information for a series from echarts' tooltip formatter. This isn't a problem now because everything is formatted as the same as the y-axis and we know the y column format
   - With `y2`, we'd need to know which axis each series belongs to so we can apply the correct formatting
   - There is definitely a way to make this work, but may require some parsing of the series name, or setting up an object with series name and yAxisIndex that can be accessed from the tooltip formatter
   - Issue open with echarts here to see if there is an easier way to access that information: https://github.com/apache/echarts/issues/18622 
- Axis colours
   - It's nice to colour the axis labels to match a series on each axis (e.g., have a red line on `y` with red axis labels, and a blue line on `y2` with blue axis labels
   - Tricky to figure out how to determine which colours to use if:
      - Multiple series exist on at least one of the axes
      - The user has overridden the colour of their line/bar/etc.
   - Solution here might be to offer an override prop to the user so they can choose a specific colour to make the axis if they wish (e.g., `yAxisColor='red'`)   
- How to handle charts where `swapXY=true`
   - Might be best to not allow a secondary axis on horizontal charts, at least as a starting point 
- Annotation components would need to be adjusted to work with `y2` 
- Get it working intuitively with multiple series on one or both axes

### Checklist

### DIFF
--- a/sites/example-project/src/components/modules/getSeriesConfig.js
+++ b/sites/example-project/src/components/modules/getSeriesConfig.js
@@ -4,6 +4,7 @@ export default function getSeriesConfig(
 	data,
 	x,
 	y,
+	y2,
 	series,
 	swapXY,
 	baseConfig,
@@ -13,10 +14,11 @@ export default function getSeriesConfig(
 	size = null,
 	tooltipTitle = null
 ) {
-	function generateTempConfig(seriesData, seriesName, baseConfig) {
+	function generateTempConfig(seriesData, seriesName, yAxisIndex, baseConfig) {
 		let tempConfig = {
 			name: seriesName,
-			data: seriesData
+			data: seriesData,
+			yAxisIndex: yAxisIndex
 		};
 		tempConfig = { ...baseConfig, ...tempConfig };
 		return tempConfig;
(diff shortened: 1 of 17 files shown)


archiewood:
Sweet!

Couple of thoughts:

## Composable Charts

- In reality there are only a couple of combinations where you can do secondary axes without it being impossible to parse the information. 
- For that reason, we may not choose to support all combinations. 

Two that definitely do exist:

### Dual Axis Line Chart
image

### Dual Axis Bar+Line Chart
The bar is either stacked, or single
image

We could start just supporting dual axis on line chart if easier

## Axis colours
The need here is working out which axis corresponds to which series. I think there are three options
- Axis colours (though may work less well if there is more than one series against a particular bar
- Axis _label_ colours (as above)
- Legend for each series, label on each Y axis

## Annotation
- Not sure y2 RefLine and RefArea is a requirement _necessarily_. We could choose to just support highlighting on primary axis

> Work intuitively with multiple series on both axes

Sounds hard, and not 100% sure we should try to support

image

image

hughess:
@archiewood agree with all that. I think a limited implementation is a good idea. Bar + line is a bit tricky because it means it has to work in a composable chart situation, but leaving it available to only 2 series types would make it simpler.

On the multiple series across axes - we'd just need to think through how to give the user feedback - e.g., do we throw an error if there are multiple series on both axes?

archiewood:
If we were doing multi series at all, I think we should only allow multiple series in y. Throw an error if there are multiple series in y2?

I'm not 100% sure how the data would be organized in this kind of situation anyway!

hughess:
Not a chart I would make, but structurally a way it could come up is if you had a table like: `date`, `sales`, `gross profit`, `orders`. And then you used them like:
```html
<LineChart
   x=date
   y={['sales', 'gross profit']}
   y2=orders
/>
```

And then maybe you want to reverse it so orders is on `y` and the other 2 are on `y2`.

Now that I think about it, based on the way I set this PR up, it may actually be more difficult to restrict multiple series usage.

ItsMeBrianD:
@hughess is this still being worked, or gone stale?

hughess:
Stale unfortunately!

hughess:
Closing in favour of #1236

PR #891, opened 2023-05-23
version bump duckdb to 0.8.0

### LINKED ISSUE #885: Support latest duckdb
## Bug Description
The Evidence duckdb connector isn't able to read the newest version of duckdb files. I assume we need a general solution here to handle future cases of this as well - where we have a version that Evidence is supporting, but users who want to upgrade to the newest duckdb version can do that and use it in Evidence.

## Severity
The dbt jaffle shop tutorial relies on a newer version of duckdb and is currently not working with evidence: https://github.com/dbt-labs/jaffle-shop-template

- [x] High - I'm not able to use Evidence until this bug is fixed

csjh:
### Description

Fixes #885 

Adds support for the latest version of DuckDB. Also removes support for older versions. The error message for older database versions is fairly descriptive on how to upgrade (and users of pre-1.0 libraries should be aware of upgrading anyways), so I don't think anything else should be necessary.

### Checklist

### DIFF
--- a/.changeset/fresh-news-refuse.md
+++ b/.changeset/fresh-news-refuse.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/duckdb': patch
+---
+
+version bump duckdb-async to 0.8.0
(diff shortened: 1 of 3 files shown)


archiewood:
I think you'll need to commit an updated version of needful_things.duckdb to get these tests to pass

Issue #898, opened 2023-05-25
Build strict produces 'VITE_BUILD_STRICT' is not recognized as an internal or external command error on Windows 10


RandomFractals:
## Steps To Reproduce

While using [template](https://github.com/evidence-dev/template) starter project and trying to build it with [build:strict](https://docs.evidence.dev/deployment/overview#buildstrict) I get the following error on Windows 10:

![evidence-build-strict-error](https://github.com/evidence-dev/evidence/assets/656833/6d5e761f-30c9-426b-b2fe-a2e27989e866)

The normal Evidence [build](https://docs.evidence.dev/deployment/overview#build) command executes without errors and builds the project for deployment successfully.

### Environment

My NodeJS installation is somewhat older, but meets your [system requirements](https://docs.evidence.dev/getting-started/install-evidence#system-requirements)

- Node version (`node -v`): `v16.17.0`
- npm version (`npm -v`): `8.19.2`
- OS: Windows 10

## Expected Behavior

Evidence project builds successfully in strict mode.

archiewood:
I've just tested `build:strict` on your Node.js and npm version (on MacOS) and it builds successfully, so I suspect the issue lies elsewhere!

PR #899, opened 2023-05-25
Make build:strict windows compatible

Winterhart:
### Description

See #898 

Add the `env` setup in the new `sade` CLI setup.

### Checklist

### DIFF
--- a/packages/evidence/cli.js
+++ b/packages/evidence/cli.js
@@ -121,6 +121,9 @@ const watchPatterns = [
 	{ sourceRelative: '.', targetRelative: './.evidence/template/src/', filePattern: 'app.css' } // custom theme file
 ];
 
+const strictMode = function () {
+	process.env['VITE_BUILD_STRICT'] = true;
+};
 const buildHelper = function (command, args) {
 	const watchers = runFileWatcher(watchPatterns);
 	const flatArgs = flattenArguments(args);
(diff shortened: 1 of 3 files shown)


packages/evidence/cli.js
```diff
@@ -185,7 +188,8 @@ prog
 	.action((args) => {
 		populateTemplate();
 		clearQueryCache();
-		buildHelper('VITE_BUILD_STRICT=true npx vite build', args);
```
ud3sh, line 188: With this change, what happens if I run 
`npm run build:strict` followed by `npm run build`?
would the `VITE_BUILD_STRICT=true` hang around, or would it be cleared because it's a new process?

Winterhart: From my testing on `template` I can run `npm run build:strict` then `npm run build` without issue.

ud3sh: is the env var cleared?   i.e `npm run build` doesn't trigger an actual strict build?

Winterhart: yes, I will send you the demo video.

ud3sh: Oh no need, just wanted to be sure we accounted for this in testing - sounds like we did!  I think to get to this part a child subprocess has to triggered so that would make sense.

PR #955, opened 2023-06-15
Console component

### LINKED ISSUE #920: DDB SQL Console Component
To be initially released as a component plugin, not installed in user projects by default. Smaller alternative to the new page types described in #765. 

Main goal here is to bring ddb wasm out of prototypes and into production ahead of #768

`<SQLConsole data={query_name} defaultQuery='select * from query_name limit 10' />`

* Implemented as a component plugin (e.g. someone could install it in their project) 
* Accepts normal query result JSON input, no parquet, no database connection impacts 
* Runs WASM DDB internally, loads the supplied data into a table 
* Defaults to `select * limit 10` or similar, with option for the user supply it as a prop (e.g. defaultQuery in the above) 
* Re-execute queries on keystroke, command enter, or run query button 
* Pass results to a data table component, or a more stripped down table if data table isn't suitable 
* Loading skeleton acceptable during startup of internal ddb instance, pre-rendering as a stretch goal

**For reference:** 

1. https://github.com/evidence-dev/duckdb-wasm-poc/blob/master/src/routes/%2Bpage.svelte
2. https://github.com/duckdb-wasm-examples/sveltekit-typescript
2. https://excalidraw.com/#room=48f7c829383ba03d96cf,0htpa0HEro3nLCTtQWCh6g

csjh:
### Description

Resolves #920 
Still need to refine it a bit and put loading skeleton

### Checklist

### DIFF
--- a/sites/example-project/src/pages/console/+page.md
+++ b/sites/example-project/src/pages/console/+page.md
@@ -3,3 +3,9 @@ select * from orders
 ```
 
 <Console data={needful_things_orders} />
+
+```needful_things_marketing_spend
+select * from marketing_spend
+```
+
+<Console data={needful_things_marketing_spend} />


csjh:
> This is so so so sweet!
> 
> Not sure if you want UX feedback yet, but a couple of (super minor) nits.
> 
> * Would be great if the text area got bigger when I write more than 4 lines of SQL
> * On mobile (yes I want to use it on mobile!) the browser zooms into the text area which is annoying
> * Could do some sql highlighting? (feel free to ignore if a pain)
> * Should it error out when you have invalid sql? (I'm not sure!)

Yeah thanks for the feedback, not too attached to any of the design so fire away

The first 3 points might be solvable by switching to a `contenteditable` div instead of the current `textarea`. I was checking out how Snowflake did their highlighting and that seems to be it. Though I tried to use one before and it seemed a bit buggy. I'll give it another go

The erroring has me conflicted right now because a. I do want some sort of feedback to the user if something's wrong, but b. there will always be at least a syntax error when they're typing. Maybe waiting 100(?)ms before putting up a toast notification or such would be best? That could still get annoying if you're a slow typer or just taking a pause though. Not sure what the ideal here is.

packages/console/components/DataTable.svelte
```diff
@@ -0,0 +1,487 @@
...
+	</div>
+{:else}
+	<ErrorChart {error} chartType="Data Table" />
+{/if}
+
+<style>
+	.table-container {
+		font-size: 9.5pt;
+		width: 97%;
+	}
```
ItsMeBrianD, line 187: Is this file copy-pasted from somewhere? We should normalize as much of this into tailwind as possible at some point. This is also a really bulky component, so if we can break it up that would be good; I didn't see any obvious places to do that though.

packages/console/components/duckdb.ts
```diff
@@ -0,0 +1,53 @@
+import { ConsoleLogger, AsyncDuckDB, AsyncDuckDBConnection } from '@duckdb/duckdb-wasm';
+import duckdb_wasm from '@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm?url';
+import duckdb_worker from '@duckdb/duckdb-wasm/dist/duckdb-browser-mvp.worker.js?worker';
+import { browser } from '$app/environment';
+
+let db: AsyncDuckDB;
```
ItsMeBrianD, line 6: How does this behave if there are multiple instances of `Console` on the page at a time? 
Is this a use-case we want to support? Are there performance implications with having multiple? 
The worker behavior is probably worth checking out in this case.

packages/console/components/Console.svelte
```diff
@@ -0,0 +1,72 @@
...
+	$: total_rows = query(`
+        SELECT COUNT(*) FROM (${sql_query.replace(/;$/, '')})
+    `);
+
+	$: paginated_results = query(`
+	    WITH query as (${sql_query.replace(/;$/, '')})
+        SELECT * FROM query
+	    LIMIT ${limit}
+	    OFFSET ${(pagination - 1) * limit};
+	`);
```
ItsMeBrianD, line 28: We should make sure that this is [debounced](https://www.npmjs.com/package/lodash.debounce)

packages/console/package.json
```diff
@@ -0,0 +1,46 @@
+{
+	"name": "@evidence-dev/console",
+	"version": "0.0.1",
+	"scripts": {
+		"dev": "evidence dev",
+		"build": "evidence build && npm run package",
```
ItsMeBrianD, line 6: Why use the evidence cli here rather than `parcel` like core-components?

packages/console/package-lock.json
ItsMeBrianD, line 1: We shouldn't have a package lock file

sites/example-project/tailwind.config.cjs
```diff
@@ -9,6 +9,7 @@ module.exports = {
 			'../../pages/**/*.{html,js,svelte,ts,md}', // This is used in end user projects to let them access tailwind classes
 			'../../src/**/*.{html,js,svelte,ts,md}', // This is used in end user projects to let them access tailwind classes
 			'./node_modules/@evidence-dev/core-components/dist/**/*.{html,js,svelte,ts,md}',
+			'./node_modules/@evidence-dev/console/dist/**/*.{html,js,svelte,ts,md}',
```
ItsMeBrianD, line 12: Do we want to consider something along these lines? This might make things a little less optimal for the sake of simplicity
```suggestion
			'./node_modules/@evidence-dev/**/dist/**/*.{html,js,svelte,ts,md}',
```

packages/console/components/DataTable.svelte
csjh: Yeah it's a slimmed down version of the `core-components` `DataTable`, so could probably share CSS between the two somehow

packages/console/components/duckdb.ts
csjh: It should work fine as long as none of the tables are duplicated names (which is actually probably something I should change; right now everything goes under `data`)

packages/console/package.json
csjh: The evidence CLI is there because it's still an evidence project, like how `npm create svelte` makes a sveltekit project. Though I'm not sure if `npm run package` is needed in the build script.

sites/example-project/tailwind.config.cjs
csjh: I think it would be better to make it computed based on all the plugins in the `evidence.plugins.yaml` so other components plugins can use tailwind as well

packages/console/components/DataTable.svelte
ItsMeBrianD: Is it similar enough to have a shared `Table` component or something that both are leveraging? This plugin can rely on core-components safely, so it could be exported from there and ingested here

csjh: There's some subtle differences that would make the script logic different but the markup might be able to be shared

packages/console/package.json
ItsMeBrianD: We should probably nail down if plugins should be evidence projects or not when we look at the data sources epic. I want to lean towards they shouldn't need to be, but I'd like to discuss more.

packages/console/components/DataTable.svelte
```diff
@@ -0,0 +1,487 @@
...
+	import getColumnSummary from '@evidence-dev/component-utilities/getColumnSummary';
+	import { convertColumnToDate } from '@evidence-dev/component-utilities/dateParsing';
+	import { formatValue } from '@evidence-dev/component-utilities/formatting';
+	import checkInputs from '@evidence-dev/component-utilities/checkInputs';
+	import { DownloadData, ErrorChart } from '@evidence-dev/core-components';
+
+	import MdFirstPage from 'svelte-icons/md/MdFirstPage.svelte';
+	import MdNavigateBefore from 'svelte-icons/md/MdNavigateBefore.svelte';
+	import MdNavigateNext from 'svelte-icons/md/MdNavigateNext.svelte';
+	import MdLastPage from 'svelte-icons/md/MdLastPage.svelte';
```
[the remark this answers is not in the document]
ItsMeBrianD: #967 addresses this

packages/console/components/duckdb.ts
csjh: Hmm looks like it (duckdb-wasm in general) is actually pretty unstable with multiple instances on the same page. It works about half the time and throws an error about the table not existing other times

mcrascal:
Closing in favour of the work on #973

PR #960, opened 2023-06-19
No separate data json per templated page

### LINKED ISSUE #900: Reduce the size of build artifacts by only creating one data object per templated page
## Steps To Reproduce

1. Create a project with a templated page, that has multiple routes.
2. `npm run build`
3. Inspect the contents of the `build` folder
4. Each route has an identical `__data.json` file

## Expected Behavior

There should be only one data file shipped ideally. Otherwise there is a lot of redundant data in the build artifact.

## Actual Behaviour

There are N `__data.json` files

image

I suspect this also slows down the build process _significantly_, as these build files are made. 

## Workarounds

Have lots of storage space whereever you host

archiewood:
### Description

Fixes #900 

### Checklist

### DIFF
--- a/sites/example-project/src/pages/+layout.js
+++ b/sites/example-project/src/pages/+layout.js
@@ -1,19 +1,10 @@
-import { browser } from '$app/environment';
-
-async function getDataOnServer(route) {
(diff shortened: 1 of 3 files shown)


archiewood:
Testing looks good

PR #962, opened 2023-06-20
Empty sources folder after copying to template

csjh:
### Description

Removes the contents of `sources` during template build

### Checklist

### DIFF
--- a/.changeset/seven-paws-switch.md
+++ b/.changeset/seven-paws-switch.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/evidence': patch
+---
+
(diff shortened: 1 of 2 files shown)


archiewood:
I could imagine wanting to include some files in sources to shop with the template. I assume this will not conflict as these will then be copied at runtime?

csjh:
> I could imagine wanting to include some files in sources to shop with the template. I assume this will not conflict as these will then be copied at runtime?

Yeah, the (sveltekit) template build script is only run when publishing

archiewood (approved):
LGTM

PR #964, opened 2023-06-20
Expose formatting to components

### LINKED ISSUE #709: Support number formatting in components
## Feature Description
Add a format property to all components to allow for setting number formats within the component rather than in SQL.

Previously, we removed the `fmt` option from the Value component because it was inconsistent with how other components treated formatting. Now that we have a number formatting system, we can apply the same format option across components.

### Initial Version
I think the initial version should only reference an existing number format - either an Evidence standard format or a custom format saved in the Settings menu:
```html
<LineChart
   x=date
   y=profit
   yFmt=usd0
/>

<LineChart
   x=date
   y=profit
   yFmt=mycustomfmt
/>
```

### Future Implementation
In the future, we could also let you write the Excel-style format code directly in the component:

```html
<LineChart
   x=date
   y=profit
   yFmt="$#,##0;($#,##0)"
/>
```

### Required Props
XY charts - needs to be applied at the column-level:
- `xFmt`
- `yFmt`
  - In cases where multiple y columns have been specified (e.g., `y={["colA", "colB"]}`), it should accept an array
- `sizeFmt` (BubbleChart)
- `series` should most often be strings, but there may be some cases where you want to format a number or append another string, in which case we can offer a `seriesFmt` prop

Individual chart components (Line, Bar, etc.)
- Will need some thought, but I believe we can add fmt props to wherever the columns can be defined
   - `Chart` can accept `xFmt`, `yFmt`, `sizeFmt`, and possibly `seriesFmt`
   - Each individual chart component can accept the same

Funnel/Sankey:
- `valueFmt`

Value:
- `fmt`

BigValue:
- `fmt` (applied to value prop column only)
- `comparisonFmt`

DataTable:
- Within the Column component, `fmt`

hughess:
### Description
This PR adds formatting control to components. It adds a new function to take a string and return an Evidence formatting object, as well as props for each component to accept formats from users.

Closes #709 

CleanShot 2023-06-20 at 18 32 39

### Setting a Format
The new format props accept a string. The format can be in any one of the categories below:
- Custom Format Tag
- Evidence Built-in Format Tag
- Excel-style format code

#### Charts:
- `xFmt`
- `yFmt`
- `sizeFmt` (for BubbleChart)
- `valueFmt` (for Sankey and Funnel Charts)

#### Tables
In the `Column` component:
- `fmt`

#### Value Component
- `fmt`

#### BigValue Component
- `fmt`
- `comparisonFmt`

#### Annotations
In ReferenceLine:
- ReferenceLine inherits the formatting of whichever column you're using
- Adding a `valueFmt` override was initially considered, but is not included in this PR

## Limitations
- Formatting of date axes on charts cannot be overridden at this time. You can still pass in a format, but it will only be used for tooltip or annotation formatting

### Checklist

### DIFF
--- a/packages/component-utilities/src/formatting.js
+++ b/packages/component-utilities/src/formatting.js
@@ -59,7 +59,7 @@ export function getFormatObjectFromString(formatString) {
 			formatTag: 'custom',
 			formatCode: potentialFormatTag,
 			valueType: 'number'
-		}
+		};
 		return newFormat;
 	}
 }
--- a/packages/component-utilities/src/formatting.js
+++ b/packages/component-utilities/src/formatting.js
@@ -45,6 +45,25 @@ export const lookupColumnFormat = (columnName, columnEvidenceType, columnUnitSum
 	return undefined;
 };
 
+export function getFormatObjectFromString(formatString) {
+	let potentialFormatTag = formatString;
+	let customFormats = getCustomFormats();
+	let matchingFormat = [...BUILT_IN_FORMATS, ...customFormats].find(
+		(format) => format.formatTag?.toLowerCase() === potentialFormatTag?.toLowerCase()
+	);
+	let newFormat = {};
+	if (matchingFormat) {
+		return matchingFormat;
+	} else {
+		newFormat = {
+			formatTag: 'custom',
+			formatCode: potentialFormatTag,
+			valueType: 'number'
+		}
+		return newFormat;
+	}
+}
+
 export const formatValue = (value, columnFormat = undefined, columnUnitSummary = undefined) => {
 	try {
 		return applyFormatting(value, columnFormat, columnUnitSummary, VALUE_FORMATTING_CONTEXT);
--- a/packages/component-utilities/src/formatting.js
+++ b/packages/component-utilities/src/formatting.js
@@ -4,6 +4,7 @@ import { CUSTOM_FORMATTING_SETTINGS_CONTEXT_KEY } from './globalContexts';
 import { findImplicitAutoFormat, autoFormat, fallbackFormat, isAutoFormat } from './autoFormatting';
 import { BUILT_IN_FORMATS } from './builtInFormats';
 import { standardizeDateString } from './dateParsing';
+import { inferValueType } from './inferColumnTypes';
 
 const AXIS_FORMATTING_CONTEXT = 'axis';
 const VALUE_FORMATTING_CONTEXT = 'value';
@@ -45,7 +46,7 @@ export const lookupColumnFormat = (columnName, columnEvidenceType, columnUnitSum
 	return undefined;
 };
 
-export function getFormatObjectFromString(formatString) {
+export function getFormatObjectFromString(formatString, valueType = undefined) {
 	let potentialFormatTag = formatString;
 	let customFormats = getCustomFormats();
 	let matchingFormat = [...BUILT_IN_FORMATS, ...customFormats].find(
@@ -57,9 +58,11 @@ export function getFormatObjectFromString(formatString) {
 	} else {
 		newFormat = {
 			formatTag: 'custom',
-			formatCode: potentialFormatTag,
-			valueType: 'number'
+			formatCode: potentialFormatTag
 		};
+		if (valueType) {
+			newFormat.valueType = valueType;
+		}
 		return newFormat;
 	}
 }
@@ -154,6 +157,9 @@ function applyFormatting(
 	if (value === undefined || value === null) {
 		return '-';
 	}
+
+	// let valueType = columnFormat.valueType ?? inferValueType(value);
+
 	let result = undefined;
 	if (columnFormat) {
 		try {
@@ -221,5 +227,7 @@ function maybeExtractFormatTag(columnName) {
 
 export function fmt(value, format) {
 	let formatObj = getFormatObjectFromString(format);
+	let valueType = inferValueType(value);
+	formatObj.valueType = valueType;
 	return formatValue(value, formatObj);
 }
--- a/packages/component-utilities/src/formatting.js
+++ b/packages/component-utilities/src/formatting.js
@@ -218,3 +218,8 @@ function maybeExtractFormatTag(columnName) {
 		return undefined;
 	}
 }
+
+export function fmt(value, format) {
+	let formatObj = getFormatObjectFromString(format);
+	return formatValue(value, formatObj);
+}
--- a/packages/component-utilities/src/formatting.js
+++ b/packages/component-utilities/src/formatting.js
@@ -158,8 +158,6 @@ function applyFormatting(
 		return '-';
 	}
 
-	// let valueType = columnFormat.valueType ?? inferValueType(value);
-
 	let result = undefined;
 	if (columnFormat) {
 		try {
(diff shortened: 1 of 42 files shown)


archiewood:
This is great. Only prop I thought might be missing is sizeFmt for the bubble chart tooltips?

1. Do we want to customize the fmt function that we also expose to users to use in js so that it uses the same underlying function that accepts both string and format code.

2. We should add this to the docs

archiewood:
Re 1: https://github.com/evidence-dev/evidence/blob/main/packages/preprocess/src/process-queries.cjs

hughess:
Ah yes, missed that one - thanks!

I didn't realize the format function was already in preprocess. I was initially going to do another PR after this one to expose that function, but that's perfect - I'll swap that out for the new function.

Will update the Number Formatting page in the docs as well.

hughess:
There's one other remaining issue here, which is date formatting - there's a few pieces to this:
- ECharts date axis formatting cannot be overridden yet - I think this one is fine for now
- The underlying formatting functions need to know if a value is a number or a date - I've set it to number as a starting place, but that causes problems for dates
   - Possible solutions here:
      - Check the data type of the column being passed into the function
      - Modify the functions to apply the formatting code regardless of the data type
- If we have the data type check for dates, another problem is created - the timezone shift we've run into in the past (e.g., you might have May 1 on your chart axis, but the tooltip shows Apr 30)
   - This one needs some investigation

I'm going to take a look into these, but depending on their complexity, we might be better off releasing component formatting only for numbers as a starting place.

archiewood:
Reading it again, I think we should try to simplify the number formatting page.

I feel like the first line should say something like:

> The easiest way to adjust formats is using component props. You can pass in excel style format codes, or one of a list of predefined format tags.


(Then an example using an excel format tag in a Value component)

And then sections with detail on:
- Formatting via column names  (if you always want a column to appear that way)
- Format function (if it's outside a component)

hughess:
Got it working!

CleanShot 2023-06-21 at 08 51 14

archiewood:
Oh thats so sick with HMR

hughess:
Update on the date issue:
- I've modified the `getFormatObjectFromString` function to accept an optional `valueType` parameter. Where possible, I've passed this in from the `columnSummary` in components to guarantee that we're using the correct value type
- I've added type inference to the `fmt` function since it won't have access to the `columnSummary`

The last unresolved piece which is likely a separate issue is this timezone shift I noticed - I believe it was when manually creating dates in a duckdb query, but will confirm.

hughess:
I've just added changes to the docs in this branch.

With this new way of using formatting, the terminology becomes a bit confusing - we were using "format tags" to describe everything, but now I think we need different terms. Here's what is in this version:
- Format tags appended in SQL --> **SQL Format Tags**
- Individual format tags (e.g., `usd`, `mydate`) --> **Format** / **Format Name**
- Format codes provided by Evidence (`usd`, `eur`, etc.) --> **Built-in Formats**
- Custom format codes (`mydate`, `mynum`, etc.) --> **Custom Formats**
- Manually specified format codes (`$#,##0`, etc.) --> **Excel-style Formats**
- Formatting directly in markdown (`{fmt(400, 'usd')}`) --> **Format Function**

With these changes, **tag** only applies to **SQL Format Tags** and everything else is referenced as a **format** (which can be built-in, custom, or Excel-style).

## To Summarize
There are 3 ways to use formats in Evidence post this PR:
1. Within components (using props)
2. Within SQL (using SQL Format Tags)
3. Directly in markdown (using the format function)

And Evidence supports these types of formats:
1. Excel-style formats
2. Built-in formats
3. Custom formats

archiewood:
In my mind, the "tag" is specifically the bit you append to the SQL column name. 

So my take would be slightly different, but similar:

| Example | Name |
|--|--|
|`usd0` | (built in) **format tag** |
|`mydate` | (custom) **format tag** |
|`$#,##0` | **format code** |
|`fmt(400, 'usd')` | **format function** |
|`xFmt` | **format prop** |

## Tags and codes

- **Format tags** each refer to a specific **format code**, which use excel syntax
- There are a number of built in **format tags**
- You can create custom **format tags**

Formatting methods
- Format props: you pass a **format tag** or **format code** to specify the format
- Format functions: you pass a **format tag** or **format code** to specify the format
- Column tags: you can append a **format tag** to a SQL column name to format it everywhere

WDYT? I think it means we have less concepts?

sites/docs/docs/core-concepts/formatting/index.md
```diff
@@ -1,78 +1,75 @@
...
 sidebar_position: 5
 hide_table_of_contents: false
-title: 'Number Formatting'
-description: 'Formats are defined using the column names in your SQL query'
+title: 'Value Formatting'
+description: 'Number and date formatting options in Evidence'
 ---
 
-[Format tags](#format-tags) are the primary way to format numbers and dates in Evidence. Format tags are appended to column names in your SQL query to format how they are displayed in components.
+The easiest way to format numbers and dates in Evidence is through component props. You can pass in any of the follwing:
```
archiewood, line 8: typo `follwing`

archiewood:
Also as side note, I think custom format tags are now much less important, as you can pass in an excel code directly to components: I might demote them in importance as a result.

hughess:
Agreed on the custom format importance. Those P1 items make sense to me.

After opening up component props, the word "tag" doesn't seem to fit as well (I used to think of the primary usage as "tagging" a column name with a format, so for me the tag included the underscore `_usd`). I think I'd find it hard to distinguish between "tag" and "code" as a new user (or at least it would require a bit of explanation).

I like your usage of **format code** to cover the excel case. 

In your layout of the terms, I would swap out the word "tag" for "name" everywhere except "column tags" - I think that would be easier for me to understand as a new user.

archiewood:
Your tag point is a good one.

I don't love "format name" though. I think name implies human readable, whereas this `usd0` is more of a shorthand.

Some other possible words

- format label
- format alias 
- format keyword

None of them are perfect

hughess:
That’s true. Format id is another option to throw into that list

packages/component-utilities/src/formatting.js
```diff
@@ -135,6 +157,9 @@ function applyFormatting(
 	if (value === undefined || value === null) {
 		return '-';
 	}
+
+	// let valueType = columnFormat.valueType ?? inferValueType(value);
+
```
ItsMeBrianD, line 162: Remove commented code

packages/core-components/src/lib/unsorted/ui/Formatting/CurrencyFormatGrid.svelte
ItsMeBrianD, line 1: As an aside, we should consolidate the format grids into a single component at some point, they look like they're all very similar

ItsMeBrianD (commented):
Looks good to me, we might want to add JSDoc to the new functions that have been added

hughess:
@ItsMeBrianD I've included JSDoc on the 2 new functions that got added in this PR - let me know what you think

hughess:
@ItsMeBrianD anything else we need to do before merging this one?

ItsMeBrianD:
This seems like a good place to start with JSDoc, this looks good to me

PR #968, opened 2023-06-22
Fix for grouped bar charts

hughess:
### Description
Due to a change introduced with ReferenceArea, grouped bar charts were being overridden and changed to stacked bar charts. This PR fixes that issue.

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/viz/ReferenceArea.svelte
+++ b/packages/core-components/src/lib/unsorted/viz/ReferenceArea.svelte
@@ -3,7 +3,7 @@
 </script>
 
 <script>
-	import { getContext } from 'svelte';
+	import { getContext, beforeUpdate } from 'svelte';
 	import { propKey, configKey } from './context';
 	import checkInputs from '@evidence-dev/component-utilities/checkInputs';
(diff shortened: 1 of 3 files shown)


packages/core-components/src/lib/unsorted/viz/ReferenceArea.svelte
```diff
@@ -194,6 +194,34 @@
 			return d;
 		});
 	}
+
+	$: chartOverrides = {
```
ItsMeBrianD, line 198: I may be missing some context on how some of the charting logic works; but why does this need to be a reactive statement? From usage in this file it looks like it could be const

ItsMeBrianD (commented):
Looks good, just 1 question

packages/core-components/src/lib/unsorted/viz/ReferenceArea.svelte
hughess: It may not need to be, but that's how it's set up elsewhere - I pulled this directly from `Bar.svelte`

ItsMeBrianD (approved):

Issue #966, opened 2023-06-21
CodeBlocks have too much padding


archiewood:
## Steps To Reproduce

````
```html
<div>
Hello
</div>
```
````

### Environment

- OS: Mac
- Browser: Firefox

## Expected Behavior

Less padding above and below the code inside the block

## Actual Behaviour

<img width="691" alt="image" src="https://github.com/evidence-dev/evidence/assets/58074498/000ff00c-32d9-421d-b5dd-9083850de48d">

tarunsamanta2k20:
I am interested to work on it. But could you please let me know this issue in detail?

archiewood:
<img width="737" alt="image" src="https://github.com/evidence-dev/evidence/assets/58074498/9de95045-a213-454c-845e-2e632bb7a187">

The vertical padding (yellow) is much wider than the horizontal padding (red).

We should reduce the vertical padding to be the same size as the horizontal padding

archiewood:
I'd take a look at this file: https://github.com/evidence-dev/evidence/blob/bed0865c188ec2a0f9db6cf98531d4213312e9fb/packages/core-components/src/lib/unsorted/ui/CodeBlock.svelte

lorinaX:
I would like to have a go at this!

archiewood:
Happy to receive a PR to fix this.

Yep it's the CodeBlock.svelte component

archiewood:
Resolved by #971

PR #970, opened 2023-06-22
Resolve extra margin on code element (class: 'my-2')

lorinaX:
### Description
Issue number: #966
> Resolves the extra margin issue on the "< code >" element with the class "my-2".
### Checklist
- [x] before https://files.slack.com/files-pri/T0239KTK4MP-F05DUG6PDFW/image.png
- [x] After https://files.slack.com/files-pri/T0239KTK4MP-F05DKDC7ZQE/image.png


lorinaX:
> Hey @lorinaX, thanks for submitting this!
> 
> This looks correct, but could you attach a screenshot of the updated code component? That helps us review UI changes more quickly. Thank you!

would this do the trick?
image

ItsMeBrianD:
Yep, that works!

I went ahead and ran the tests, if they pass we should be good to go

lorinaX:
> Yep, that works!
> 
> I went ahead and ran the tests, if they pass we should be good to go

Uh oh I think that when I ran pnpm install it modified the pnpm-lock file?

PR #982, opened 2023-06-26
[next] Partial support for multiple data sources

ItsMeBrianD:
### Description

- Adds a (very) basic CLI that makes debugging / testing the plugin connector easier in real projects
- Adds support for multiple data sources, using a `./sources/[source-name]` structure
- Adds a function to `@evidence-dev/postgres` which lets it be used as a plugin
- Includes functionality to exec all data source queries, loading them each into their own parquet file
  - [ ] This needs to be updated to not place the parquet files immediately next to the source query, but instead into the `static` directory (with the same directory structure as sources)
  - [ ] This also needs to generate a manifest of some type; this may be out of scope for this PR, and the format needs to be discussed with @csjh 

### Checklist


packages/universal-sql/src/build-parquet.js
```diff
@@ -0,0 +1,54 @@
...
+		// TODO: Make this work properly.
+
+		/** @type {import("apache-arrow").Vector} */
+		let values;
+		if (c.evidenceType === 'number') {
+			values = vectorFromArray(rawValues, new Float64());
+		} else if (c.evidenceType === 'string') {
+			values = vectorFromArray(rawValues, new Utf8());
+		} else if (c.evidenceType === 'date') {
+			values = vectorFromArray(rawValues, new Timestamp());
```
csjh, line 31: This should be `new DateMillisecond()`

csjh: Also might break if passed a stringified date type instead of a `Date` object, which I think some adapters do - might be worth shifting it to be the responsibility of the database adapter to return a `Date` object?

mcrascal:
@ItsMeBrianD I think a bit early for me to give feedback -- LGTM

mcrascal (approved):

packages/plugin-connector/src/data-sources/exec-source.js
```diff
@@ -0,0 +1,31 @@
+import { buildParquetFromResultSet } from '@evidence-dev/universal-sql';
+import fs from 'fs/promises';
+/**
+ *
+ * @param {DatasourceSpec} source
+ * @param {PluginDatabases} supportedDbs
```
Winterhart, line 6: I think the `return` might be missing in the `JSDoc` here

packages/plugin-connector/src/data-sources/exec-source.js
```diff
@@ -0,0 +1,31 @@
...
+				result: await runner(q.content, q.filepath)
+			};
+		})
+	);
+
+	for (const query of results) {
+		const { result } = query;
+		if (!result) continue;
+		const parquetBuffer = await buildParquetFromResultSet(result.columnTypes, result.rows);
+		await fs.writeFile(query.filepath.split('.')[0] + '.parquet', parquetBuffer);
```
Winterhart, line 29: Any chance we could get file named `myFile.prod.sql` ?

packages/plugin-connector/src/data-sources/get-datasource-plugins.js
```diff
@@ -0,0 +1,51 @@
...
+		 * ensuring that no duplicate databases are added.
+		 * @param {Promise<Record<string, { package: EvidencePluginPackage<EvidenceDatabasePackage>, factory: DatabaseConnectorFactory }>>} _acc - A promise representing the current state of the package map
+		 * @param {EvidencePluginPackage<EvidenceDatabasePackage>} v - The plugin package to be added to the map
+		 * @returns {Promise<Record<string, { package: EvidencePluginPackage<EvidenceDatabasePackage>, factory: DatabaseConnectorFactory }>>} - A promise representing the updated package map
+		 */
+		async (_acc, v) => {
+			// TODO: Handle Overrides
+
+			// Wait for the current state of the package map to resolve
+			const acc = await _acc;
```
Winterhart, line 25: What's an `acc`?

packages/plugin-connector/src/data-sources/get-datasource-plugins.js
```diff
@@ -0,0 +1,51 @@
...
+			// Build a DatabaseConnectorFactory for the plugin package's databases
+			const factory = await buildConnector(
+				path.join(v.path, v.package.main),
+				v.package.evidence?.databases ?? []
+			);
+			// For each database in the plugin package...
+			v.package.evidence.databases?.forEach((d) => {
+				// If a plugin with the same database already exists in the map, throw an error
+				if (d in acc) {
```
Winterhart, line 34: `d` is for `database` ?

packages/plugin-connector/src/data-sources/get-sources.js
```diff
@@ -0,0 +1,92 @@
...
+ * parses it, and returns a validated datasource specification.
+ *
+ * @param {string} sourceDir - The directory containing the connection.yaml file.
+ * @return {Promise<DatasourceSpecFile>} A Promise that resolves to a validated datasource specification.
+ */
+async function getConnectionParams(sourceDir) {
+	const connParamsRaw = await fs
+		.readFile(path.join(sourceDir, 'connection.yaml'))
+		.then((r) => r.toString());
+	const connParamsUnchecked = yaml.parse(connParamsRaw);
```
Winterhart, line 69: Can `yaml.parse` throw an error?

packages/plugin-connector/src/plugin-discovery/resolve-evidence-config.js
```diff
@@ -50,6 +50,28 @@ export const loadConfig = async (rootDir) => {
 	}
 };
 
+/**
+ * Wrapper function to create a package validator function
+ * @param {string} rootDir
+ * @returns
```
Winterhart, line 56: Should the `JSDoc` contain the returned `struct` `package, path`?

packages/universal-sql/src/build-parquet.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+export async function buildParquetFromResultSet(columns, data) {
+	const m = columns.map((c) => {
+		const rawValues = data.map((d) => d[c.name] ?? `Null`);
+
+		// TODO: Make this work properly.
+
+		/** @type {import("apache-arrow").Vector} */
+		let values;
+		if (c.evidenceType === 'number') {
```
Winterhart, line 26: This could be a `switch`

Winterhart (commented):
just few questions

packages/plugin-connector/src/data-sources/exec-source.js
ItsMeBrianD: We could add the `.prod`, but we wouldn't want to use `.sql` here I would imagine; these are binary columnar files

packages/plugin-connector/src/data-sources/get-datasource-plugins.js
ItsMeBrianD: accumulator

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#parameters

packages/plugin-connector/src/data-sources/get-datasource-plugins.js
ItsMeBrianD: Yes

packages/universal-sql/src/build-parquet.js
ItsMeBrianD: > might be worth shifting it to be the responsibility of the database adapter to return a Date object?

This is probably a good way to go; do you have docs for DateMillisecond? Apache Arrow docs had examples set up this way

csjh: No docs, but only the `DateDay` and `DateMillisecond` class extend the `Date_` class from what I can tell, and the type of `vectorFromArray` that accepts `Date` objects requires that

packages/plugin-connector/src/data-sources/exec-source.js
```diff
@@ -27,6 +29,10 @@ export const execSource = async (source, supportedDbs) => {
 		const { result } = query;
 		if (!result) continue;
 		const parquetBuffer = await buildParquetFromResultSet(result.columnTypes, result.rows);
-		await fs.writeFile(query.filepath.split('.')[0] + '.parquet', parquetBuffer);
+		const fileparts = query.filepath.split('/')
+		const outputFilename = fileparts.pop()?.split(".")[0]  + '.parquet'
+		const outputSubdir = fileparts.join("/").split("sources").slice(1).join("/")
```
csjh: This and the other direct usages of `/` might start problems on Windows

packages/plugin-connector/src/data-sources/exec-source.js
```diff
@@ -27,6 +29,10 @@ export const execSource = async (source, supportedDbs) => {
 		const { result } = query;
 		if (!result) continue;
 		const parquetBuffer = await buildParquetFromResultSet(result.columnTypes, result.rows);
-		await fs.writeFile(query.filepath.split('.')[0] + '.parquet', parquetBuffer);
+		const fileparts = query.filepath.split('/');
```
Winterhart, line 35: 👍🏼

Winterhart (approved):

.gitignore
```diff
@@ -4,6 +4,7 @@ dist
 .env
 .pnpm-debug.log
 .vscode
```
ud3sh, line 6: Using JetBrains Fleet?

packages/plugin-connector/src/data-sources/get-datasource-plugins.js
ud3sh: Pretty common in identifier in reduce / fold type functions.

.gitignore
ItsMeBrianD: Yeah have been giving it another spin; I try to look at it every couple months to see when it is in a more usable state

packages/component-utilities/src/formatting.js
```diff
@@ -199,3 +228,16 @@ function maybeExtractFormatTag(columnName) {
...
+/**
+ * Formats a value to whichever format is passed in
+ * @param {*} value the value to be formatted
+ * @param {string} format string containing an Excel-style format code, or a format name matching a built-in or custom format
+ * @returns a formatted value
+ */
+export function fmt(value, format) {
+	let formatObj = getFormatObjectFromString(format);
+	let valueType = inferValueType(value);
+	formatObj.valueType = valueType;
```
ud3sh, line 241: Looks like getFormatObjectFromString handles this already, i.e `getFormatObjectFromString(format, inferValueType(value));`

packages/component-utilities/src/formatting.js
```diff
@@ -45,6 +46,33 @@ export const lookupColumnFormat = (columnName, columnEvidenceType, columnUnitSum
...
+export function getFormatObjectFromString(formatString, valueType = undefined) {
+	let potentialFormatTag = formatString;
+	let customFormats = getCustomFormats();
+	let matchingFormat = [...BUILT_IN_FORMATS, ...customFormats].find(
+		(format) => format.formatTag?.toLowerCase() === potentialFormatTag?.toLowerCase()
+	);
+	let newFormat = {};
+	if (matchingFormat) {
+		return matchingFormat;
+	} else {
```
ud3sh, line 64: Is `custom` format tag (when the tag doesn't appear in built on or custom formats) meant to support `fmt` being called directly from markdown?  I am guessing if excel formatter can't handle it, it will just be caught and the original value returned.

sites/example-project/sources.bak/DEP_query.sql
ud3sh, line 1: sources.bak accidental commit?

ud3sh:
Looks, great @ItsMeBrianD 

When it's complete, it would be good to see to see an example with couple of sources (besides the duckdb example), along with the three different approaches to injecting .env-example connection values.  On cloud, we are using `env-example` files, which are not gitignored to give us an idea what an .env file would look like.

ud3sh (approved):

packages/plugin-connector/src/data-sources/get-sources.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Retrieves the contents of all query files in the source directory,
+ * excluding the 'connection.yaml' file, and returns them as an array of
+ * objects containing the filepath and content of each query file.
+ *
+ * @param {string} sourceDir - The path to the source directory.
+ * @param {Array<string>} contents - An array of filenames in the source directory.
+ * @return {Promise<DatasourceQuery[]>} - A promise that resolves to an array of objects
+ * containing the filepath and content of each query file.
+ */
```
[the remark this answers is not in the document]
ItsMeBrianD: Yes, it would be up to the duckdb plugin to filter that out. All files except the connection.yaml are passed to the plugin

We don't want to assume a specific extension for all queries, because non-sql adapters will probably have their own (e.g. .gql for graphql).

packages/component-utilities/src/formatting.js
ItsMeBrianD: I think this was introduced from a change on main, this wasn't part of this PR

Issue #979, opened 2023-06-26
Universal SQL: Inlined Queries


mcrascal:
Move inlined queries to leverage the #978, 

Considerations: 
1. HMR 
2. Performant data display (e.g. use the inlined db for pagination)

Issue #978, opened 2023-06-26
Universal SQL: Page level ddb instance


mcrascal:
Make parquet available in client side ddb: ok to have all of them, with no frontmatter would be good to be doing this lazily, but not required for #973

PR #985, opened 2023-06-27
[next] page level duckdb instance & inline queries

csjh:
### Description

#979, #978

- allows for variable interpolation into queries
- all queries are sourced from (soon to be) generated parquet files
- limited `getColumnSummary` analysis functions to first 1000 rows for perf reasons - might want to make a better fix
- sets up some infrastructure for #982 

One note, if anyone has suggestions on how prerendering queries could work, would very much like to hear them, because right now there's a FLER (flash of empty results)
Ideas:
- ignoring and documenting that you'll briefly have an empty array for queries on page (yucky)
- preprocessing in an `onDestroy` that captures the state of all variables in scope at the end of the script tag (very funky, possibly takes a while to run? also possibly not possible)
- similar to above, extracting script tag into "virtual" javascript file and getting values of all the scoped variables (is this even possible? would probably be faster than above though)
- require interpolated variables to have a static value on initialization (ie `let interpolated_var = 5;` rather than `let interpolated_var = func();`)
- ideal would be something like capturing state during SSR and feeding into the client-side load function

Also, to test locally just add the [`taxis.parquet` file](https://github.com/evidence-dev/duckdb-wasm-poc/tree/master/static/data) to the `static` folder

### Checklist


packages/component-utilities/src/getColumnSummary.js
```diff
@@ -4,55 +4,27 @@ import { lookupColumnFormat } from './formatting';
...
-	let columnSummary = [];
+	const columnSummary = {};
 
-	var colExtentsLegacy;
+	for (const colName of Object.keys(data[0])) {
+		const evidenceColumnType = getColumnEvidenceType(data, colName);
+		const type = evidenceColumnType.evidenceType;
+		const columnUnitSummary = getColumnUnitSummary(data.slice(0, 1000), colName);
+		const format = lookupColumnFormat(colName, evidenceColumnType, columnUnitSummary);
+		const extentsLegacy = getColumnExtentsLegacy(data.slice(0, 1000), colName);
```
ItsMeBrianD, line 14:
```suggestion
		const columnUnitSummary = getColumnUnitSummary(data, colName);
		const format = lookupColumnFormat(colName, evidenceColumnType, columnUnitSummary);
		const extentsLegacy = getColumnExtentsLegacy(data, colName);
```

PR #998, opened 2023-06-29
[next] Wrap connectors into database plugins

ItsMeBrianD:
### Description

Adds basic types and a wrapping function to allow existing connectors to function as plugins

### Checklist

### DIFF
--- a/packages/bigquery/index.cjs
+++ b/packages/bigquery/index.cjs
@@ -62,6 +62,10 @@ const standardizeResult = async (result) => {
 	return output;
 };
 
+
+/**
+ * @param {BigQueryOptions} database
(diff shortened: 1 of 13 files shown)


csjh (approved):
Should probably add a standard `columnTypes`, `rows`, and `queryRunner` type in a `.d.ts` somewhere (maybe `db-commons`) so there aren't `typedef`s in every file

ItsMeBrianD:
> Should probably add a standard columnTypes, rows, and queryRunner type in a .d.ts somewhere (maybe db-commons) so there aren't typedefs in every file

Agree; I'll open an issue, should probably go into db-commons

PR #1003, opened 2023-06-29
getColumnSummary cleanup

csjh:
### Description

only runs the expensive bits if it's actually needed now (which is often, since most queries have a number column)

### Checklist


ItsMeBrianD (approved):

ItsMeBrianD:
Approving for now; for phase 2 we should sub this for the SQL statements

PR #1008, opened 2023-07-06
Swap out favicon and apple icons for smaller size

hughess:
### Description
The favicon and apple touch icons that are shipped with each Evidence project are high res and quite large. This replaces them with a much smaller file.

Hoping these won't appear blurry, but doesn't seem so after testing on my devices.

### Checklist


ItsMeBrianD (approved):

hughess:
@ItsMeBrianD is this the right place to swap these out? The changeset shows only the `components` package changing. I'm having trouble following the thread on how the `.evidence/template` picks up the images

hughess:
Ok got it, thanks. Are the images in `example-project` the right ones to swap out, or the ones in the `evidence` package under `template`?

ItsMeBrianD:
The ones in example-project are the ones you'll want to swap, they are copied into the template when the evidence cli is built

PR #1010, opened 2023-07-06
[next] Better Schema Explorer; Add Query Console

ItsMeBrianD:
### Description

`/explore/schema`
image

`/explore/console`
image

### Checklist

### DIFF
--- a/.changeset/thirty-mirrors-yawn.md
+++ b/.changeset/thirty-mirrors-yawn.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/evidence': patch
+---
+
+Add a query console page; improve usability of schema explorer
--- a/sites/example-project/src/pages/explore/console/+page.svelte
+++ b/sites/example-project/src/pages/explore/console/+page.svelte
@@ -1,0 +1,40 @@
+<script>
+    import {DataTable} from "@evidence-dev/core-components"
+
+	export let data;
+	let { __db: db } = data;
+
+    let sql_query, results = [], loading
+    async function run() {
+        loading = true
+        results = await db.query(sql_query)
+        loading = false
+    }
+
+    let input
+
+</script>
+
+<div class="container mx-auto relative font-mono bg-black rounded-xl" 
+    on:click={() => input.focus()}
+    on:keypress={e => e.key === 'Enter' && (e.getModifierState('Control')) && run()}
+    >
+    <h3 class="mx-4 py-4 m-0 text-white select-none font-mono">Query Console</h3>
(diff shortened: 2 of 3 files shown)


csjh (approved):

archiewood:
This is really exciting!

Some thoughts based on testing https://vercel-universal-sql.vercel.app/

All just suggestions, and I'm sure some easier to implement than others


## Routes
Not sure about the route naming. I think we want these routes to be really memorable. I used looker's SQL console and I still remember that the route was `myinstance.looker.com/sqlrunner`

### Ideas
You need the database explorer when you are writing the SQL, (what are my columns called?) so they should be co-located on the same page.

- Maybe you could have a single route: eg: `/explore`, '/query`, `/sql` or `/console`
- Repurpose the sidebar for the explorer 
  - Have a way to navigate  `< Back` to the rest of the nav links at the top of the sidebar so you aren't "trapped" in the explorer
- Console is in the main screen


## Console
Love the styling. It's really clean.
- **Keybinding:** Can we hook up `Ctrl`+`Enter` / `Cmd`+`Enter` (or similar) to run the query - and perhaps put in a UI hint about that functionality on wide viewports
- **Save as source**: Some really easy workflow to turn a scratch query into a query stored in your project. I'm imagining: 
   - (Button) Save as Source
   - What do you want to name this query? x.sql
   - Optional checkbox to save as raw CSV rather than .sql
   - [copies SQL (or raw data) into `/sources`]
- **Failure:** For some reason, after 2-3 queries, it seems to fail and I have to refresh the page. The button is greyed out after the failure so I can't re-run it.
- **Store State:** When I refresh the query is lost - should we store the state somehow? 
     - Custom URL? (This makes it super shareable in organizations, which I think could add to internal product virality)
     - raw:`/explore/console?query=select%20%0A%27Canada%27%20as%20country%2C%20%0A%20200%20as%20sales`
     - or for cloud users we could store something under a unique ID, which might be cleaner: `/explore/console?query_id=00001`

 
## Schema
- **Hierarchy:** Could there ever be multiple schemas or databases? https://duckdb.org/docs/sql/information_schema
- **Source:** Feels like it could eventually be helpful to display the source of the data? Did it come from Snowflake, a csv, an API call?
- **UI**: For me my mental model of a database is a tree: database->schema->table->column. I think I'd want to be able to see the columns for multiple tables at once, which doesn't really work if you have to "tab" through them.
- **Mobile**: This side by side view doesn't really have enough space.


## UI Inspo

| Tool | UI |
|--|--|
| [SQLTools](https://vscode-sqltools.mteixeira.dev/en/home) | image |
| Snowflake | image |

PR #1014, opened 2023-07-07
Sources are schemas

### LINKED ISSUE #991: Universal SQL: Preventing name conflicts between sources files with same name in different sources
e.g. current behavior is:
```
sources/
  a/
    test.sql -> becomes test
  b/
    test.sql -> becomes test
```

We need to decide on a method of differentiating these queries.

ItsMeBrianD:
### Description

Resolves #991

### Checklist


packages/universal-sql/src/client-duckdb/browser.js
```diff
@@ -42,24 +42,42 @@ export async function initDB() {
 	await db.open({ query: { castBigIntToDouble: true, castTimestampToDate: true } });
 }
 
+/**
+ * Updates the duckdb search path to include only the list of included schemas
+ * @param {string[]} schemas
+ * @returns {Promise<void>}
+ */
+export async function updateSearchPath(schemas) {
```
csjh, line 50: This and `node.js`' definitions could be merged into something in `both.js` with an extra `db` param and re-exported here (also applies to `setParquetURLs`

packages/universal-sql/src/client-duckdb/browser.js
```diff
@@ -42,24 +42,42 @@ export async function initDB() {
...
+	await connection.query(`PRAGMA search_path='${schemas.join(',')}'`);
+
+	await connection.close();
+}
+
 /**
  * Adds a new view to the database, pointing to the provided parquet URL.
  *
- * @param {string[]} urls
+ * @param {Record<string, string[]>} urls
```
csjh, line 63: Should update the `.d.ts` files

packages/universal-sql/src/client-duckdb/browser.js
ItsMeBrianD: It looks like there are some interop conflicts happening here, going to leave as-is for now

Issue #1018, opened 2023-07-10
BigValue Format Tag Error on data update


mcrascal:
The most basic way to reproduce:

```
<script>
  import { BigValue } from '@evidence-dev/core-components';

  let testData = [{ test: 1 }];
  setTimeout(() => {
    testData = [{ test: 2 }];
  }, 1000);
</script>

<BigValue data={testData} value="test" />

```

on the second render (after the 1000ms timeout), I get an exception:

`Uncaught (in promise) potentialFormatTag.toLowerCase is not a function`

archiewood:
I've just repro'd this error again:

https://github.com/evidence-dev/northstar-report


1. Run the project
1. Change some sql

Error:


![CleanShot 2023-07-19 at 17 15 37](https://github.com/evidence-dev/evidence/assets/58074498/f302692c-6609-446c-8e9a-aea033460fe9)

csjh:
@archiewood I don't think the fix has released yet, still in #1045

archiewood:
Ah, that tracks

PR #1020, opened 2023-07-10
Fix BigValue format error

csjh:
### Description

Only runs `getFormatObjectFromString` if `fmt` is a string

### Checklist


hughess:
@csjh what was the error? Wondering if this will need to be changed in other places format props are accepted as well (e.g., `xFmt`, `yFmt` in charts)

csjh:
There's a reproduction in #1018, the issue was just that all the reactive portions of `Value.svelte` are in the one block, so whenever `data` updates the entire block reruns, which causes `fmt` (which is a prop that's originally a string but is reassigned in the block) to update again with `fmt = getFormatObjectFromString(fmt, columnSummary[0].format.valueType)`, which messes up the assumption of `fmt` being a string in `getFormatObjectFromString`. Hopefully when we shift everything to JSDoc we'll catch things like this much easier

csjh:
It might need to be changed if we reassign `fmt` anywhere else in that same kind of pattern without type checks

hughess:
Ahh ok. I believe these are the only other places that happens:
- Chart
- SankeyChart
- USMap
- FunnelChart

Is an alternative to use a separate variable rather than overwriting the initial variable?

csjh:
Yeah, separate variable is possible, I just wasn't sure if it was a specific choice meant so there was an option to in format objects or not. If that isn't intended then switching to a different variable is best

hughess:
The intention in this case was only to check if they've supplied a `fmt` argument, but the assumption was that it would always be a string

hughess (approved):

PR #1021, opened 2023-07-11
Fix Svelte Peer Dependencies

### LINKED ISSUE #1017: Conflicting dep between icon pack and svelte version causes issues upgrading
- @evidence-dev/core-components requires @steeze-ui/svelte-icon ^1.4.1
- @steeze-ui/svelte-icon 1.5.0 requires svelte ^4.0.0
- @evidence-dev/component-utilities requires svelte 3.55.0

ItsMeBrianD:
### Description

This pins all versions of svelte to `^3.59.0`

Resolves #1017 

### Checklist


packages/core-components/package.json
mcrascal, line 1: @ItsMeBrianD shouldn't we be pinning the icon pack as well?

ItsMeBrianD:
Looks like there were a few svelte occurrences that got missed; I've also removed svelte-icons because it is no longer being used.

mcrascal (approved):
LGTM -- I would pin the icon pack version to a specific version number, but if you feel confident that they won't bump the svelte dep against the 1.4.* versions, it seems fine to me.

PR #1032, opened 2023-07-12
Fix most tests

csjh:
### Description

Clean up the buggy tests minus buggy linting


ItsMeBrianD:
Can we lint/format? Should be good to go after that

PR #1044, opened 2023-07-17
Details component

hughess:
### Description
As described originally by Hugh in slack, using the built-in `<details>` tag in markdown to include metric definitions on a page.

This component wraps the `details` tag to make the syntax a bit simpler and more consistent with our other components.

A couple of outstanding items for this:
- Naming: it may make more sense for the name to be general like "Details" (edit: have gone with `Details`)
- Styling: may need some spacing and sizing adjustments
- Icon: Triangle probably good for now, but we can consider an icon that would fit better with our other UI elements

CleanShot 2023-07-17 at 12 49 54

### Checklist

### DIFF
--- a/.changeset/calm-grapes-thank.md
+++ b/.changeset/calm-grapes-thank.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/core-components': patch
+'@evidence-dev/components': patch
+---
+
+Add Definitions component
--- a/.changeset/little-baboons-hear.md
+++ b/.changeset/little-baboons-hear.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/preprocess': patch
+---
+
+stop exporting the metadata object from mdsvex frontmatter
--- a/.changeset/mean-pumas-run.md
+++ b/.changeset/mean-pumas-run.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+limit `getFormatObjectFromString` in Value.svelte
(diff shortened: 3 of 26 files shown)


ItsMeBrianD (approved):
Love it!

hughess:
After reflecting on the name, it seems like Definitions is probably too specific.

A couple of alternatives:
- Details (same as html tag)
- Info

archiewood:
> it seems like Definitions is probably too specific.

agreed

I feel like whatever the name we settle on should indicate it's functionality:
- It stores information that is hidden by default, but can be opened if required.
- It's not the same as the HTML details component, and the syntax you use should differ: so web devs don't get confused

I think there is also an argument for not building this component, as it is so similar to an HTML tag.

In that case, one approach we could take would be to add a section on: "Useful HTML components" to the docs, adding a small example, and then pointing users to the MDN docs.

hughess:
In most cases I'd agree with going with the html tag, but in this case I think the syntax warrants a simpler component.

The one downside I can see to doing it this way is the section won't be expandable in a github markdown preview like the `<details>` tag is. But I have yet to see those previews being used extensively for Evidence projects because the rest of our components don't render there, so my hunch is that isn't a big issue.

We can always change or remove this if these assumptions are wrong.

archiewood:
If we were optimizing for GitHub / other markdown engines I think there's a bunch of stuff we'd want to change.

I think there's a future path to this, which is generating md files as part of the build outputs (as well as html).
- charts are generated and saved as images
- no reactivity
- templated pages are all created

But anyway this is out of scope

PR #1049, opened 2023-07-18
update pnpm

csjh:
### Description

Netlify seems to want >=8 now


ItsMeBrianD:
LGTM, once checks are passing.

Might need to update pnpm in the github actions

csjh:
Yeah, I thought setting the `packageManager` field would make the action use the right version but it still isn't

ItsMeBrianD:
https://github.com/evidence-dev/evidence/blob/main/.github/workflows/release.yml#L20

This is likely what you're looking for

PR #1052, opened 2023-07-20
Add `profile` function to measure load / query times

ItsMeBrianD:
### Checklist


csjh (approved):
LGTM after a format

PR #1057, opened 2023-07-21
[next] improve zod schemas

ItsMeBrianD:
### Description

- Should validate only the first row of a returned dataset (reduce memory usage, improve performance)
- More stringent row checking (columnTypes must match the returned object keys now)

### Checklist

### DIFF
--- a/packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
+++ b/packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
@@ -13,12 +13,7 @@ export const QueryResultSchema = z
 				// result has no rows, we can't validate this
 				// but this is a correct result set
 				if (data.length === 0) return true;
-				try {
-					z.record(z.string(), z.any()).parse(data[0]);
-				} catch {
-					return false;
-				}
-				return true;
+				return z.record(z.string(), z.any()).safeParse(data[0]).success;
 			},
 			{ message: 'Data connector returned invalid rows' }
 		),
--- a/packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
+++ b/packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
@@ -1,15 +1,88 @@
 import { z } from 'zod';
 
-export const QueryResultSchema = z.object({
-	rows: z.array(z.record(z.string(), z.any())),
-	columnTypes: z.array(
-		z.object({
-			name: z.string(),
-			evidenceType: z.enum(['boolean', 'number', 'string', 'date']),
-			typeFidelity: z.union([z.literal('precise'), z.literal('inferred')])
-		})
-	)
-});
+export const QueryResultSchema = z
+	.object({
+		// Note that this only validates that the first item in the array
+		// is a record with string keys. If the connector returns some
+		// inconsistent array (e.g. [{}, 1]), it will not detect the
+		// invalid row.
+		rows: z.array(z.any()).refine(
+			(data) => {
+				// result is not an array, fail
+				if (!Array.isArray(data)) return false;
+				// result has no rows, we can't validate this
+				// but this is a correct result set
+				if (data.length === 0) return true;
+				try {
+					z.record(z.string(), z.any()).parse(data[0]);
+				} catch {
+					return false;
+				}
+				return true;
+			},
+			{ message: 'Data connector returned invalid rows' }
+		),
+		columnTypes: z.array(
+			z.object({
+				name: z.string(),
+				evidenceType: z.enum(['boolean', 'number', 'string', 'date']),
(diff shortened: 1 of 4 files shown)


packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
```diff
@@ -1,15 +1,90 @@
...
+export const QueryResultSchema = z
+	.object({
+		// Note that this only validates that the first item in the array
+		// is a record with string keys. If the connector returns some
+		// inconsistent array (e.g. [{}, 1]), it will not detect the
+		// invalid row.
+		rows: z.array(z.any()).refine(
+			(data) => {
+				// result is not an array, fail
+				if (!Array.isArray(data)) return false;
```
Winterhart, line 12: Another way would be to have the `z.any()...` as another `zod` object..

e.g.

```js

rows: z.array(rowItem)

//...

export const rowItemModal = z.record(z.string), z.any())
export type rowItem = z.infer<typeof rowItemModal>;
```

csjh (approved):
LGTM, though I think we should get rid of `columnTypes` very soon; I don't think inferred Dates would survive the parquet transformation, and after nothing's inferred we can just use `typeof` to identify types

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
```diff
@@ -1,15 +1,90 @@
...
+				// result has no rows, we can't validate this
+				// but this is a correct result set
+				if (data.length === 0) return true;
+				try {
+					z.record(z.string(), z.any()).parse(data[0]);
```
Winterhart, line 17: There's also a `safeParse` equivalent of `parse` but doesn't throw `Error`

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
```diff
@@ -1,15 +1,90 @@
...
+				}
+				return true;
+			},
+			{ message: 'Data connector returned invalid rows' }
+		),
+		columnTypes: z.array(
+			z.object({
+				name: z.string(),
+				evidenceType: z.enum(['boolean', 'number', 'string', 'date']),
+				typeFidelity: z.union([z.literal('precise'), z.literal('inferred')])
```
Winterhart, line 29: There's also the method `.nativeEnum`

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
```diff
@@ -1,15 +1,90 @@
...
+	})
+	.refine(
+		(data) => {
+			// Validate that all columnTypes appear
+			if (data.rows.length) {
+				// Filter to column types where name is not in row
+				// Then map columnTypes to their names to make things easier
+				// If there are any columns that were not filtered out; provide an error to zod
+				const missingColumns = data.columnTypes
+					.filter((ct) => !(ct.name in data.rows[0]))
```
Winterhart, line 41: What's a `ct`?

Winterhart (commented):
Looks great! only few questions

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
ItsMeBrianD: I tried that initially and that didn't prevent it from validating every row, doing `z.any()` and then checking the first row in the refine method was the only thing that prevented the full scan

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
ItsMeBrianD: We can make that change if/when we clean up the types in `db-commons`, for now I don't think that value isn't really used so we just want to make sure that it matches the expected interface until we change it

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
ItsMeBrianD: `ct` is a columnType

packages/plugin-connector/src/data-sources/schemas/query-runner.schema.js
```diff
@@ -1,15 +1,90 @@
...
+				if (!Array.isArray(data)) return false;
+				// result has no rows, we can't validate this
+				// but this is a correct result set
+				if (data.length === 0) return true;
+				try {
+					z.record(z.string(), z.any()).parse(data[0]);
+				} catch {
+					return false;
+				}
+				return true;
```
ItsMeBrianD, line 21:
```suggestion
				return z.record(z.string(), z.any()).safeParse(data[0]).success;
```

PR #1064, opened 2023-07-24
support massive files + non-directories in sources

csjh:
### Description

- returns `null` + filepath for massive files so adapters have the option of reading them manually
- checks if a source dir is a directory before attempting to read as such

### Checklist


ItsMeBrianD (approved):

ItsMeBrianD:
Linting & Tests, then should be good

Issue #1066, opened 2023-07-24
MotherDuck Support for DuckDB Connector


archiewood:
## Feature Description

The ability to connect to MotherDuck as a data source

Currently this is prevented by the `../../` that is appended to all filenames

Some implementation details here: 
- [Browser Auth](https://motherduck.com/docs/getting-started/connect-query-from-python/installation-authentication#using-human-user-authentification): for dev
- [Service token](https://motherduck.com/docs/authenticating-to-motherduck#authentication-using-a-service-token): for dev
- [Environment Variable](https://motherduck.com/docs/authenticating-to-motherduck#saving-the-service-token-as-an-environment-variable): suspect this will be the route for deployment

And some more details on using Node.js in [Motherduck's Slack](https://motherduckcommunity.slack.com/archives/C058PJL4R0S/p1690225690421389?thread_ts=1690221286.190379&cid=C058PJL4R0S)

archiewood:
@edgarrmondragon FYI

archiewood:
Will go out with next release

edgarrmondragon:
Thanks for quick turnaround on this @archiewood!

PR #1067, opened 2023-07-24
Support Motherduck

archiewood:
### Description

Addresses https://github.com/evidence-dev/evidence/issues/1066

After:

image

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/ui/Databases/DuckdbForm.svelte
+++ b/packages/core-components/src/lib/unsorted/ui/Databases/DuckdbForm.svelte
@@ -26,7 +26,7 @@
 			id: 'filename',
 			label: 'Filename',
 			type: 'filename',
-			additionalInstructions: 'Name of file stored in the same directory as your Evidence project',
+			additionalInstructions: 'Name of file stored in the same directory as your Evidence project. For MotherDuck, use "md:?motherduck_token=[TOKEN], and select No extension.',
 			optional: false,
 			override: false,
 			placeholder: 'mydatabase',
@@ -113,8 +113,9 @@
 				bind:value={ext}
 				on:change={handleKey}
 			>
-				<option value=".db">.db</option>
 				<option value=".duckdb">.duckdb</option>
+				<option value=".db">.db</option>
+				<option value="">No extension</option>
 			</select>
 
 			<input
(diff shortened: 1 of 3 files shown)


packages/duckdb/index.cjs
```diff
@@ -42,7 +42,12 @@ const mapResultsToEvidenceColumnTypes = function (rows) {
 
 const runQuery = async (queryString, database) => {
 	const filename = database ? database.filename : getEnv(envMap, 'filename');
-	const filepath = filename !== ':memory:' ? '../../' + filename : filename;
+	var filepath;
+	if (filename.includes('.db') || filename.includes('.duckdb')) {
```
archiewood, line 46: Not sure if this is exactly the right logic to be most future proof here.

This will catch all possible duckdb files that people add via the UI. 

I guess it's possible someone could have a .db in their token name, or that it could be a duckdb file stored in an S3 bucket or similar

packages/core-components/src/lib/unsorted/ui/Databases/DuckdbForm.svelte
```diff
@@ -113,8 +113,9 @@
 				bind:value={ext}
 				on:change={handleKey}
 			>
-				<option value=".db">.db</option>
 				<option value=".duckdb">.duckdb</option>
+				<option value=".db">.db</option>
+				<option value="">No extension</option>
 			</select>
```
archiewood, line 119: This UI is not great.

Firstly it doesn't tell you what to do (it's telling you it needs a file, but in the case of motherduck it's a string)

Also Evidence really only works properly if you pass a service token (otherwise it tries to reauth for every query, which is unusable), and this would be much clearer if there was a separate box for it

edgarrmondragon: I wonder if the `MOTHERDUCK_TOKEN` works for the JS duckdb client library

archiewood: It does :)

image

edgarrmondragon: Oh sorry I meant the env var that's referenced in the docs: https://motherduck.com/docs/authenticating-to-motherduck#saving-the-service-token-as-an-environment-variable

If that works then the user could simply type `md:` in that field and set the env var elsewhere.

archiewood: Hmm good q, not sure

packages/duckdb/index.cjs
archiewood: this should not be breaking at least

PR #1082, opened 2023-08-02
Dropdown component / infrastructure for ddb in components

csjh:
### Description

adds a declarative SQL dropdown component, and builds up support for duckdb queries in components (prerendering, types, etc)

### Checklist


packages/core-components/src/lib/unsorted/viz/Dropdown.svelte
```diff
@@ -0,0 +1,53 @@
...
+
+	const search = browser
+		? debounce((query) => profile(db.query, query).then((value) => (data = value)), 200)
+		: (query) => (data = profile(db.query, query, component_identifier));
+
+	$: db = $page.data.__db;
+	/** @type {{ tag: unknown, value: unknown }[]} */
+	$: data = $page.data.data?.[component_identifier] ?? [];
+	// todo: check if `from (select * from (${from})) would be better, allow for subqueries
+	$: search(`select (${tag}) as tag, (${value}) as value from ${from} where (${where})`);
```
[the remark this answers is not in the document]
ItsMeBrianD: https://uwdata.github.io/mosaic/api/sql/queries.html#fro
`from` can be a subquery it looks like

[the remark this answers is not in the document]
csjh: image

doesn't work, I suspect that means it exclusively accepts a subquery built with the query builder

mcrascal:
@csjh @ItsMeBrianD If I'm understanding the discussion: I like the component issuing the queries itself, rather than requiring a separate code fence to supply the values. We've tested that with folks a few times, and people find it fairly intuitive. It's also a lot more terse than needing to write the query as a standalone just to reference it in the component.

packages/core-components/src/lib/unsorted/viz/Dropdown.svelte
```diff
@@ -0,0 +1,60 @@
...
+
+	/** @type {string} */
+	export let name;
+	$: component_identifier = `_Dropdown-${name}`;
+
+	const search = browser
+		? (query) => profile(db.query, query).then((value) => (data = value))
+		: (query) => (data = profile(db.query, query, component_identifier));
+
+	$: db = $page.data.__db;
```
ItsMeBrianD, line 38: Can we pass functions as part of the page store? 
If so, we should just make the search function part of the page store, that will help keep this from spreading to all components that need it

packages/core-components/src/lib/unsorted/viz/index.js
```diff
@@ -8,6 +8,7 @@ export { default as Bubble } from './Bubble.svelte';
 export { default as Chart } from './Chart.svelte';
 export { default as Column } from './Column.svelte';
 export { default as DataTable } from './DataTable.svelte';
+export { default as Dropdown } from './Dropdown.svelte';
```
ItsMeBrianD, line 11: This component is an `atom`, we should not add new components to `unsorted` anymore

ItsMeBrianD (changes requested):
A few nitpicks wrt the dropdown itself; the pre-rendering changes here have a smell imo:

- Why do we need to write into the `.svelte-kit` folder? This seems like an antipattern

- Why do we need to keep track of all the queries that run on each page? Can we modify the dropdown component in some way to work within the bounds of sveltekit's existing pre-rendering?

packages/core-components/src/lib/unsorted/viz/Dropdown.svelte
csjh: How should the generalized search function look? I think the only possible way to generalize it would be something like `$page.data.search(query, component_identifier).then((value) => data = value)` with a weird synchronous `then` function returned during SSR which imo is less clear

csjh:
Keeping track of all the pages queries is necessary so we know which prerendered queries to fetch in the load function

I couldn't think of a way to make it work within SK's prerendering because the expectation from SK is that everything you want to prerender is known within the load function (a reasonable expectation), which isn't true for us -- the client can have any arbitrary amount of `Dropdown`s or any other usql component within the page

To "fix" that it's writing directly into `.svelte-kit` where it would have written the prerendered responses

packages/core-components/src/lib/unsorted/viz/Dropdown.svelte
csjh: Generalized with a third parameter callback so it would look something like `db.query(query, (value) => (data = value))` (for both SSR and CSR) in a new component

packages/core-components/src/lib/atoms/Dropdown.svelte
```diff
@@ -0,0 +1,59 @@
...
+	$: search(`select (${label}) as label, (${value}) as value from ${from} where (${where})`);
+	$: selected = data[0]?.value;
+	$: $inputs[name] = selected;
+
+	/*
+	initial query when db is still loading awaits `database_initialization`, so dummy query works 
+	fine as a substitute for weird export stuff
+	*/
+	let loadingDuckDB = browser;
+	$: browser && db.query('SELECT 1').then(() => (loadingDuckDB = false));
```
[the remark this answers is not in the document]
csjh: Yeah, seems reasonable to have a `load` function on `__db` that just returns `database_initialization`

packages/core-components/src/app.d.ts
```diff
@@ -4,7 +4,36 @@ declare global {
...
+					query: string,
+					options?: {
+						query_name?: string;
+						callback?: (results: Record<string, unknown>[]) => unknown;
+					}
+				): void;
+				// Waits for database to finish loading
+				// This includes fetching `manifest.json`, intitializing the WASM and WebWorker
+				// setting up parquet connections, and updating the schema search path
+				load(): Promise<void>;
```
ItsMeBrianD, line 19:
```suggestion
				/**
				 * Waits for database to finish loading
				 * This includes fetching `manifest.json`, intitializing the WASM and WebWorker
				 * setting up parquet connections, and updating the schema search path
				 */
				load(): Promise<void>;
```

packages/core-components/src/lib/atoms/Dropdown.svelte
```diff
@@ -32,7 +32,10 @@
 	$: component_identifier = `_Dropdown-${name}`;
 
 	const search = (query) =>
-		profile(db.query, query, component_identifier, (value) => (data = value));
+		profile(db.query, query, {
+			query_name: component_identifier,
+			callback: (value) => (data = value)
+		});
```
ItsMeBrianD, line 38: Does passing an object here work? I've always written this as
```suggestion
		profile(db.query, query, component_identifier, (value) => (data = value) );
```

packages/core-components/src/lib/atoms/Dropdown.svelte
```diff
@@ -46,7 +49,7 @@
 	fine as a substitute for weird export stuff
 	*/
 	let loadingDuckDB = browser;
-	$: browser && db.query('SELECT 1').then(() => (loadingDuckDB = false));
+	$: browser && db.load().then(() => (loadingDuckDB = false));
```
ItsMeBrianD, line 52: 🚀

packages/preprocess/src/process-queries.cjs
```diff
@@ -51,7 +51,11 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 				guaranteed to have run
 			*/''}
 				$: if (!browser || dev) {
-					profile(__db.query, _query_string_${id}, "${id}", (value) => ${id} = value);
+					profile(
+						__db.query,
+						_query_string_${id},
+						{ query_name: "${id}", callback: (value) => ${id} = value }
+					);
```
ItsMeBrianD, line 58: Ditto above

packages/core-components/src/lib/atoms/Dropdown.svelte
csjh: Yeah, I just switched it to accept an object instead of an arbitrary amount of parameters so it's easier to add onto in the future

ItsMeBrianD: An array would be a better choice than an object in this case, ordering is important; I think this would be an okay one to refactor in the future if we need to though

packages/core-components/src/lib/unsorted/viz/Dropdown.svelte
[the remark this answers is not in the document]
ItsMeBrianD: https://github.com/evidence-dev/query-store/blob/2227a545c1b4a48dbd3527c4cef48f6d74c4dc7f/src/lib/query.store.ts#L134-L138

mcrascal:
@csjh -- coming back to this, I think the syntax is just a little off and possibly too permissive. 

Let me know what you think of this: 

* `label` -> should be the label on the actual dropdown (e.g. the form label) 
* `value_label` -> optional label to be shown in the dropdown
* `value` -> column to be used for the actual returned value when referenced. If there is no value label, this is displayed.  
* `value` and `value_label` should be fragments of a select which is run against the from + where arguments. I think having the ability to inline the additional where clause in value label is pretty tricky.
* `order` -> optional column into the order by of the generated query and resulting option list 
* Deduplication: depending on the source table, we could have duplicate entries here. I think we should dedupe across the pair of value and value_label. This still leaves the possibility of duplicate value_labels (e.g. two customers with different ids, and the same first name, but the user has the ability to fix that up on their own) 

**Current**

```
<Dropdown 
 from=reviews 
 label="select first_name || ' ' || last_name from orders where id = order_id" 
 value=order_id 
 name=selected_order_id 
 where="nps_score > 7" />
```

**Proposed**

```
<Dropdown 
 from=reviews 
 label="Customer" -- the actual label on the 
 value=order_id -- the column that is being selected 
 value_label = first_name || ' ' || last_name 
 name=selected_order_id 
 where="nps_score > 7" 
 order= "nps_score desc"
 />
```

Resulting internal query 

```
select 
first_name || ' ' || last_name as value_label,
order_id as value, 
from reviews
where nps_score > 7
order by nps_score desc 
group by all -- dedupe 
```

mcrascal (approved):
LGTM!

Issue #1058, opened 2023-07-21
Formatting options for Tabs


archiewood:
## Feature Description

Ability to make tabs stand out more on the page

Ideas:
- Prop to make tabs take up full width
- Prop for background color to the tab picker

https://evidencedev.slack.com/archives/C023MMFSWKU/p1689943428457949

kuatroka:
The inspiration could be taken form here or just directly reuse it.
![tabs](https://github.com/evidence-dev/evidence/assets/9093546/c6fd803c-6dc2-4cdf-aa68-8b8375c6090d)

[Flowbite Tabs](https://flowbite-svelte.com/docs/components/tab)

commitsovercoffee:
@kuatroka I would like to work on this. Can you assign this to me ?

kuatroka:
Hi @commitsovercoffee that'd be awesome, but I can't assign anything as I'm not Evidenece's team member, but @archiewood can.

mcrascal:
@commitsovercoffee thanks for offering to help! This is all yours. 

I don't think we need to expose width as an option, just go full width by default.

PR #1087, opened 2023-08-08
Issue #1058 improve tab and tab picker visibility

commitsovercoffee:
### Checklist

### Description

Made below changes as mentioned in #1058 
1. Updated style for tabs and tab picker.
2. Added a prop named `activeTabColor` for background color of tab picker button.

|      Before      |      After      |
| before-1058 | after-1058 |

### Queries

- I have made tabs to take full-width using `flex-1` , was this the ask ?

|      After (with 2 tabs)      |      After  (with 5 tabs)    |
| two-tabs | five-tabs |

- Shall I add styling for dark-mode also ?


archiewood:
Hey @commitsovercoffee,

Thanks so much for the PR. Functionally speaking this is excellent.

I see the styling pretty closely matches the example provided by another user in the original issue. I think that implementation is stylistically different from our project, so I have some suggestions on how we can match the Evidence "look and feel" a bit better.

- Keep the body of the tabs as transparent, rather than grey
- Maintain the same default tab header shape as before, but add the background colour when a tab is active
- A slightly more subtle active background colour. Perhaps a lighter variant of the accent colour you have currently, or else a grey
- Can we remove the  `{label}` from the body of the tab. I see that the example provided in the issue had it, but I think for our implementation it would be more flexible for users if it wasn't included.

Once again - thank you for the contribution!

commitsovercoffee:
@archiewood Thanks a lot for your feedback! I'll work on these changes to ensure the implementation blends seamlessly with Evidence's design :)

commitsovercoffee:
Hey @archiewood, Please confirm the style that suits best. 

_GIF 1_
changes

For the border, I have updated the border shape with `rounded-t`. When you say _**Maintain the same default tab header shape**_, Should I keep the bottom border also ( similar to original design ) as shown below. 🤔

_GIF 2_
bottom-blue

_GIF 3_
active-bottom

To implement GIF 2/3, a prop need to be exported to define bottom border color. 🧐

archiewood:
Style 3 looks good to me!

Depending on how you implement, you could maybe just have one colour, with the background being a more transparent version of the line underneath?

commitsovercoffee:
### Description

Added a `color` prop which takes a 6 digit hex color code and sets it as bottom border color. Also it sets the background color for the active tab to the provided color with 0.1 opacity. If color is not provided it defaults to the original color.

### Implementation

Passing string as a variable is not allowing to change the opacity with rgba(). But passing the color as an RGB array works. So I have written a function to convert the provided hex color code into an RGB array, which is passed as `--activeTabColor`, a css variable and is used with rgba() to vary opacity where needed.

| without color prop | with color prop |
| defaut | custom |

## Usage

```svelte
## Tabs

<Tabs id="my-tabs" color="#96b6c5">  <!-- or "96b6c5" -->
    <Tab label="Tab 1">
        This is the content of Tab 1
    </Tab>
    <Tab label="Tab 2">
        This is the content of Tab 2
    </Tab>
</Tabs>
```

## Limitations 

value passed to `color` prop should be a 6 digit hex color code, which is a common way to describe color. If needed I can add rgb or hsl support but that seemed overkill hence mentioning here before adding it.
Please let me know if any changes are required.

Let me know if any changes are needed .

archiewood:
In all our other colour props, we support (CSS name | hexademical | RGB | HSL)

eg https://docs.evidence.dev/components/bar-chart#series

While I think CSS name is overkill (and maybe hard), if possible it would be great to support hex, rgb and HSL so users are not confused by this.

I believe you can implement transparency in all of these relatively easily:
- hex: https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color
- RGB: as you've done
- HSL: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl

commitsovercoffee:
Shall I follow the same approach though ? As I did for hex color input ? 

```svelte
<Tabs id="my-tabs" color="#96b6c5">
```
- `color` prop takes only string input. That's okay right ?
- a function verifies if it is a valid hex color. And if so it is passed as a css var to style the tab button.

So I'll just have to extend the function to handle rgb, hsl and css colors ?

archiewood:
string is how we handle color elsewhere so I think fine to continue with that.

My thought would be just reading the start of the string to work out what type of code it is?

```
if (string starts with #) { 
    // hex 
    add 10% opacity by appending `1a`, `##fa1167 -> `##fa11671a`
}
else if (string starts with RGB or rbg) {
    // rgb 
    add 10% opacity via css
} 
else if (string starts with hsl) {
    // hsl
    add 10% opacity by appending to string
}
```
or something like that?

.changeset/red-brooms-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@evidence-dev/core-components': major
```
ItsMeBrianD, line 2:
```suggestion
'@evidence-dev/core-components': patch
```

commitsovercoffee:
@archiewood I have made the required changes. 

```svelte
<Tabs id="my-tabs" color="hsl(166, 26%, 48%)">
```

image

```svelte
<Tabs id="my-tabs" color="rgb(160, 118, 249)">
```

image

## Alternative Implementation 

Instead of passing a prop to `<Tabs/>` to define active color, we can also send custom css vars directly to a component as mentioned in the svelte [tutorial](https://learn.svelte.dev/tutorial/component-styles), doing so we can directly use the passed color (including css colors) in the component style, and wont need to a function. 

But a bit weird syntax : 
```svelte
<Tabs id="my-tabs" --color=red>
<Tabs id="my-tabs" --color=rgb(160, 118, 249)>
```

Should I do it this way instead ?

PS @ItsMeBrianD I have updated the changeset to patch :)

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
```diff
@@ -12,6 +12,33 @@
 	 */
 	export let id;
 
+	/**
+	 * color can be provided to set custom background color for active tab.
+	 * @type {string}
+	 */
```
ItsMeBrianD, line 18: 🚀  Thanks for adding types for the props

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
```diff
@@ -40,32 +67,32 @@
...
-			@apply border-blue-500
-            hover:border-blue-600 hover:bg-gray-200
-            active:border-blue-700 active:bg-gray-100;
-		}
+	nav button:not(.active) {
+		@apply text-gray-400 hover:bg-gray-200 hover:text-gray-600 active:bg-gray-100;
+	}
+
+	button.active {
+		@apply text-black active:bg-gray-100;
```
ItsMeBrianD, line 94: I believe you can refactor this a little bit to avoid using `@apply`, it has [some performance concerns](https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply).

In the `<script>` section; declare a class string for your `:not(.active)`:

```javascript
const classes = {
    notActive: "text-gray-400 hover:bg-gray-200 hover:text-gray-600 active:bg-gray-100",
    active: "text-black active:bg-gray-100"
}
```

Then in the class string:
```
class="mt-2 p-4 rounded-t flex-1 
            text-sm font-sans 
            whitespace-nowrap
            transition duration-200 ease-in
           { $tabItems.active === tab.id ? classes.active : classes.notActive }
"
```


You can also reference `--bgColor` and `--borderColor` using [tailwind arbitrary values](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values). You can then include this in your
class strings above

That would look something like this:
```
bg-[--bgColor]
border-b-color-[--borderColor]
border-b-2. (check that this matches what is expected)
border-solid
```

sites/example-project/src/pages/ui-components/tabs/+page.md
```diff
@@ -1,6 +1,6 @@
 ## Tabs
 
-<Tabs id="my-tabs">
```
ItsMeBrianD, line 3: Can we add a second set of tabs on this page that demonstrate how to use the colored variant?

ItsMeBrianD (commented):
Hey @commitsovercoffee, thanks for the PR!

This looks great, I have a few tips to clean it up a little bit; let me know if you have any questions! 🚀

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
commitsovercoffee: One clarification please, 
> bg-[--bgColor]
>border-b-color-[--borderColor]
>border-b-2. (check that this matches what is expected)
>border-solid

Shouldn't the above go in the active property of classes ?
```svelte
const classes = {
    notActive: "text-gray-400 hover:bg-gray-200 hover:text-gray-600 active:bg-gray-100",
    active: "text-black active:bg-gray-100"
}
```

sites/example-project/src/pages/ui-components/tabs/+page.md
commitsovercoffee: Sure. Is this okay ? 

image

sites/example-project/src/pages/ui-components/tabs/+page.md
```diff
@@ -1,10 +1,19 @@
...
-        This is the content of Tab 1
+        This is the content of Tab 1.
     </Tab>
     <Tab label="Tab 2">
-        This is the content of Tab 2
+        This is the content of Tab 2.
+    </Tab>
+</Tabs>
+
+<Tabs id="my-tabs" color="#f6635c">
```
ItsMeBrianD, line 12:
```suggestion
<Tabs id="my-colored-tabs" color="#f6635c">
```

ItsMeBrianD:
> @ItsMeBrianD I've made the changes. Thanks for the detailed feedback :)

Absolutely, glad to help; thank you for submitting the PR!

I'm going to kick off the tests once that last suggestion is added, and then I think we're good to go 😄

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
```diff
@@ -16,23 +16,27 @@
...
-		const hexDigits = hexColor.replace(/\s+/g, '').replace('#', '');
+	const bgColor = isValidColorString(color) ? addOpacityToColor(color) : '#268aed1a';
+	const borderColor = isValidColorString(color) ? color : '#268aed';
 
-		// Check if the provided color contains valid hex characters.
-		if (!/^[0-9a-fA-F]{6}$/.test(hexDigits)) {
-			return null;
-		}
+	function isValidColorString(inputColor) {
+		const hexRegex = /^#([0-9A-Fa-f]{3}|[0-9A-Fa-f]{6})$/;
```
archiewood, line 31: Since you've done the hard work to create the regex strings, should you put them into an e.g. isHex(inputColor) function which you can then reuse below when adding opacity?

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
```diff
@@ -16,23 +16,27 @@
...
+		const rgbRegex = /^rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/;
```
archiewood, line 32: is this regex case sensitive?

I believe "RGB(255,0,0)" is a valid css colour, for example

archiewood (changes requested):

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
```diff
@@ -6,12 +6,44 @@
...
+
+	function isValidColorString(inputColor) {
+		const hexRegex = /^#([0-9A-Fa-f]{3}|[0-9A-Fa-f]{6})$/;
+		const rgbRegex = /^rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/;
+		const hslRegex = /^hsl\(\s*(\d{1,3})\s*,\s*(\d{1,3}%)\s*,\s*(\d{1,3}%)\s*\)$/;
+		return hexRegex.test(inputColor) || rgbRegex.test(inputColor) || hslRegex.test(inputColor);
+	}
+
+	function addOpacityToColor(colorString) {
+		if (colorString.startsWith('#')) {
```
archiewood, line 38: ... `if( isHex(colorString) )`

packages/core-components/src/lib/unsorted/ui/Tabs/Tabs.svelte
commitsovercoffee: Yes this is case sensitive. But "RGB(255,0,0)" works since prior to calling this function, the input string is converted into lowercase.

```svelte
color = color.replace(/\s+/g, '').toLowerCase(); // clean string
```

archiewood (approved):
This is a great improvement, thank you!

I've just done a couple of things:

- Added some more examples
- Reduced some of the spacing around the tabs
- bumped up a couple of the grays to make them more visible

PR #1094, opened 2023-08-10
Add delta content type to datatable

hughess:
### Description
Adds a new `contentType` to `Column` called `delta` (name TBC).

Props/options in `Column`:
- `contentType=delta`
- `downIsGood` - true/false - same as in BigValue. Default false (positive = green, negative = red)
- `deltaSymbol` - true/false - shows or hides triangle symbol

### Example

```html
    <Column id=yoy contentType=delta fmt=pct title="Y/Y Chg"/>
```

CleanShot 2023-08-10 at 17 59 34@2x

#### With `deltaSymbol=false`:
CleanShot 2023-08-10 at 17 59 45@2x

#### Custom Format
You can also use custom formatting to replace the symbol, like in the Value column below:

```html
    <Column id=value_usd contentType=delta fmt='[>=0]"+ "$#,##0;"- "$#,##0' deltaSymbol=false/>
```

CleanShot 2023-08-10 at 18 19 09@2x

### Considerations
Ideally, would have wanted symbol on the left, but it's difficult to align everything well. We need the symbol aligned on the left and the numbers aligned on the right, without too much space showing up in between. `display: flex` and `justify-content: space-between` almost gets there.

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/viz/DataTable.svelte
+++ b/packages/core-components/src/lib/unsorted/viz/DataTable.svelte
@@ -434,6 +434,26 @@
 												)}
 											{/if}
 										</a>
+									{:else if column.contentType === 'delta' && row[column.id] !== undefined}
+										<div class="m-0 text-xs font-medium font-ui" style={`color:${((row[column.id] >= 0 && !column.downIsGood) || (row[column.id] < 0 && column.downIsGood)) ? 'var(--green-700)' : 'var(--red-700)'}`}>
+											<div style="text-align: right;">
+												<span>
+												{formatValue(
+													row[column.id],
+													column.fmt
+														? getFormatObjectFromString(
+																column.fmt,
+																safeExtractColumn(column).format.valueType
+														)
+														: safeExtractColumn(column).format,
+													safeExtractColumn(column).columnUnitSummary
+												)}
+												</span>
+												{#if column.deltaSymbol}
+													<span>{@html row[column.id] >= 0 ? '&#9650;' : '&#9660;'}</span>
+												{/if}
+											</div>
+										</div>
 									{:else}
 										{formatValue(
 											row[column.id],
--- a/.changeset/curvy-schools-complain.md
+++ b/.changeset/curvy-schools-complain.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/core-components': minor
+'@evidence-dev/components': minor
+---
+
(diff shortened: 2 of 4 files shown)


packages/core-components/src/lib/unsorted/viz/DataTable.svelte
```diff
@@ -434,6 +434,34 @@
...
+											style={`color:${
+												(row[column.id] >= 0 && !column.downIsGood) ||
+												(row[column.id] < 0 && column.downIsGood)
+													? 'var(--green-700)'
+													: 'var(--red-700)'
+											}`}
+										>
+											<div style="text-align: right;">
+												<span>
+													{formatValue(
```
archiewood, line 449: Do you think users might like the ability to use _just_ the triangle indicator as well? (Even more compact way to scan data)

So you'd be able to have a column with just triangles pointing up or down?

Could be a `showValue` prop, for example

archiewood (commented):
Love the feature.

Other thoughts.
- Add the option to the docs?
- I think it would be better if we could show the triangle first. 
   - Understand the formatting issue
   - Perhaps we could left pad the numbers with whitespace, detecting when any numbers are "too short" relative the the maximum number of displayed characters

packages/core-components/src/lib/unsorted/viz/DataTable.svelte
hughess: I like that idea - I've just added it as `showValue`. I wonder if the props that are specific to a content type should have a prefix for the content type (E.g., `deltaShowValue`). I'm not sure if that's clearer or more confusing

hughess:
> Love the feature.
> 
> Other thoughts.
> 
> * Add the option to the docs?
> * I think it would be better if we could show the triangle first, not a deal breaker
>   
>   * Understand the formatting issue
>   * Perhaps we could left pad the numbers with whitespace, detecting when any numbers are "too short" relative the the maximum number of displayed characters in other rows.

I'll add this to the docs.

I agree on the appearance - I'd also rather have it on the left. Will give it a quick try, but if it turns out to be difficult, I think we should get this one out and add a left-side version later

mcrascal:
I think icon on the right side of the number looks sick. 

I would never use it with the icon on the left and the decimal points misaligned in the numbers. My vote would be to just ship it defaulted to the right side without the option to move it.

ItsMeBrianD (approved):

PR #1097, opened 2023-08-11
Conditional formatting for DataTable columns

hughess:
### Description
Adds conditional formatting as a content type in the `Column` component.

To do:
- Clean up naming
- Adjust colours

Options:
- `contentType=colorscale`
- `colorMax` - can pass a value to treat as the maximum of the column (e.g., if you set it to 100, anything above 100 would receive the max colour)
- `scaleColor` - "green" (default), "blue", or "red"

CleanShot 2023-08-11 at 10 07 10@2x
CleanShot 2023-08-11 at 18 52 38@2x

### Checklist

### Considerations

- Should we offer `downIsGood` like in BigValue, to reverse the direction of the colour scale?
- Multi-colour scales/heatmaps
   - I think this could be done by combining 2 colour scales - basically we’d split the column’s data into 2 based on value and apply a different colour scale to each (e.g., above median values go on the green scale, below median values go on a reversed red scale)
- Setting custom minimum for the scale (similar to `colorMax`)
   - Right now the colour is determined by the % of the max value in the column
   - If you have a lot of similar values - e.g., 122, 124, 125, 126 - the colours will look the same
   - We might want to offer a custom minimum like colorMax or we might need logic to use the min value as the floor for the scale calculation, rather than 0 as it’s doing now
- If you have a small # of columns, the width of each column can vary a lot, so you might have 2 metric columns where the width of one is double the width of the other


mcrascal:
For v1 of this, we're going to just add the color min functionality, and look at reversing the scale, creating diverging scales, and the comment on cell width in the future.

csjh:
Added `colorMin` to default to lowest value of the column, original approach would get weird at negative values

jackwelty:
Big +1 to multi-color scales, and the ability to add custom colors.

jackwelty:
> @jackwelty I think I know what you mean, but can you give an example of how you would use those?
> 
> For multi-color, I’m imagining a diverging scale (e.g., green to red).
> 
> For custom color, I’m imagining setting a color and having a gradient created from that color. The other way to do this one would be to manually set each color in the scale (e.g., 10 colors, effectively creating buckets)

so sorry for the delay here but that's exactly what I was thinking! One item I would add is custom diverging colors, i.e. red - white - blue, with white as the midpoint.

PR #1098, opened 2023-08-12
adds stepping prop for line & area chart.

### LINKED ISSUE #937: Stepped Line Chart
## Feature Description
Option to treat line chart as a "stepped" line rather than interpolation between points.

I think this is a prop to include in `LineChart` and `AreaChart` - naming TBD:
```html
<LineChart
   data={mydata}
   x=date
   y=sales
   stepped=true
/>
```

## Goal of Feature
More accurately show changes over time, especially for metrics where the values swing up and down significantly.

## Current Solution
Use a typical line chart

## Examples / Screenshots
CleanShot 2023-06-08 at 14 22 51@2x

## Implementation Notes
This is a feature within ECharts, so shouldn't be too hard to add:
https://echarts.apache.org/en/option.html#series-line.step

commitsovercoffee:
### Checklist

### Description
Attempts to resolve #937

- I have added prop named `stepping` to lineChart and areaChart. As a result, either of these charts can now have step line if stepping prop is given.
- I have also added them in the example project's line & area chart pages respectively.

| Line Chart | Area Chart |
| line-chart | area-chart |

### Usage

`stepping` defines whether to show as a step line. It can be true, false. Or 'start', 'middle', 'end'. Which will configure the turn point of step line. [Reference](https://echarts.apache.org/en/option.html#series-line.step)

```svelte
<AreaChart data={simpler_bar} x=year y=value series=country stepping={true}/>
<LineChart data={simpler_bar} x=year y=value series=country stepping={"start"}/>
```
Please let me know if any changes are required.


commitsovercoffee:
While working on this, I noticed that the [symbolSize](https://apache.github.io/echarts-handbook/en/how-to/chart-types/scatter/basic-scatter/#symbol-size) or MarkerSize is very small for area chart compared to other charts. Is that an issue or is it by design ? 🤔 

| Line Chart | Area Chart |
| line-size | area-chart |

As shown above, in the line chart it is clear where we are pointing at any instance as the dot is clearly visible on the line. But such is not the case for the area graph.

hughess:
@commitsovercoffee this is great, thanks!

What's the default position of the step if you set it to `true`? (i.e. start, middle, end)

There's a couple of minor things I would change for the prop:
- I'd call the prop `step`
- To keep the syntax looking as simple as possible, make the prop accept strings rather than requiring the curly braces, so it would look like `<LineChart step=true />`
   - The `markers` prop in `Line.svelte` is a good example of that

hughess:
> While working on this, I noticed that the [symbolSize](https://apache.github.io/echarts-handbook/en/how-to/chart-types/scatter/basic-scatter/#symbol-size) or MarkerSize is very small for area chart compared to other charts. Is that an issue or is it by design ? 🤔
> 
> As shown above, in the line chart it is clear where we are pointing at any instance as the dot is clearly visible on the line. But such is not the case for the area graph.

Good catch! I think those should be the same. Feel free to open an issue for that one, or add it in a PR if you're interested!

hughess:
@commitsovercoffee that sounds like a good idea. I'd call the second prop `stepPosition`.

Can we make the default "end"? I think that makes the most sense since if there's a spike, it would appear at the specific point rather than the point before

commitsovercoffee:
@hughess I have incorporated the changes you asked :)

- [ ] added `step` prop with _true_ defaulting to end position. 
- [ ] added `stepPosition` which can be 

Reference Syntax :
```svelte
<LineChart data={simpler_bar} x=year y=value series=country step=true />
<LineChart data={simpler_bar} x=year y=value series=country step=true stepPosition=start/>
```
Reference Image :

image

hughess:
This looks perfect!

Last thing to get this in is just docs - can you add this to the line and area docs? I think it’ll be an extra 2 rows in the props table for each of those charts, plus a new example for each (like you’ve got in example-project).

commitsovercoffee:
I see the images in examples are all in `svg` format. How do I save a chart in svg ?

archiewood:
@hughess is your man for this. Otherwise I have no issue with png

hughess:
PNG is good!

.changeset/brave-jokes-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@evidence-dev/core-components': major
+'@evidence-dev/components': major
```
ItsMeBrianD, line 3:
```suggestion
'@evidence-dev/core-components': patch
'@evidence-dev/components': patch
```

packages/core-components/src/lib/unsorted/viz/Area.svelte
```diff
@@ -30,6 +30,10 @@
 
 	export let handleMissing = 'gap';
 
+	export let step = false; // shows step line. It can be true, false. Or 'start', 'middle', 'end' to configure turn point of step line.
```
ItsMeBrianD, line 33:
```suggestion
        /**
         * Configures the turn point of the step line.
         * @type {boolean | 'start' | 'middle' | 'end' }
         */
	export let step = false; // shows step line. It can be true, false. Or 'start', 'middle', 'end' to configure turn point of step line.
```

packages/core-components/src/lib/unsorted/viz/Area.svelte
```diff
@@ -30,6 +30,10 @@
...
+	$: step = step === 'true' || step === true;
```
ItsMeBrianD, line 34: I'm not entirely sure I'm following, what is the expected behavior if step is `'start', 'middle', or 'end'` here?

packages/core-components/src/lib/unsorted/viz/Line.svelte
```diff
@@ -32,6 +32,10 @@
 
 	export let handleMissing = 'gap';
 
+	export let step = false; // shows step line. It can be true, false. Or 'start', 'middle', 'end' to configure turn point of step line.
+	$: step = step === 'true' || step === true;
+	export let stepPosition = 'end';
```
ItsMeBrianD, line 37: Same as above in `Area.svelte`

sites/example-project/src/pages/charts/line-chart/+page.md
```diff
@@ -92,7 +92,7 @@ select '2023-04-14' as start_date, null as end_date, 'Campaign C' as label
 
 <LineChart data={simpler_bar} x=year y=value series=country/>
 
-## Muliple y Column Line
+## Multiple y Column Line
```
ItsMeBrianD, line 95: Good catch!

ItsMeBrianD (commented):
Example photos look great!
I have a couple questions about the `step` prop, but otherwise this looks good to me

packages/core-components/src/lib/unsorted/viz/Area.svelte
```diff
@@ -30,6 +30,10 @@
...
+	export let stepPosition = 'end';
```
ItsMeBrianD, line 35:
```suggestion
        /**
         * Configures position of steps (e.g. before or after)
         * @type {'start' | 'middle' | 'end' }
         */
	export let stepPosition = 'end';
```

ItsMeBrianD:
@archiewood anything else needed here?

archiewood:
No, looks great

PR #1112, opened 2023-08-17
Install method cleanup

archiewood:
### Description

Recommends VSCode as the primary install method

This does currently make the install page longer which I'm not super happy about

Experimented with moving the other install methods to another page, but it feels like an annoying extra redirection if you wanted the CLI install.

### Before

| Docs | Readme |
| ----- | --------| 
| CleanShot 2023-08-17 at 12 17 09@2x | CleanShot 2023-08-17 at 12 20 28@2x |

### After
| Docs | Readme |
| ----- | --------| 
| CleanShot 2023-08-17 at 12 22 41@2x | CleanShot 2023-08-17 at 12 19 17@2x |

### Checklist


archiewood:
The one-page option looks like this:

CleanShot 2023-08-17 at 12 31 11@2x

PR #1118, opened 2023-08-18
1117 - Make Column component reactive to props changing.

### LINKED ISSUE #1117: DataTable Column titles aren't reactive
## Steps To Reproduce

Have a data table Column's title prop be assigned to a reactive svelte variable. When the variable value changes, the column title does not.

We need this because we have filters that change aggregation dynamically, so the column title should sometimes change dynamically.

## Expected Behavior

Column title updates dynamically.

## Actual Behaviour

Column title does not change.

## Workarounds

Use a column title that makes sense regardless of context.

jdimmerman:
### Description

Make Column component reactive to props changing.

Fixes #1117 

### Checklist


packages/core-components/src/lib/unsorted/viz/Column.svelte
```diff
@@ -56,35 +56,61 @@
...
+					...d.columns.slice(0, matchingIndex),
+					options,
+					...d.columns.slice(matchingIndex + 1)
+				];
+			}
+			return d;
+		});
+	};
+	$: id,
+		title,
```
jdimmerman, line 102: Is there a more succinct way of effectively indicating "all props" here?

ItsMeBrianD: You may be able to reference `$$props`, but I'm not sure if that's reactive

jdimmerman: After some quick testing, getting unexpected results and bad performance. I think it gets a new reference every time even if the specific constituent props didn't change, which would cause the columns to constantly rerender.

ItsMeBrianD: Ah yeah that's not ideal

Options could be declared this way potentially:

```javascript
$: options = {
    id, title, // ...
}
```

ItsMeBrianD: Awesome! I'm glad that worked

packages/core-components/src/lib/unsorted/viz/Column.svelte
```diff
@@ -66,7 +66,8 @@
 	export let deltaSymbol = true;
 	deltaSymbol = deltaSymbol === 'true' || deltaSymbol === true;
 
-	let options = {
+	let options;
```
ItsMeBrianD, line 69: You don't actually _need_ this declaration; Svelte adds it automatically for you based on the line below

archiewood, line 69: Is this necessary now you have the `$: options `

packages/core-components/src/lib/unsorted/viz/Column.svelte
```diff
@@ -83,8 +84,20 @@
 		showValue: showValue
 	};
 
-	props.update((d) => {
-		d.columns.push(options);
-		return d;
-	});
+	const updateProps = () => {
```
ItsMeBrianD, line 87:
```suggestion
        /**
         * Ensures that column props (e.g. title) are reflected in the table's state.
         * Without this function, props are only used on first render, and are not reactive
         * @returns {void}
         */
	const updateProps = () => {
```

ItsMeBrianD (approved):
A small note, and some comments

This looks great, thank you for submitting it!

packages/core-components/src/lib/unsorted/viz/Column.svelte
jdimmerman: TIL: nope!

jdimmerman: Very cool, I didn't know that. Removed, thanks

jdimmerman:
@ItsMeBrianD do I need to do anything here to kick off the tests following your approval?

ItsMeBrianD:
> @ItsMeBrianD do I need to do anything here to kick off the tests following your approval?

Yes, I just kicked them off 🚀

jdimmerman:
> > @ItsMeBrianD do I need to do anything here to kick off the tests following your approval?
> 
> Yes, I just kicked them off 🚀

@ItsMeBrianD same thing for the Tests-sources by any chance?

ItsMeBrianD:
> > > @ItsMeBrianD do I need to do anything here to kick off the tests following your approval?
> > 
> > 
> > Yes, I just kicked them off 🚀
> 
> @ItsMeBrianD same thing for the Tests-sources by any chance?

I'm not sure why those haven't been kicked off; I don't have anything to start them

ItsMeBrianD:
Looks like I could do it manually

archiewood:
Approval has to be done separately for the docs / workspace builds and the tests. Not sure why tbh,

jdimmerman:
(This is probably obvious but just in case) I don't have access to merge, I'm just assuming someone on your team will when it looks good. Thanks!

PR #1122, opened 2023-08-21
`getCompletedData` tests & bigfixes

ItsMeBrianD:
### Checklist


csjh:
fwiw, when I was testing this one out, seemed like it only happened with `Date` objects, and seemed related to some code in `tidyjs` that used a `Set` to get all unique values. Since `new Set([new Date(), new Date()]).length == 2`, it caused some weirdness. Didn't get to the bottom of a fix though.

packages/component-utilities/src/getCompletedData.js
```diff
@@ -1,132 +1,83 @@
...
+/**
+ * This function fills missing data points in the given data array for a specific series.
+ *
+ * @param {Array<object>} data - The data as an array of objects.
+ * @param {string} x - The property used as x-axis.
+ * @param {string} y - The property used as y-axis.
+ * @param {string} series - The specific series in the data to be filled.
+ * @param {boolean} [nullsZero=false] - A flag indicating whether nulls should be replaced with zero.
+ * @param {boolean} [fillX=false] - A flag indicating whether the x-axis values should be filled (based on the found interval distance).
+ * @return {Array<object>} An array containing the filled data objects.
```
csjh, line 14:
```suggestion
 * @param {Record<string, unknown>[]} data - The data as an array of objects.
 * @param {string} x - The property used as x-axis.
 * @param {string} y - The property used as y-axis.
 * @param {string} series - The specific series in the data to be filled.
 * @param {boolean} [nullsZero=false] - A flag indicating whether nulls should be replaced with zero.
 * @param {boolean} [fillX=false] - A flag indicating whether the x-axis values should be filled (based on the found interval distance).
 * @return {Record<string, unknown>[]} An array containing the filled data objects.
```
`object` will make Typescript whine if you access keys on the object

packages/component-utilities/src/getCompletedData.js
```diff
@@ -1,132 +1,83 @@
...
-	return interval;
-}
-
-export default function getCompletedData(data, x, y, series, nullsZero = false, fillX = false) {
-	let xDistinct = getDistinctValues(data, x);
-	let interval;
-	let filledData;
+		if (value[0]?.[x] instanceof Date) {
+			xDistinct = getDistinctValues(
+				value.map((d) => ({ ...d, [x]: d[x].getTime() })),
```
csjh, line 37:
```suggestion
				value.map((d) => ({ [x]: d[x].getTime() })),
```
Only `x` is accessed in the created array so shouldn't bother spreading `d`

packages/component-utilities/src/getCompletedData.js
```diff
@@ -1,132 +1,83 @@
...
+			xIsDate = true;
+		} else {
+			xDistinct = getDistinctValues(value, x);
+		}
 
-	if (series) {
-		if (fillX) {
-			interval = findInterval(xDistinct);
+		/** @type {number} */
+		let interval = findInterval(xDistinct);
```
csjh, line 46: Should inline this into the if statement on line 59 since it isn't always used, same with `xDistinct`

packages/component-utilities/src/getDistinctValues.js
```diff
@@ -1,9 +1,6 @@
...
-	const distinctValueSet = new Set();
-	data.forEach((d) => {
-		distinctValueSet.add(d[column]);
-	});
-	distinctValues = [...distinctValueSet];
-	return distinctValues;
+	return data.reduce((a, v) => {
+		if (a.includes(v[column])) return a;
+		return [...a, v[column]];
+	}, []);
```
csjh, line 5:
```suggestion
    const set = new Set(data.map((val) => val[column]));
	return Array.from(set);
```
`O(n^2)` -> `O(n)`

csjh, line 5:
```suggestion
    const set = new Set(data.map((val) => val[column]));
	return Array.from(set);
```
`O(n^2)` -> `O(n)`

packages/component-utilities/src/helpers/getCompletedData.helpers.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * This function is used to find the minimum and maximum values in an array.
+ *
+ * @param {Array} values - An array from which min and max values should be determined.
+ * @param {Function} [valueof] - An optional function that defines how to obtain the measuring value.
+ * @return {Array} An array containing the minimum and maximum of numbers, respectively.
+ */
+export function extent(values, valueof) {
```
csjh, line 45: Adding the identity function as a default for `valueOf` would allow chopping the duplicate for loop

packages/component-utilities/src/helpers/getCompletedData.helpers.js
```diff
@@ -0,0 +1,125 @@
...
+			} else {
+				if (min > value) min = value;
+				if (max < value) max = value;
+			}
+		}
+	} else {
+		let index = -1;
+		for (let value of values) {
+			if (typeof value !== 'number') continue;
+			if ((value = valueof(value, ++index, values)) != null) {
```
csjh, line 64: `index` only increments if `typeof value === 'number'`, which might not be correct. Though I don't think the `valueOf` parameter is used anywhere so might not matter atm

packages/component-utilities/src/helpers/getCompletedData.helpers.js
ItsMeBrianD: This function was left as-is from previous implementation, there's a lot of room for improvement in this file. Going to leave as-is because this PR is already over estimate

packages/component-utilities/src/helpers/getCompletedData.helpers.js
ItsMeBrianD: Same as above

PR #1126, opened 2023-08-23
Ensure database output lines up with assigned `EvidenceType`

csjh:
### Description

Makes sure all databases return their proper types (numbers are numbers, booleans are booleans, dates are dates, etc)

### Checklist


packages/db-commons/index.cjs
```diff
@@ -86,21 +86,163 @@ const inferColumnTypes = function (rows) {
...
+
+/**
+ * Checks if all the columns of type `evidenceType` in a set of rows
+ * conform to the proper `typeof` or `instanceof`
+ * @param {Record<string, unknown>[]} rows
+ * @param {ColumnDefinition[]} columns
+ * @param {EvidenceType} evidenceType
+ * @returns {boolean}
+ */
+function conformsTo(rows, columns, evidenceType) {
```
ItsMeBrianD, line 126: What the performance implications of these functions?

There was an issue with zod, where it was validating every row in the returned dataset that was causing a lot of slowdown; might be worth testing this with very (very) large arrays

ItsMeBrianD: Also; we should write unit tests for these added functions using `vitest`

I've found that AI Helpers (e.g. Jetbrains' AI Assistant) have been helpful in whipping up unit tests and test cases effectively

packages/db-commons/index.cjs
```diff
@@ -86,21 +86,163 @@ const inferColumnTypes = function (rows) {
...
+	}
+}
+
+/**
+ *
+ * @param {Record<string, unknown>[]} results
+ * @param {ColumnDefinition[]} columns
+ * @returns {Record<string, unknown>[]}
+ */
+function convertStringColumns(results, columns) {
```
ItsMeBrianD, line 153: Is this something that could be handled when loading into parquet files?

ItsMeBrianD (commented):
Couple things wrt to the helper functions; the connectors themselves lgtm

packages/db-commons/index.cjs
csjh: `conformsTo` is fast (only checks first row), the others are slow if there's a column that doesn't conform

We could definitely look into perf gains on the adapter level, functions which recreate every row like `standardizeRows` are present in a few adapters which probably aren't ideal

packages/db-commons/index.cjs
csjh: It could make sense to run it before parquet conversion yeah, arrow just calls `String` on non-strings by default so this would get a bit better compat for objects

ItsMeBrianD: I don't think we need to worry about support for objects /  dicts yet; we don't handle those very well in other places

ItsMeBrianD:
Going to pick this back up when we make changes to the Connector API post-release

PR #1128, opened 2023-08-23
Markdown Partials

ItsMeBrianD:
### Checklist


packages/preprocess/tsconfig.json
```diff
@@ -0,0 +1,113 @@
+{
+	"include": ["types/index.d.ts", "src/**/*.js"],
```
csjh, line 2: Should we include `.cjs` or is this meant to just cover the tests

PR #1132, opened 2023-08-25
Invisible links on datatable for crawler

csjh:
### Description

Adds invisible links for SvelteKit to crawl in `DataTable` and `USMap` components

### Checklist


packages/core-components/src/lib/atoms/InvisibleLinks.svelte
```diff
@@ -0,0 +1,14 @@
+<script>
+	import { building } from '$app/environment';
```
ItsMeBrianD, line 2: I would add a comment here explaining why this is needed / useful

packages/core-components/src/lib/atoms/InvisibleLinks.svelte
```diff
@@ -0,0 +1,14 @@
...
+
+	/** @type {Record<string, unknown>[]} */
+	export let data;
+	/** @type {string}*/
+	export let link;
+</script>
+
+{#if building}
```
csjh, line 10:
```suggestion
<!-- we only care about links being crawled during build time -->
{#if building}
```

PR #1142, opened 2023-08-29
remove broken conditional in `extractQueries.server.js`

csjh:
### Description

previously, a conditional in `extractQueries.server.js` would do two things when a `+page.md` is empty:
- return `[{}]`, causing `QueryToast.svelte` to try to read properties off the `{}`, causing a broken toast notification
- skip `updateDirectoriesandStatus`, which would normally update `.evidence-dev/queries/extracted/<hash>/queries.json` to have the proper extracted queries, which ends up making old queries stay on the page in evidence's eyes

### Checklist

### DIFF
--- a/sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js
+++ b/sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js
@@ -46,66 +46,58 @@ const updateDirectoriesandStatus = function (queries, routeHash) {
 export const getStatusAndExtractQueries = function (route) {
 	let routeHash = md5(route);
 	let fileRoute = `./src/pages/${route}/+page.md`;
-	let content = fs.readFileSync(fileRoute);
-	content = content ? content.toString() : null;
+	let content = fs.readFileSync(fileRoute, 'utf-8');
 
-	if (content) {
-		let partialInjectedContent = preprocessor.injectPartials(content.toString());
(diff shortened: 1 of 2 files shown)


sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js
```diff
@@ -46,66 +46,58 @@ const updateDirectoriesandStatus = function (queries, routeHash) {
...
-								if (!query.inline && referencedQuery.inline) {
-									throw new Error(
-										`Cannot reference inline query from SQL File. (Referenced ${referencedQueryID})`
-									);
-								}
-								const queryString = `(${referencedQuery.compiledQueryString})`;
-								query.compiledQueryString = query.compiledQueryString.replace(
-									reference,
-									queryString
+	for (let i = 0; i <= maxIterations; i++) {
```
Comment (ItsMeBrianD) [inline sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js:58]: This has far too many levels of nesting; can we do some refactoring in this PR to make this more legible?

Also a good opportunity to write tests before refactoring to ensure there aren't regressions

Comment (csjh) [reply sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js:58]: Is it worth refactoring and adding tests considering [it's all ripped out in next](https://github.com/evidence-dev/evidence/blob/next/sites/example-project/src/pages/api/status/%5B...route%5D/extractQueries.server.js)?

PR #1151, opened 2023-09-06
Ensure that x axis is sorted for numeric and dated charts

### LINKED ISSUE #1134: Sort by x axis by default in linechart
A line chart like this one with a numeric x axis should plot in the order of the x axis and render as a line, not a squiggle. Suspect this can be achieved by setting the `sort` option on linechart to default to `true`. 

image

ItsMeBrianD:
### Description

Resolves #1134

### Checklist


packages/component-utilities/src/echarts.js
```diff
@@ -424,6 +424,17 @@ export default (node, option) => {
 
 	const chart = init(node, 'evidence-light', { renderer: 'svg' });
 
+	// If the x-axis of a series is numeric, or a date; ensure that it is in order
+	option.series = option.series.map((s) => {
+		if (typeof s.data[0][0] === 'number') {
+			s.data = s.data.sort((a, b) => a[0] - b[0]);
+		} else if (s.data[0][0] instanceof Date) {
+			s.data = s.data.sort((a, b) => a.getTime() - b?.getTime() ?? 0);
```
csjh, line 432: Shouldn't this be `a[0].getTime() - b[0]?.getTime() ?? 0` or similar?

csjh: Also the `getTime()` isn't strictly necessary because JS implicitly converts Dates to numbers when you try to do arithmetic on them

ItsMeBrianD: > Shouldn't this be `a[0].getTime() - b[0]?.getTime() ?? 0` or similar?

Yes

```suggestion
			s.data = s.data.sort((a, b) => a[0].getTime() - b[0]?.getTime() ?? 0);
```

> Also the `getTime()` isn't strictly necessary because JS implicitly converts Dates to numbers when you try to do arithmetic on them

True, but this makes it explicit, which I think is helpful when you get into the weeds like this code is

csjh (approved):

packages/component-utilities/src/echarts.js
```diff
@@ -424,6 +424,17 @@ export default (node, option) => {
 
 	const chart = init(node, 'evidence-light', { renderer: 'svg' });
 
+	// If the x-axis of a series is numeric, or a date; ensure that it is in order
+	option.series = option.series.map((s) => {
+		if (typeof s.data[0][0] === 'number') {
+			s.data = s.data.sort((a, b) => a[0] - b[0]);
+		} else if (s.data[0][0] instanceof Date) {
```
csjh, line 431:
```suggestion
		if (typeof s.data?.[0][0] === 'number') {
			s.data = s.data.sort((a, b) => a[0] - b[0]);
		} else if (s.data?.[0][0] instanceof Date) {
```
Seems like `/charts/bar-chart` from `example-project` errors out w/o this

csjh (changes requested):

mcrascal:
@ItsMeBrianD -- I think I accidentally put PR review feedback into the issue. The concern here was that we want to turn on the existing sort option by default, rather than implementing new sort functionality, which I think is what's happening here.

ItsMeBrianD:
> @ItsMeBrianD -- I think I accidentally put PR [feedback into the issue](https://github.com/evidence-dev/evidence/issues/1134#issuecomment-1711753724). The concern here was that we want to turn on the existing sort option by default, rather than implementing new sort functionality, which I think is what's happening here.

Gotcha; I'll resolve in #1155

PR #1155, opened 2023-09-07
X Axis Label Tweaks

### LINKED ISSUE #1137: Improve default text wrapping in BarChart x Axis
Barcharts hide axis labels too often. 

A better default behaviour would: 

1. Allow for some text wrapping to multiple lines, 
2. Allow for some ellipsis on overflow for some labels 

Before going on to hide some axis labels as is the current behaviour. 

image

ItsMeBrianD:
## **Note**

This is merging into another feature branch until that one is closed. It branched off the target branch because of storybook tweaks that existed there.

### Description

This adds a new prop, `showAllXAxisLabels`, that forces the xAxis to show _every_ value, instead of eCharts' default behavior of only showing a subset. It does so by truncating with an ellipsis.

Resolves #1137

This also addresses feedback that was missed on https://github.com/evidence-dev/evidence/issues/1134#issuecomment-1711753724

### Checklist

### Screenshots

image
image


mcrascal:
This seems close -- I think showAllXAxisLabels should be defaulted yes for categorical axis, and false for all others.

hughess:
This is the case where I've most frequently seen this as an issue - a small number of categories, but missing labels. Truncated is probably better in this scenario, but I think the ideal solution would be wrapped text.

CleanShot 2023-09-08 at 11 06 11@2x


I looked into wrapping a while ago - ECharts lets you wrap text on axis labels, but for it to work it requires a specific width to be set on each axis label, so I assume you'd need a calculation to get (chart width / # of bars) and use that to set the width for the label container in ECharts.

mcrascal:
@hughess agreed that wrapping would be ideal, but the challenge is it looks like we can't easily control the number of lines when wrapping (e.g. stop at 3 lines then truncate). The concern is that for extremely long labels, the wrapping will expand to many lines and look horrible.

mcrascal:
@ItsMeBrianD looking at the categorical vars [here](https://deploy-preview-1155--evidence-development-workspace.netlify.app/charts/bar-chart/) I don't think this is coming on by default.

packages/core-components/src/lib/unsorted/viz/BarChart.svelte
```diff
@@ -31,7 +31,7 @@
 	export let yMin = undefined;
 	export let yMax = undefined;
 	export let swapXY = false;
-	export let showAllXAxisLabels = xType === 'category';
+	export let showAllXAxisLabels = xType === 'category' || typeof data?.[0]?.[x] === "string";
```
mcrascal, line 34: are we OK to rely on typeof here, instead of the evidence type shipped with the data object?

PR #1169, opened 2023-09-15
Fix for getCompletedData function

hughess:
### Description
A customer noticed an issue with a multi-series bar chart, which showed the series shifted across the x-axis. It turns out this was related to the `getCompletedData` function, which wasn't filling in missing values for the specific situation in that chart (a "missing Y" situation per the `example-project` terminology.

Issue can be seen here in a previous deploy preview (stacked area and bar charts for Missing Y):
https://deploy-preview-1166--evidence-development-workspace.netlify.app/chart-testing/dateSeries/

Here's what's been changed:
- Removed the loop around the tidyJS completion function  - each iteration didn't have access to the full combination of x and series needed to fill everything in
   - As part of this change, used `data` instead of `value` from the loop
- In the `nullySpec`, changed to referenced the `series` column name dynamically
- Added `x` to the `expandKeys` for scenarios where `x` is not to be filled  

### Checklist

### DIFF
--- a/.changeset/honest-dogs-invent.md
+++ b/.changeset/honest-dogs-invent.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/component-utilities': patch
+'@evidence-dev/components': patch
+---
+
+Patch for getCompletedData function
--- a/packages/component-utilities/src/getCompletedData.js
+++ b/packages/component-utilities/src/getCompletedData.js
@@ -27,8 +27,8 @@ export default function getCompletedData(data, x, y, series, nullsZero = false, 
 
 	const output = [];
 
-	let xIsDate = Object.values(groups)[0][0]?.[x] instanceof Date; 
-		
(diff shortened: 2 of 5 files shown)


ItsMeBrianD:
This LGTM, can we make sure there's a test case that covers this behavior?

hughess:
Yup will try to add one.

Getting a strange error on the PR tests - SIGTRAP. I get the same thing locally as well

ItsMeBrianD:
> Getting a strange error on the PR tests - SIGTRAP. I get the same thing locally as well

I'd be happy to take a look at this as well later this am

hughess:
@ItsMeBrianD I realized I hadn't documented the original function very well/at all, so I just added a page to example-project under Chart Testing > 1. Background with more of the info.

I've tried adding a couple of tests in `getCompletedData.spec`, but I'm not sure if I'm doing it correctly.

If it looks good as is (after the SIGTRAP issue), it might be worth releasing quickly so we've got a fix for the customer, and then can add the new tests separately

Issue #988, opened 2023-06-28
Universal SQL: What to do with QueryStatus


ItsMeBrianD:

ItsMeBrianD:
For re-execution it may be a good idea to allow database connectors to expose some function `onSourceFileChange` or similar; this would get passed the connector options and the filepath of the changed source file, and possibly a way for the database connector to invoke HMR / a page refresh (from vite/rollup).

Implementing it this way would allow connectors to take control of when things need to be re-executed (e.g. postgres probably only cares if the .sql files change, while duckdb cares if the .duckdb file changes _or_ if the .sql files change)

PR #1183, opened 2023-09-18
[next] source refresh experience

csjh:
### Description

Adds HMR-like behavior to source query changes while in dev mode

Solves the following from #988:
- [x] Polish the overall build sources system
- [ ] Offer a convenient way to re-execute source queries
^ partial, though could also trivially implement a button on QueryViewers
- [x] Offer enough feedback in dev mode on which queries are running (prevent a broken app state while they are running)
- [ ] Prevent booting the dev mode without built sources / auto build on boot, with some cacheing
^ this one should just be adding `evidence build:sources --changed` to template
- [ ] VS Code button to execute a source rebuild for all sources
- [x] Not need to re-boot dev server upon building sources
- [x] Remove query status as appropriate

### Checklist


packages/core-components/src/lib/unsorted/ui/QueryToast.svelte
```diff
@@ -19,15 +19,26 @@
...
+				Rebuilding
+			{:else}
+				Error while rebuilding
+			{/if}
+			{#if status.id.endsWith('.connection')}
+				all queries in {status.id.split('.')[0]}
+			{:else}
+				{status.id}
+			{/if}
 		</span>
```
ItsMeBrianD, line 42: We should drive this based on QueryStore in another PR

ItsMeBrianD (approved):

PR #1184, opened 2023-09-18
feat: enable custom sveltekit configs

### LINKED ISSUE #1182: HMR in VSCode devcontainer
> Hot reloading does not appear to be working with Evidence hosted in docker in on WSL. We are using VSCode devcontainer for development as a standard (for dev environment consistency).
This appears to be a vite issue but the workaround doesn’t seem possible with Evidence. Can I enable watch.usePolling in the vite config when running locally?
https://github.com/vitejs/vite/issues/1153

See: https://evidencedev.slack.com/archives/C023LRB9Z40/p1695060760559759

ItsMeBrianD:
### Description

An evidence project can now have a `svelte.config.js` file at the root of the project that will be merged _over_ Evidence's own configuration. Certain fields that are required for Evidence and it's components to function properly are ignored.

- Preprocessors form the basis of Evidence, users should not mess with them (yet)
- File locations are hard-coded into the CLI; changing them will completely break the app
- Extensions are also hard-coded in many places, and changing them will likely break everything.

Resolves #1182

### Checklist

### DIFF
--- a/sites/test-env/svelte.config.js
+++ b/sites/test-env/svelte.config.js
@@ -1,16 +1,15 @@
 /** @type {import("@sveltejs/kit").Config} */
 export default {
-    preprocess: [],
-    extensions: [],
-    kit: {
-        files: {}
-    },
-    vite: {
-        server: {
-            watch: {
-                usePolling: true
-            }
-        }
-    }
-}
-
+	preprocess: [],
+	extensions: [],
+	kit: {
+		files: {}
+	},
+	vite: {
+		server: {
+			watch: {
+				usePolling: true
+			}
+		}
+	}
+};
--- a/sites/test-env/svelte.config.js
+++ b/sites/test-env/svelte.config.js
@@ -1,0 +1,16 @@
+/** @type {import("@sveltejs/kit").Config} */
+export default {
+    preprocess: [],
+    extensions: [],
+    kit: {
+        files: {}
+    },
+    vite: {
+        server: {
+            watch: {
+                usePolling: true
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


mcrascal:
I get where your head is at, but I would strongly prefer to support HMR in a dev container out of the box without needing to override the builtin svelte config. 

Is there a reason we can't just make the required vite overrides directly to our included config?

csjh:
> Is there a reason we can't just make the required vite overrides directly to our included config?

`usePolling` isn't a good default, causes [high CPU usage](https://github.com/paulmillr/chokidar#performance)

mcrascal:
> usePolling isn't a good default, causes [high CPU usage](https://github.com/paulmillr/chokidar#performance)

Ah got it. I'm good w/ the approach in here.

sites/test-env/svelte.config.js
mcrascal, line 1: Probably shouldn't leave this behind :)

mcrascal (approved):
LGTM

sites/test-env/svelte.config.js
ItsMeBrianD: Left this file here as a test; there are a few console messages that are output when the test-env loads that are a sanity check

archiewood:
I'm thinking this is a good feature. I think there are a few things people might want to do with the svelte.config.js file that means it's helpful to have it as editable.

## Overriding defaults

We have an increasing number of these "drop in this file and we'll override the default" style escape hatches.

You can now do this with:
- app.html
- app.css
- +layout.svelte
- some other stuff?

It would be good if we could document this behaviour. Perhaps in https://docs.evidence.dev/themes-and-layouts/. 
(Not required for this PR)

PR #1200, opened 2023-09-20
Add redshift env vars to PG connector

### LINKED ISSUE #1199: Redshift Suggested Env Vars do not work
## Steps To Reproduce

Use the Redshift connector, look at the suggested env vars in the deploy menu.

image

### Environment
- Database: redshift

## Expected Behavior

These env vars should allow a project to deploy

## Actual Behaviour

A project will not deploy as the redshift connector is a thin wrapper on the postgres connector

## Workarounds

Use the postgres equivalent vars, as suggested in [docs](https://docs.evidence.dev/cli/#environment-variables)

archiewood:
### Description

fixes #1199

I feel like there is probably a more elegant way where we could put the env maps into the redshift connector directly, but I dont really understand cjs and therefore how to achieve this

### Checklist

### DIFF
--- a/packages/postgres/index.cjs
+++ b/packages/postgres/index.cjs
@@ -6,48 +6,72 @@ const envMap = {
 	host: [
 		{ key: 'EVIDENCE_POSTGRES_HOST', deprecated: false },
 		{ key: 'POSTGRES_HOST', deprecated: false },
+		// Redshift uses this env var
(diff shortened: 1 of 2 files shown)


csjh (approved):
It might be possible to put it in the redshift adapter, but probably not worth the effort - would have to rework a good amount of the logic in the pg connector

PR #1208, opened 2023-09-22
feat: query stores

ItsMeBrianD:
### Description

This PR adds `QueryStore`; which is a major shift in the way queries are handled by Evidence under the hood. The interface that a end user interacts with should still be available (and work the same way), but there will be additional functionalities available.

The `QueryStore` should:
 - Query for results lazily, instead of eagerly
 - Provide loading state information
 - Create the building blocks for more functionality in the future (e.g. `QueryStore` being aware of it's own error state, meaning consuming components can be aware as well)

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/viz/BigValue.svelte
+++ b/packages/core-components/src/lib/unsorted/viz/BigValue.svelte
@@ -143,8 +143,6 @@
 	{/if}
 </div>
 
-<pre>{JSON.stringify(data[0][0].duration)}</pre>
-
 <style>
 	/* 
         TODO: Identify if this can be moved to app.css, or scoped to this component.
--- a/packages/core-components/src/lib/unsorted/viz/BigValue.svelte
+++ b/packages/core-components/src/lib/unsorted/viz/BigValue.svelte
@@ -34,53 +34,54 @@
 	let sparklineData = {};
 
 	let error = undefined;
-	$: try {
-		error = undefined;
+	$: if (data)
+		try {
+			error = undefined;
 
-		if (!value) {
-			throw new Error('value is required');
-		}
+			if (!value) {
+				throw new Error('value is required');
+			}
 
-		if (!Array.isArray(data)) {
-			data = [data];
-		}
+			if (!Array.isArray(data)) {
+				data = [data];
+			}
 
-		checkInputs(data, [value]);
+			checkInputs(data, [value]);
 
-		let columnSummary = getColumnSummary(data, 'array');
+			let columnSummary = getColumnSummary(data, 'array');
 
-		// Fall back titles
-		let valueColumnSummary = columnSummary.find((d) => d.id === value);
-		title = title ?? (valueColumnSummary ? valueColumnSummary.title : null);
+			// Fall back titles
+			let valueColumnSummary = columnSummary.find((d) => d.id === value);
+			title = title ?? (valueColumnSummary ? valueColumnSummary.title : null);
 
-		if (comparison) {
-			checkInputs(data, [comparison]);
-			let comparisonColumnSummary = columnSummary.find((d) => d.id === comparison);
-			comparisonTitle =
-				comparisonTitle ?? (comparisonColumnSummary ? comparisonColumnSummary.title : null);
-		}
+			if (comparison) {
+				checkInputs(data, [comparison]);
+				let comparisonColumnSummary = columnSummary.find((d) => d.id === comparison);
+				comparisonTitle =
+					comparisonTitle ?? (comparisonColumnSummary ? comparisonColumnSummary.title : null);
+			}
 
-		if (data && comparison) {
-			positive = data[0][comparison] >= 0;
-			comparisonColor =
-				(positive && !downIsGood) || (!positive && downIsGood)
-					? 'var(--green-700)'
-					: 'var(--red-700)';
-		}
-		// populate sparklineData from data where timeseries is the key and value is the value
-		if (data && sparkline && value) {
-			// allow to load the LinkedChart
-			let sortedData = getSortedData(data, sparkline, true);
-			for (let i = 0; i < sortedData.length; i++) {
-				sparklineData[sortedData[i][sparkline]] = sortedData[i][value];
+			if (data && comparison) {
+				positive = data[0][comparison] >= 0;
+				comparisonColor =
+					(positive && !downIsGood) || (!positive && downIsGood)
+						? 'var(--green-700)'
+						: 'var(--red-700)';
+			}
+			// populate sparklineData from data where timeseries is the key and value is the value
+			if (data && sparkline && value) {
+				// allow to load the LinkedChart
+				let sortedData = getSortedData(data, sparkline, true);
+				for (let i = 0; i < sortedData.length; i++) {
+					sparklineData[sortedData[i][sparkline]] = sortedData[i][value];
+				}
+			}
+		} catch (e) {
+			error = e;
+			if (strictBuild) {
+				throw error;
 			}
 		}
-	} catch (e) {
-		error = e;
-		if (strictBuild) {
-			throw error;
-		}
-	}
 
 	/**
 	 * Hack to let time to LinkedChart to be loaded
@@ -141,6 +142,8 @@
 		{/if}
 	{/if}
 </div>
+
+<pre>{JSON.stringify(data[0][0].duration)}</pre>
 
 <style>
 	/* 
--- a/packages/preprocess/src/process-queries.cjs
+++ b/packages/preprocess/src/process-queries.cjs
@@ -12,31 +12,6 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 		const valid_ids = Object.keys(duckdbQueries).filter((queryId) =>
 			queryId.match('^([a-zA-Z_$][a-zA-Z0-9d_$]*)$')
 		);
-		// queryDeclarations += `
-		//     import debounce from 'debounce';
-		//     import { browser } from '$app/environment';
-		// 	import {profile} from '@evidence-dev/component-utilities/profile';
-		//
-		// 	// partially bypasses weird reactivity stuff with \`select\` elements
-		// 	function data_update(data) {
-		// 		${valid_ids.map((id) => `${id} = data.${id} ?? [];`).join('\n')}
-		// 	}
-		//
-		// 	$: data_update(data);
-		//
-		//
-		//     ${valid_ids
-		// 					.map(
-		// 						(id) => `
-		//         let ${id} = data.${id} ?? [];
-		//         const _query_${id} = browser
-		// 			  ? debounce((query) => profile(__db.query, query).then((value) => ${id} = value), 200)
-		// 			  : (query) => (${id} = profile(__db.query, query, "${id}"));
-		//         $: _query_${id}(\`${duckdbQueries[id].replaceAll('`', '\\`')}\`);
-		//     `
-		// 					)
-		// 					.join('\n')}
-		// `;
 
 		const queryStores = valid_ids.map(
 			(id) => `
@@ -44,7 +19,7 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 			\`${duckdbQueries[id].replaceAll('`', '\\`')}\`,
 			queryFunc,
 			\`${id}\`,
-			{ initialData: profile(__db.query, \`${duckdbQueries[id].replaceAll('`', '\\`')}\`) }
+			{ initialData: data.${id} ?? profile(__db.query, \`${duckdbQueries[id].replaceAll('`', '\\`')}\`) }
 		);
 		/** @type {QueryStore} */
 		let ${id};
--- a/packages/preprocess/src/process-queries.cjs
+++ b/packages/preprocess/src/process-queries.cjs
@@ -12,31 +12,55 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 		const valid_ids = Object.keys(duckdbQueries).filter((queryId) =>
 			queryId.match('^([a-zA-Z_$][a-zA-Z0-9d_$]*)$')
 		);
+		// queryDeclarations += `
+        //     import debounce from 'debounce';
+        //     import { browser } from '$app/environment';
+		// 	import {profile} from '@evidence-dev/component-utilities/profile';
+		//
+		// 	// partially bypasses weird reactivity stuff with \`select\` elements
+		// 	function data_update(data) {
+		// 		${valid_ids.map((id) => `${id} = data.${id} ?? [];`).join('\n')}
+		// 	}
+		//
+		// 	$: data_update(data);
+		//
+		//
+        //     ${valid_ids
+		// 					.map(
+		// 						(id) => `
+        //         let ${id} = data.${id} ?? [];
+        //         const _query_${id} = browser
+		// 			  ? debounce((query) => profile(__db.query, query).then((value) => ${id} = value), 200)
+		// 			  : (query) => (${id} = profile(__db.query, query, "${id}"));
+        //         $: _query_${id}(\`${duckdbQueries[id].replaceAll('`', '\\`')}\`);
+        //     `
+		// 					)
+		// 					.join('\n')}
+        // `;
+
+		const queryStores = valid_ids.map(id => `
+		const _${id} = new QueryStore(
+			\`${duckdbQueries[id].replaceAll('`', '\\`')}\`,
+			queryFunc,
+			\`${id}\`,
+			{ initialData: profile(__db.query, \`${duckdbQueries[id].replaceAll('`', '\\`')}\`) }
+		);
+		/** @type {QueryStore} */
+		let ${id};
+		$: ${id} = $_${id};
+		`)
+
+
 		queryDeclarations += `
-            import debounce from 'debounce';
-            import { browser } from '$app/environment';
-			import {profile} from '@evidence-dev/component-utilities/profile';
-			
-			// partially bypasses weird reactivity stuff with \`select\` elements
-			function data_update(data) {
-				${valid_ids.map((id) => `${id} = data.${id} ?? [];`).join('\n')}
-			}
-
-			$: data_update(data);
-
-
-            ${valid_ids
-							.map(
-								(id) => `
-                let ${id} = data.${id} ?? [];
-                const _query_${id} = browser
-					  ? debounce((query) => profile(__db.query, query).then((value) => ${id} = value), 200)
-					  : (query) => (${id} = profile(__db.query, query, "${id}"));
-                $: _query_${id}(\`${duckdbQueries[id].replaceAll('`', '\\`')}\`);
-            `
-							)
-							.join('\n')}
-        `;
+		import {browser} from "$app/environment";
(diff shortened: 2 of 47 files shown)


packages/query-store/src/utils/buildId.ts
```diff
@@ -0,0 +1,10 @@
+const simpleHash = (str: string): string => {
```
Winterhart, line 1: Why not using an established `hash` library?

packages/query-store/src/mutations/agg.ts
```diff
@@ -0,0 +1,30 @@
...
+	}
+	if ('avg' in aggConfig) {
+		for (const column of forceArray(aggConfig.avg)) {
+			const alias =
+				typeof column === 'object' && 'alias' in column ? column.alias : `avg_${column}`;
+			q.select({ [alias]: avg(column) });
+		}
+	}
+
+	console.log(q.toString());
```
Winterhart, line 27: maybe ✂️

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+			{
+				get: (self, _prop) => {
+					// Intercept numeric indices. This implies we're trying to access rows (data) in the store.
+					// If the data has not been loaded, initiate the async #update method to fetch the data.
+					let prop: string | symbol | number = _prop;
+					if (typeof prop === 'string' && /^[\d.]+$/.exec(prop)) prop = parseInt(prop);
+					if (typeof prop === 'number') {
+						if (!self.#loaded) {
+							try {
+								const r = self.#fetchData();
```
Winterhart, line 130: is `fetchData` `async`? and what is `r`

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+
+		// TODO: Strip any trailing ; from queries
+		// This is hard because of comments
+		// We might want to just error out if the querystring contains a ; for simplicity
+
+		if (typeof query === 'string') {
+			this.#query.from({ __userQuery: sql`(${query})` }).select('*');
+		} else this.#query = query;
+		// @ts-expect-error Passing undocumented parameter
+		this.#exec = (...args: Parameters<Runner>) => exec(args[0], args[1] ?? this.id);
```
Winterhart, line 117: Maybe `ts` complains because `args[0]` is not checked if `length > 0`

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+		query: string | Query,
+		exec: Runner,
+		id?: string,
+		private readonly opts: QueryStoreOpts = { disableCache: false },
+		private readonly root?: QueryStore
+	) {
+		super();
+		Object.freeze(opts);
+		// Ensure an ID Exists
+		if (!id) this.id = buildId(query);
```
Winterhart, line 106:
```typescript

this.id = (id === undefined) ? buildId(query) : id

```

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+								if (r instanceof Promise)
+									r.catch((e) => {
+										throw new Error('Failed to update query store', { cause: e });
+									});
+							} catch (e) {
+								throw new Error('Failed to update query store', { cause: e });
+							}
+						}
+
+						if (!self.#values[prop]) return self.#mockResult;
```
Winterhart, line 140: Was that added for `testing`?

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+						return self.#values[prop];
+					}
+
+					// Intercept 'length' property. This implies we're trying to get the total number of rows (data) in the store.
+					// If the length has not been correctly updated, initiate the async #updateLength method to calculate it.
+					else if (prop === 'length') {
+						// TODO: Conditional
+						if (!self.#lengthLoaded) {
```
Winterhart, line 148: To move to the `else if` maybe?

```
prop === 'length' && !self.#lengthLoaded

File: packages/query-store/src/QueryStore.ts [hunk 8]
```diff
@@ -0,0 +1,340 @@
...
+									});
+							} catch (e) {
+								throw new Error('Failed to update query store length', { cause: e });
+							}
+						}
+					}
+
+					if (prop in self) {
+						// @ts-expect-error Typescript gets mad about this for some reason
+						if (typeof self[prop] === 'function') return self[prop].bind(self);
```
Winterhart, line 163: Maybe `ts` wants you to check that `self[prop` is not `undefined`

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+		}, this.#exec(queryWithComment));
+	};
+
+	#fetchMetadata = () => {
+		return handleMaybePromise((queryResult: QueryResult[]) => {
+			this.#columns = queryResult.map((row) => ({
+				name: row.column_name!.toString(),
+				type: row.column_type!.toString()
+			}));
+			this.#mockResult = Object.fromEntries(this.#columns.map((c) => [c.name, null]));
```
Winterhart, line 267: Is this only for `testing`?

packages/query-store/src/mutations/agg.ts
```diff
@@ -0,0 +1,30 @@
...
+type AggArgs = {
+	sum: MaybeAliased[] | MaybeAliased;
+	avg: MaybeAliased[] | MaybeAliased;
+};
+
+export const agg = (q: Query, aggConfig: AggArgs): Query => {
+	if ('sum' in aggConfig) {
+		for (const column of forceArray(aggConfig.sum)) {
+			const alias =
+				typeof column === 'object' && 'alias' in column ? column.alias : `sum_${column}`;
```
Winterhart, line 15: Maybe this need a comment/explanation

Winterhart (commented):
Very cool, to see the QueryStore upcoming! I saw smalls item...

- Some typescript `function/const` don't have a clear return type
- I am not sure sometimes there's `mock...` variable I don't know if it's for testing or a real variable
- I usually expect `fetch` to be `async`

packages/query-store/src/QueryStore.ts
ItsMeBrianD: `#mockResult` is an empty object that contains keys based on the rows returned from the metadata query; this prevents errors coming from `cannot read property "x" of undefined` when accessing `query[0].x` before the data has been fetched

packages/query-store/src/QueryStore.ts
ItsMeBrianD: The type error that was was that `prop` may not be a string, and typescript expects `self` to only be index accessed by numbers

packages/query-store/src/QueryStore.ts
ItsMeBrianD: See above comment

packages/query-store/src/utils/buildId.ts
ItsMeBrianD: Didn't seem worth adding a dependency for something simple like this. we can revisit in future though

packages/query-store/src/mutations/agg.ts
ItsMeBrianD, line 27:
```suggestion
```

packages/query-store/src/QueryStore.ts
ItsMeBrianD: `fetchData` _may_ be `async`, but not always (hence the promise check below). `r` is the result of the fetch

packages/query-store/src/QueryStore.ts
ItsMeBrianD: The `Runner` type only expects a single argument, we're providing an extra here for debugging.

We should be able to remove this though.

packages/query-store/src/QueryStore.ts
ItsMeBrianD: See also:

```typescript
	/**
	 * Svelte bases iteration on the `length` property; so that has to exist before it will try to pass a number
	 * However, we don't want to load everything if `length` is accessed - only the data itself.
	 * Therefore, there is a tick when the array is full of undefined, because we are still fetching.
	 * In that time period, we return this empty object so we don't get `cannot access x of undefined` errors.
	 * This may break in instances where the user may have nested arrays / dicts
	 */
	#mockResult: QueryResult = {};
```

packages/preprocess/src/process-queries.cjs
```diff
@@ -12,31 +12,56 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
+		// 					.join('\n')}
+		// `;
+
+		const queryStores = valid_ids.map(
+			(id) => `
+		const _${id} = new QueryStore(
+			\`${duckdbQueries[id].replaceAll('`', '\\`')}\`,
+			queryFunc,
+			\`${id}\`,
+			{ initialData: profile(__db.query, \`${duckdbQueries[id].replaceAll('`', '\\`')}\`) }
```
csjh, line 47: Shouldn't this be `data.${id} ?? profile(__db.query...)` to take advantage of query prerendering?

packages/preprocess/src/process-queries.cjs
```diff
@@ -12,31 +12,56 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
+		);
+		/** @type {QueryStore} */
```
csjh, line 49: Not sure if the jsdoc type matters in the preprocessor-inserted script

packages/query-store/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
ud3sh, line 1: what's the purpose of this package given that it's empty right now?
Is the idea to move the `QueryStore` code in evidence here down the road?

packages/query-store/src/types.ts
```diff
@@ -0,0 +1,22 @@
+import type { QueryStore } from './QueryStore';
+import type { Query } from '@uwdata/mosaic-sql';
+
+export type Subscriber<T> = (value: T) => unknown;
+
+export type MaybePromise<T> = Promise<T> | T;
+
+export type Runner = (query: string) => MaybePromise<QueryResult[]>;
```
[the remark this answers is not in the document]
ItsMeBrianD: It is the same type; defined it here so that it can more easily be mocked

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+	/** Text of the query represented by this store */
+	get text() {
+		// TODO: This needs a formatter
+		return this.#query.toString();
+	}
+
+	/** Name and Type information about the result columns */
+	#columns: ColumnMetadata[] = [];
+	get columns() {
+		return Array.from(this.#columns);
```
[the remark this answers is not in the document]
ItsMeBrianD: Yep; prevents the actual array from being modified. This may need to be a little more resilient because the objects themselves may be mutable. Could just freeze them though

packages/query-store/src/utils/buildId.ts
```diff
@@ -0,0 +1,10 @@
...
+	let hash = 0;
+	for (let i = 0; i < str.length; i++) {
+		const char = str.charCodeAt(i);
+		hash = (hash << 5) - hash + char;
+		hash &= hash; // Convert to 32bit integer
+	}
+	return new Uint32Array([hash])[0].toString(36);
+};
+export const buildId = (...args: any[]) => simpleHash(JSON.stringify(args));
```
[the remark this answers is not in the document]
ItsMeBrianD: Just makes things faster to compare shorter strings (and easier to read when debugging logs), doesn't make a huge difference here just a preference thing

packages/query-store/src/utils/buildId.ts
```diff
@@ -0,0 +1,10 @@
+const simpleHash = (str: string): string => {
+	let hash = 0;
+	for (let i = 0; i < str.length; i++) {
+		const char = str.charCodeAt(i);
+		hash = (hash << 5) - hash + char;
+		hash &= hash; // Convert to 32bit integer
+	}
+	return new Uint32Array([hash])[0].toString(36);
```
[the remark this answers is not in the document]
ItsMeBrianD: Shortens the output id making it easier to compare at a glance

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+	 * and hence is not recommended for normal use.
+	 */
+	readonly #proxied: QueryStoreValue;
+	get proxy() {
+		return this.#proxied;
+	}
+
+	/** Text of the query represented by this store */
+	get text() {
+		// TODO: This needs a formatter
```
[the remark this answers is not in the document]
ud3sh: Typically, I'd expect the formatting would be done in the UI itself rather than in the model

packages/query-store/src/utils/forceArray.ts
```diff
@@ -0,0 +1,3 @@
+export const forceArray = <T>(maybeArray: T | T[]): T[] => {
```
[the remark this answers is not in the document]
ItsMeBrianD: This sets us up to not end up with a massive file (applies to the mutations stuff as well)

packages/query-store/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"keywords": [],
+	"author": "The Evidence Engineering Team",
+	"license": "MIT",
+	"devDependencies": {
+		"typescript": "^5.2.2",
+		"vitest": "^0.34.1"
+	},
+	"dependencies": {
+		"@evidence-dev/universal-sql": "workspace:",
+		"@sqltools/formatter": "^1.2.5",
```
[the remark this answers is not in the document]
ItsMeBrianD: `nanoid` probably needs to get ripped out;
This package is included because it is used when displaying the queries on the page; there is an issue with it though when trying to use it in an evidence project; so I need to hammer that down at some point.

packages/query-store/src/QueryStore.ts
```diff
@@ -0,0 +1,340 @@
...
+
+				return cachedQuery.#proxied;
+			}
+			// Construct a new store, subscribe and cache it
+			const newStore = new QueryStore(
+				newQuery,
+				this.#exec,
+				id,
+				{
+					...this.opts,
```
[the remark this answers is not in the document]
ItsMeBrianD: For modifications like `where`, `orderby`, `limit` and `offset`; the columns don't change, so passing some initial data can help keep _something_ on the DOM / in the result set to minimize the amount of time the user experiences loading.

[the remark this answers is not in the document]
ud3sh: So is it just for the column headers ? It seems like we are not actually applying these desired aggregation directly on the data set returned by `parent` at any point?

sites/example-project/src/pages/query-store/+page.md
```diff
@@ -0,0 +1,9 @@
+```test_query
+SELECT * FROM marketing_spend
+```
+
+{test_query instanceof QueryStore}
+
+<DataTable data={test_query}/>
+<DataTable data={test_query.groupBy("marketing_channel").agg({ avg: ["spend"], sum: ["spend"]})}/>
```
ud3sh, line 8: Good to see an example of how aggregations work!

One hand it seems pretty cool.  

On the other hand, as I interpret it, it seems like we are providing a custom Evidence DSL to get around writing aggregation directly in SQL and avoid repeating re-writing the query query ... looking at the implementation of `#withStoreCache` it seems like we are creating a new query store that simply has a new SQL query derived from the original.

So questions (only applicable if I am interpreting this right):
1. Are these aggregation JS functions intended to support a particular feature / request ? 
    * I don't think this is part of the added functionality mentioned in the description, and I am guessing this is to make it easier for the user to reuse an existing query without having to copy/paste a new query and type up a new aggregation.
3. When these queries fail at runtime, would it be easy for the user to figure out JS aggregations are bad?
   * e.g suppose the original query was  `test_query='SELECT * from marketing_spend ORDER BY spend DESC LIMIT 10'`, and they write up their <DataTable> exactly as on line 8, then I think we'd end up with a where clause after a limit, and this would be due to a bad JS aggregation.

ItsMeBrianD: > looking at the implementation of #withStoreCache it seems like we are creating a new query store that simply has a new SQL query derived from the original.
This is exactly correct; the reason we do it this way (instead of mutating the query) is to ensure that you can re-use in a few places, e.g. if you paginate a query from your markdown for the datatable, we don't want it to be paginated on charts.

> I don't think this is part of the added functionality mentioned in the description, and I am guessing this is to make it easier for the user to reuse an existing query without having to copy/paste a new query and type up a new aggregation.

It is; but not for project authors

> Are these aggregation JS functions intended to support a particular feature / request ?

These are provided for _component_ development; "end users" (e.g. project authors) shouldn't have to touch or know about this API at all.

> When these queries fail at runtime, would it be easy for the user to figure out JS aggregations are bad?

The example you've point out here is a good one, I think there has to be some level of assumption that users are familiar with SQL (Especially if we format & print the query that is being run). Would be curious to hear thoughts on this

ud3sh: Ok, I didn't catch this was for component dev - starting to make more sense to me!

In that case, I'd think the specificities about error is less important as it would be up the component developer to ensure the aggregations are correct. Right?

packages/query-store/src/QueryStore.ts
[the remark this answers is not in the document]
ItsMeBrianD: > So is it just for the column headers ?

Right now; yes, this might be worth rethinking a bit in the future to instead pass column-metadata so `#mockResult` can be used

packages/query-store/src/QueryStore.ts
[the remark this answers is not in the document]
ItsMeBrianD: Hopefully; as a chunk of this in the future it will try to clean up the querybuilder syntaxes (e.g. the user's original query gets wrapped as a CTE or Subquery)

Formatting just makes this ready-to-use on the page, which we'd like to expose for component consumers

packages/core-components/src/lib/unsorted/viz/BigValue.svelte
```diff
@@ -142,6 +143,8 @@
 	{/if}
 </div>
 
+<pre>{JSON.stringify(data[0][0].duration)}</pre>
```
csjh, line 146: Was this intentionally left in?

packages/preprocess/src/process-queries.cjs
```diff
@@ -12,31 +12,31 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
-			// partially bypasses weird reactivity stuff with \`select\` elements
-			function data_update(data) {
-				${valid_ids.map((id) => `${id} = data.${id} ?? [];`).join('\n')}
-			}
-
-			$: data_update(data);
 
+		const queryStores = valid_ids.map(
+			(id) => `
+		const _${id} = new QueryStore(
```
csjh, line 18:
```suggestion
		$: _${id} = new QueryStore(
```
It's possible that 
```js
\`${duckdbQueries[id].replaceAll('`', '\\`')}\`
```
has interpolated variables in it, so this has to be reactive to that

PR #1222, opened 2023-09-29
Universal SQL: QueryStores keep error state

ItsMeBrianD:
### Description

When QueryStores encounter an error running (any) query they enter an error state; because the queries are immutable, the query store basically locks up and does nothing, when a new query string is provided, a new query store is created, which causes the page to start executing again.

This can most easily be tested using the Query console in the `test-env` project

### Checklist

### DIFF
--- a/packages/component-utilities/src/inferColumnTypes.js
+++ b/packages/component-utilities/src/inferColumnTypes.js
@@ -50,11 +50,7 @@ export const inferValueType = function (columnValue) {
 
 export default function inferColumnTypes(rows) {
 	if (rows instanceof QueryStore) {
-		return rows.columns.map((ct) => ({
-			name: ct.name,
-			evidenceType: 'string', // TODO: Do we need to translate ddb types to evidence types more effectively?
-			typeFidelity: TypeFidelity.INFERRED
-		}));
+		return rows._evidenceColumnTypes;
 	}
 	if (rows && rows.length > 0) {
 		let columns = Object.keys(rows[0]);
--- a/packages/component-utilities/src/inferColumnTypes.js
+++ b/packages/component-utilities/src/inferColumnTypes.js
@@ -1,5 +1,7 @@
 // To-do, replace with import from db-commons
 
+import { QueryStore } from "@evidence-dev/query-store";
+
 var EvidenceType;
 (function (EvidenceType) {
 	EvidenceType['BOOLEAN'] = 'boolean';
@@ -47,6 +49,13 @@ export const inferValueType = function (columnValue) {
 };
 
 export default function inferColumnTypes(rows) {
+	if (rows instanceof QueryStore) {
+		return rows.columns.map((ct) => ({
+			name: ct.name,
+			evidenceType: "string", // TODO: Do we need to translate ddb types to evidence types more effectively?
+			typeFidelity: TypeFidelity.INFERRED
+		}))
+	}
 	if (rows && rows.length > 0) {
 		let columns = Object.keys(rows[0]);
 		let columnTypes = columns?.map((column) => {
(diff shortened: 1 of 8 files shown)


packages/component-utilities/src/inferColumnTypes.js
```diff
@@ -47,6 +49,13 @@ export const inferValueType = function (columnValue) {
 };
 
 export default function inferColumnTypes(rows) {
+	if (rows instanceof QueryStore) {
+		return rows.columns.map((ct) => ({
+			name: ct.name,
+			evidenceType: 'string', // TODO: Do we need to translate ddb types to evidence types more effectively?
+			typeFidelity: TypeFidelity.INFERRED
+		}));
+	}
```
csjh, line 58:
```suggestion
		return rows._evidenceColumnTypes;
	}
```

packages/query-store/src/QueryStore.ts
```diff
@@ -55,6 +56,14 @@ export class QueryStore extends AbstractStore<QueryStoreValue> {
 		return Array.from(this.#columns);
 	}
 
+	get _evidenceColumnTypes() {
+		return Array.from(this.#columns).map((ct) => ({
+			name: ct.name,
+			type: 'STRING',
+			typeFidelity: 'inferred'
+		}));
```
csjh, line 64:
```suggestion
		return Array.from(this.#values._evidenceColumnTypes);
```

packages/query-store/src/QueryStore.ts
```diff
@@ -93,6 +102,32 @@ export class QueryStore extends AbstractStore<QueryStoreValue> {
...
+			// This needs some testing; might cause a race, but we want to prevent the query from loading anything at this point.
+			this.#length = 0;
+			this.#lengthLoaded = false;
+			this.#lengthLoading = false;
+		}
+
+		if (this.opts.errorNotifier) this.opts.errorNotifier(this.error!);
+
+		this.publish();
+	};
```
csjh, line 121:
```suggestion
	#setError = (err: Error | unknown): void => {
		console.debug(`QueryStore encountered a non-fatal error`, err);
		this.#error = err;

		if (err) {
			// This needs some testing; might cause a race, but we want to prevent the query from loading anything at this point.
			this.#length = 0;
			this.#lengthLoaded = false;
			this.#lengthLoading = false;
		}

		if (this.opts.errorNotifier) this.opts.errorNotifier(this.error!);

		this.publish();
	};
```

packages/core-components/src/lib/unsorted/viz/Chart.svelte
```diff
@@ -191,6 +192,8 @@
 	let columnSummaryArray;
 	let dateCols;
 
+	$: if (data.error) error = data.error.message;
```
csjh, line 195: Maybe switch to a nullish coalesce? This + the `export let data = undefined` is a bit unnerving (even though realistically you wouldn't use a chart w/o `data` supplied

csjh: Also, should this be outside of the big try/catch block? `error` is reset at the start of that, and it's rerun every time `data` is changed, so it might cancel this assignment out

packages/core-components/src/lib/unsorted/viz/DataTable.svelte
```diff
@@ -283,6 +286,8 @@
 					$props.columns.map((d) => d.id)
 			  )
 			: data;
+
+	$: if (data.error) error = data.error.message;
```
csjh, line 290: Same as above

csjh (approved):

packages/core-components/src/lib/unsorted/viz/Chart.svelte
ItsMeBrianD: > Also, should this be outside of the big try/catch block? error is reset at the start of that, and it's rerun every time data is changed, so it might cancel this assignment out

I was able to get errors from deeper in the chart and from the querystore to appear properly so I think we're good there

PR #1236, opened 2023-10-03
Add secondary y-axis for line charts

hughess:
### Description
See #874 for initial draft of this approach.

```html
<LineChart
   data={orders}
   x=month
   y=sales_usd0k
   y2=num_orders_num0
   yAxisTitle=true
   y2AxisTitle=true
/>
```

image

This PR adds a secondary y-axis called `y2` to the LineChart component. This will be extended to other chart types in the future.

### Improvements Since Draft PR #874

- **Axis Colours:** by default when using `y2`, the y-axis will be coloured the same as the first series, with the y2-axis getting coloured the same as the first series on the secondary axis. This behaviour can be turned off for either axis using the `yAxisColor` and `y2AxisColor` props. A specific colour can also be passed to these props to set it to a custom colour (E.g., `yAxisColor=red`)
- **Tooltip formatting:** now contains logic to determine which y-axis a datapoint should be on, and applies the appropriate format for that axis
- **Axis Overrides:** a hidden secondary y-axis is included in all charts by default, and is turned on within child chart components (`Line.svelte` in this PR's case)
- **Axis Formatting:** the secondary axis has a `y2Fmt` prop to override the number format

### Remaining Problems

- Usage with multiple chart types
- Axis colours when custom colours chosen
- Scatter and Bubble tooltips will need special attention when implementing for those chart types

### Open Questions

- Should the axis titles for `y` and `y2` be turned on by default when `y2` is used?

### General Improvements to Library

- Adds a `totalSeriesCount` within Chart.svelte, which makes it easier to determine if a chart is multi-series or not
- Standardizes the colour imports used throughout the library into `chartColours` (colours to use for chart series) and `uiColours` (colours for UI elements like axis labels)
   - Previously, the colours were defined in multiple places 

### Checklist


archiewood:
A test case we should double check against here:

y=number of somethings
y2=percent of somethings 

eg number of emails sent, % open rate

hughess:
@archiewood here's an example with $ and %:

CleanShot 2023-10-13 at 11 06 35@2x

And with axis titles:
CleanShot 2023-10-13 at 11 07 14@2x

Seems like it would be a lot better to have a bar in here for one of the series

hughess:
@archiewood would you turn on the axis titles by default when the secondary axis is used, or leave off by default?

archiewood:
On by default I think, and I would turn off the legend by default if that was possible

hughess:
@archiewood thanks - added the default on for axis titles. The legend logic is a bit too complicated to do at the moment

PR #1241, opened 2023-10-06
support URL parameters in getStatusAndExtractQueries

brandonros:
This is kind of a "big security vulnerability" so not sure if we even feel comfortable making it `dev mode only`.

Not sure if it could be done even better to support proper parameterization or what we'd consider "good enough".

It enables you to write queries like this in Markdown on pages:

````
```sql query1
select
    COUNT(*)
from quote_snapshots
WHERE scraped_at >= '{{expiration_date}}T13:30:00' AND scraped_at <= '{{expiration_date}}T20:00:00'
```
````

image

Then you visit it in the browser at `http://localhost:10000/test/2023-10-03/`, etc.

I think this addresses some part of https://github.com/evidence-dev/evidence/issues/728

### DIFF
--- a/sites/example-project/src/pages/api/status/[...route]/+server.js
+++ b/sites/example-project/src/pages/api/status/[...route]/+server.js
@@ -1,9 +1,8 @@
 import { getStatusAndExtractQueries } from './extractQueries.server.js';
 import { json } from '@sveltejs/kit';
 
-export async function GET({ params }) {
-	const { route } = params;
-	const status = getStatusAndExtractQueries('/' + route);
+export async function GET(event) {
+	const status = getStatusAndExtractQueries(event);
 	if (status) {
 		return json({ status });
 	}
--- a/sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js
+++ b/sites/example-project/src/pages/api/status/[...route]/extractQueries.server.js
@@ -43,12 +43,38 @@ const updateDirectoriesandStatus = function (queries, routeHash) {
 	return status;
 };
 
-export const getStatusAndExtractQueries = function (route) {
+const extractUrlParametersFromReferer = (route, referer) => {
+	const routeSegments = route.split('/');
+	const refererPath = new URL(referer).pathname;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by csjh** [comment]
Hi, this is going to be added in #993, with syntax along the lines of:
```sql
```sql query1
select
    COUNT(*)
from quote_snapshots
WHERE scraped_at >= '${$page.params.expiration_date}T13:30:00' AND scraped_at <= '${$page.params.expiration_date}T20:00:00'
`` `
```
(maybe just `${params.expiration_date}` but same syntax either way)

mcrascal:
@brandonros this is pretty darn interesting, thanks for sharing.

Take a look at #993 and let us know what you think. I think proposal here is something we would like to come back to _after_ #993. 

- In what's here, you could pass params down to your source DB, which would be particularly useful when you have extremely large underlying queries that you want to parameterize. This would be great
- My understanding of what's here is that you could only pass params that were known at build time, so end-user inputs (date pickers etc.) would not be possible, which is also something we are very motivated to support

brandonros:
> @brandonros this is pretty darn interesting, thanks for sharing.
> 
> Take a look at #993 and let us know what you think. I think proposal here is something we would like to come back to _after_ #993.
> 
> * In what's here, you could pass params down to your source DB, which would be particularly useful when you have extremely large underlying queries that you want to parameterize. This would be great
> * My understanding of what's here is that you could only pass params that were known at build time, so end-user inputs (date pickers etc.) would not be possible, which is also something we are very motivated to support

Coming from `Express` I think a lot of us are familiar with things like `GET /users/:userId`, just refactoring that to match the basically "requirement" due to `:` not playing nice with filesystem directory names of `[userId]` made a lot of sense.

I *also* toyed with `?userId=1234` but the way the evidence + svelte template + svelte-kit page server side renders then loads the queries through like `/api/status/page-info`, it didn't get passed along very well.

I am beyond fine to close this. Glad to see it is being addressed in any capacity and appreciate everything the entire team is doing here. Thank you for making such an awesome product.

Issue #1186, opened 2023-09-19
Universal SQL: Sources settings UI supports multiple data sources 


mcrascal:
- Everything after installing a source adapter can be executed in the UI 
- Create a new source -> creates the directory, required yaml files, presents form to populate required fields 
- Source adapters export the information needed to create the forms. Branching form logic supported in connection interface (see example below). 
- Forms support N connections from N adapters (e.g. two postgres connections and a snowflake connection) 
- Test, save, edit supported 
- Unit tests 

**Example interface** 

```javascript
module.exports.options = {
 host: { type: "string", secret: false, description: "Database hostname to connect to" },
 authMethod: { type: "select", values: ["userpass", "made-up"], children: {
  "userpass": {/* more field */},
  "made-up": { /* more field */ }
 }}
}
```

PR #1248, opened 2023-10-10
Universal SQL: Deployment menu uses new environment variables

ItsMeBrianD:
### Description

This converts the deployment panel to use the environment variable structure of usql, rather than the predefined variables that were there previously.

This is dynamic, based on what is provided to the datasource, (e.g. in the user's own environment, `connection.options.yaml`, and `connections.yaml`), so while it may not be 100% correct _all the time_, it will exactly match the user's local environment.

Once we complete #1186 this will not be as much of an issue; as the UI for configuration _will_ use options defined by the sources themselves.

As a potential addition to this PR, we may want to consider a "copy all" button, or file an issue for having on in the future; as copy-pasting for more than one or two connectors would probably be long and tedious.

### Checklist

image

### DIFF
--- a/packages/plugin-connector/src/data-sources/build-connector.js
+++ b/packages/plugin-connector/src/data-sources/build-connector.js
@@ -9,7 +9,7 @@ import { DatabaseConnectorSchema } from './schemas/query-runner.schema';
 export const buildConnector = async (packageMain, supports) => {
 	// https://github.com/nodejs/node/issues/31710 thanks windows
 	const crossPlatformPackage = new URL(`file:///${packageMain}`).href;
-	const connectorPackage = await import(crossPlatformPackage);
+	const connectorPackage = await import(crossPlatformPackage /* @vite-ignore */);
 	const connector = DatabaseConnectorSchema.parse({ ...connectorPackage, supports });
 
 	return connector;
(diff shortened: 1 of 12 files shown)


packages/plugin-connector/src/data-sources/build-connector.js
```diff
@@ -9,7 +9,7 @@ import { DatabaseConnectorSchema } from './schemas/query-runner.schema';
 export const buildConnector = async (packageMain, supports) => {
 	// https://github.com/nodejs/node/issues/31710 thanks windows
 	const crossPlatformPackage = new URL(`file:///${packageMain}`).href;
-	const connectorPackage = await import(crossPlatformPackage);
```
csjh, line 12: What changed to make this necessary?

csjh (approved):

packages/plugin-connector/src/data-sources/build-connector.js
ItsMeBrianD: Vite complains in a few cases when importing the plugin connector because it can't optimize the dynamic dynamic import; so the ignore comment suppresses that warning

sites/example-project/src/pages/settings/+page.server.js
```diff
@@ -0,0 +1,13 @@
+import { dev } from '$app/environment';
+
+export const load = async () => {
+	if (dev) {
+		const { getDatasourceOptions } = await import('@evidence-dev/plugin-connector');
```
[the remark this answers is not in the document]
ItsMeBrianD: Keeps it from appearing in the built bundle; also mirrors the other load for the settings page

PR #1249, opened 2023-10-10
Use `-` instead of `{column.linkLabel}` for nullish linkLabel values

csjh:
### Description

Before:
image

After:
image

### Checklist

### DIFF
--- a/.changeset/thin-pens-care.md
+++ b/.changeset/thin-pens-care.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


mcrascal:
@csjh looks like this introduced a regression. 

You can see the case in new tables in dev workspace. Previously where the text did not map to a column name, we'd display the text, so you could label it something arbitrary like "more ->" now, this is viewing those as nullish (no matching column) and swapping in `-`. 

I think the cascade you want here is:
1. matches a column name -> current behaviour as of this PR 
2. doesn't match a column name -> old behaviour (drop the string into the column) 

going to re-open.

Issue #1244, opened 2023-10-10
Add Example Query for Sankey Diagram


nattaylor:
## Request Type

- [ ] Small Fix (wording, errors, typos, etc.)
- [x] Explanation Improvement (make a description clearer for the reader of the docs)
- [ ] Walkthrough Request (new tutorial for a particular feature or use case)
- [ ] Other

## Description

I struggled to implement a sankey diagram.

First I just tried dumping data in, which resulted in this.

<img width="595" alt="image" src="https://github.com/evidence-dev/evidence/assets/2789593/17850382-19f5-4330-b2f2-72d6b439d435">

With a limit, it got better.

Then I searched and came upon the examples in [sites/example-project/src/pages/diagrams/sankey.md](sites/example-project/src/pages/diagrams/sankey.md) and see the the required `UNION ALL` syntax

So I think some better docs would go a long way.

Maybe just a few simple tips:
> The component works with up to about 20 rows and one-level of links
> To add additional levels of links, you must implement each level as a query and then UNION ALL the queries together.  See example at sites/example-project/src/pages/diagrams/sankey.md

## If improving on existing content, where is that content located?

https://docs.evidence.dev/components/sankey-diagram

## Are you interested in contributing this improvement to the docs?

Yes, I will open a PR I think although I don't have a dev env set up yet.

Feel free to open a pull request with your suggested changes to the docs! If you're interested in contributing, but you need help, please let us know here or in Slack and we'll be happy to help you get started.

archiewood:
Thanks for pointing out the lack of clarity in the docs

The union all syntax is not required.

Format wise you just need to end up with three columns, a "source Name" column, a "target name" column and a value column

archiewood:
Oh I see what you are getting at.

Is there an easier way you'd prefer to structure your data in the case of multi level sankey without union all ing them together?

I guess we could accept a table like:

SourceCol | level1Col | targetCol | value
-- | -- | -- | --

nattaylor:
For my use case that would great!  ...and it would definitely be pretty neat (and fit great with SQL) if it could just go pairwise from left to right through N-columns to do that  ...but it does seem kind of like a hateful-to-implement special purpose thing that maybe should belong in a plugin.

Maybe its not that bad, but you have to iterate over each pair and calculate the filtered intermediate sums.

That being said, I have a tiny bunny brain and maybe its not as hard as I'm making it.

[...]

PR #1254, opened 2023-10-12
add details to sankey docs

nattaylor:
### Description

Adds more details to SankeyDiagram component to address #1244

NOTE: the nested code blocks are... eh because there's an invisible charcter before the middle set of ``` to prevent parsing as closing block

image


sites/docs/docs/components/sankey-diagram.md
```diff
@@ -4,9 +4,14 @@ title: Sankey Diagram
 hide_table_of_contents: false
 ---
 
+The SankeyDiagram component accepts a query and displays a flow from one set of values to another.
+
+To display a flow with multiple levels, like these examples, see [Mutl-level](#multi-level) below.
```
archiewood, line 9:
```suggestion
To display a flow with multiple levels, like these examples, see [Mutli-level](#multi-level) below.
```

sites/docs/docs/components/sankey-diagram.md
```diff
@@ -29,6 +34,52 @@ hide_table_of_contents: false
...
+select 'pinterest' as source, 'all_traffic' as target, 25 as count
+union all
+select 'all_traffic' as source, '/' as target, 50 as count
+union all
+select 'all_traffic' as source, '/docs' as target, 150 as count
+union all
+select 'all_traffic' as source, '/blog' as target, 25 as count
+union all
+select 'all_traffic' as source, '/about' as target, 75 as count
+‍```
```
archiewood, line 71: I suspect this example is a bit confusing because it uses many union statements when in reality if you have an actual DB connection rather than generating fake data with SQL, you only need to to one union per extra level in the sankey:

Perhaps this is a better example?

```sql
select 
    channel as source,
    'all_traffic' as target
    count(user_id) as count
from web_events
group by 1,2

union all

select 
    'all_traffic' as source,
    page_route as target
    count(user_id) as count
from web_events
group by 1, 2
```

archiewood (changes requested):
Thanks for the PR! 
One typo and a comment

archiewood (approved):

archiewood:
Thanks @nattaylor !

PR #1259, opened 2023-10-12
Handle empty code block language better

csjh:
### Description

Prints an additional error to the console when a code block has an empty language, and displays the query on the page as a regular markdown code block
image
image

Alternatives:
- Make it a full-blown page error:
image
Forces action, more obvious that something is wrong

- Display the error in `QueryViewer`
image
More consistent with other errors, but weirder if you intended a normal markdown code block

### Checklist


mcrascal:
@csjh there is an [existing path for handling untitled queries](https://github.com/search?q=repo%3Aevidence-dev%2Fevidence%20untitled&type=code), and having the issue treated as if the query was literally titled untitled, and was throwing a SQL error. Think it would be possible to resurrect that instead?

csjh:
> Think it would be possible to resurrect that instead?

Yeah that should be fine, seems like behavior w/ that path is to just have an `untitled error` toast notification + an error in console.

mcrascal:
I believe we'd also show the code block as if it was titled untitled and the sql expression was returning an error.

ItsMeBrianD (approved):
This looks good to me, is it waiting for anything else?

PR #1260, opened 2023-10-13
Adjust chart print width to fix Chrome print issue

hughess:
### Description
Noticed that charts in Chrome are getting cut off when I try to export to PDF.

@mcrascal you might already know what's going on here - let me know what you think

chrome-print-old

But the same issue doesn't happen in Safari:

safari-print-old

I think there are 2 options here, but I don't totally understand why some things are working and others aren't, so maybe there are better ways:

### 1. Bring print width down from 645px to 606px
I took a look at the print width which was set to `645px` and adjusted it until it would fit on my page in Chrome. `606px` was the max I could do without it overflowing the margins of the page.

#### Chrome
chrome-print-606

#### Safari
safari-print-606

### 2. Use 100% as the chart width
This works for bringing the charts onto the page fully, but it causes the spacing above/below the charts to be bigger than it appears on the webpage, which isn't great.

#### Chrome
chrome-print-100p

#### Safari
safari-print-100p

### Checklist

### DIFF
--- a/.changeset/nervous-keys-yell.md
+++ b/.changeset/nervous-keys-yell.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Update print settings
--- a/packages/core-components/src/lib/unsorted/ui/QueryToast.svelte
+++ b/packages/core-components/src/lib/unsorted/ui/QueryToast.svelte
@@ -56,4 +56,10 @@
 	span.status {
 		@apply font-medium;
 	}
+
+	@media print {
+		#toast {
+			display: none;
+		}
+	}
 </style>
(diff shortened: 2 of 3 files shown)


mcrascal:
Ah good shout, looking at some other examples, I might actually go smaller than 606? 

Intuitively, 100% makes sense but it results in different print outputs depending on the size of the browser window when the print event is fired, with small browser windows resulting in vertical clipping.

hughess:
What size would you do?

mcrascal:
590 looks like it works -- It just looks like the horizontal bars are getting clipped in the preview here.

hughess:
Just made the change, thanks.

I also noticed the `print:hidden` tailwind class wasn't working on query toasts for some reason, so I've added a media query to hide them on print

PR #1272, opened 2023-10-16
Docs: note that Snowflake columns will be lowercase

DArtagan:
### Description

Documentation update: in the Snowflake section of Data Sources, add a note that all column names will be converted to lowercase in Evidence.  This is significant because Snowflake has the (weird) convention of all table and column names being being uppercase.

It looks like this was design choice made when the Snowflake connector was written, with a comment saying that an issue should be opened to consider this further:
https://github.com/evidence-dev/evidence/blob/cafd4b1f6002a25873949c6c1b62cb72cc4eacd7/packages/snowflake/index.cjs#L158C5-L158C5

### DIFF
--- a/sites/docs/docs/core-concepts/data-sources/index.md
+++ b/sites/docs/docs/core-concepts/data-sources/index.md
@@ -89,7 +89,7 @@ Now you can copy the access token and use it in your Evidence project.
 
 ### Snowflake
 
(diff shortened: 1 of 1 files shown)


DArtagan:
I imagine you folks don't want a changeset added for a documentation update?  Otherwise happy to do so.

archiewood:
TY - LGTM

PR #1276, opened 2023-10-17
Universal SQL: Add loading state to BigValue and Value

ItsMeBrianD:
### Checklist


packages/core-components/src/lib/unsorted/viz/Value.svelte
```diff
@@ -88,7 +89,9 @@
 	}
 </script>
 
-{#if placeholder}
+{#if data.loading}
+	<span class="placeholder">Loading...</span>
```
csjh, line 93: We might want to make this user configurable in the future

packages/plugin-connector/src/build-plugins/preprocess.js
```diff
@@ -7,16 +7,18 @@ import autoImport from 'sveltekit-autoimport';
...
+			/** @type {Record<string,string[]>} */
+			const packages = {};
+			for (const [component, data] of Object.entries(components)) {
+				if (!packages[data.package]) packages[data.package] = [];
+				const import_name = data.aliasOf ? `${data.aliasOf} as ${component}` : component;
+				packages[data.package].push(import_name);
+			}
+			return packages;
+		})
+		.catch(() => ({}));
```
csjh, line 21: What was the error coming from here? Is it safe to handle quietly?

csjh (approved):
LGTM

packages/plugin-connector/src/build-plugins/preprocess.js
ItsMeBrianD: This is happening when there is no evidence.plugins.yaml, but it is only a problem in the monorepo; there are other places that catch this error when in actual use. We want to prevent the svelte language server from crashing when it is missing.

Issue #1188, opened 2023-09-19
Universal SQL: Build-time performance guardrails


mcrascal:
- Warn on parquet file size and row count landing in sources 
- Simple heuristics ok, no need to account for consumption in the client (e.g. this file is only going to ever be sent in small partitions)

PR #1281, opened 2023-10-19
Client and build guardrails

csjh:
### Description

Adds a toast notification when large queries (10MB estimated RAM usage) are sent to the page
image

Also a warning notification when building large source parquet (100MB estimated RAM usage)
image

Can convert the ram estimate to a score as discussed in #1188 but RAM seems more concrete

### Checklist


csjh:
image
Also possible to add more warnings like so, but IMO it's a bit repetitive, because they'll usually all be triggered

ItsMeBrianD (commented):
~~I think it would be best to have a predefined set of toast statuses; rather than letting component add their own. Consistency can deteriorate much more easily with this approach~~

ItsMeBrianD (commented):
Should be tailwind-ified

packages/core-components/src/lib/unsorted/ui/ToastWrapper.svelte
```diff
@@ -0,0 +1,21 @@
+<script>
+	import Toast from './Toast.svelte';
+	import { toasts } from '@evidence-dev/component-utilities/stores';
+</script>
+
+<div class="container">
```
ItsMeBrianD, line 6: Should be tailwind

packages/core-components/src/lib/unsorted/ui/Toast.svelte
```diff
@@ -0,0 +1,52 @@
...
+	in:scale
+	out:fly|local={{ x: 1000, duration: 1000, delay: 0, opacity: 0.8 }}
+>
+	<span class="title">{title}</span>
+	<span class="message">{message}</span>
+</div>
+
+<style>
+	.toast {
+		border-radius: 4px;
```
ItsMeBrianD, line 22: Should be tailwind

packages/core-components/src/lib/unsorted/QueryStatus.svelte
```diff
@@ -20,25 +19,37 @@
 				await invalidateAll();
 			}
 
-			statuses.push(data);
-			statuses = statuses;
+			const { status, id } = data;
+
+			const style =
```
ItsMeBrianD, line 24: I think it would be best to have a predefined set of toast statuses; rather than letting component add their own. Consistency can deteriorate much more easily with this approach

packages/query-store/src/QueryStore.ts
```diff
@@ -347,13 +353,24 @@ export class QueryStore extends AbstractStore<QueryStoreValue> {
...
 				this.#metaLoaded = true;
+				this.#calculateScore();
 				this.publish();
 			},
 			() => this.#exec(`--col-metadata\nDESCRIBE ${this.#query.toString()}`),
 			this.#setError
 		);
 	};
 
+	#calculateScore = () => {
```
ItsMeBrianD, line 364: This function feels a little odd to me; I don't expect something called `calculateScore ` to notify the user; might want to split this into 2 functions

1 should calculate and return the score; the other should be responsible for the notification 

We can also expose the score once it's split up

packages/universal-sql/src/calculateScore.js
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ *
+ * @param {string} column_type
+ * @returns {number}
+ */
+function columnTypeToScore(column_type) {
+	const NUMBER = 12; // 8 bytes in a double + 4 bytes heap overhead?
+	const BOOLEAN = 4; // booleans are stored as 4 byte integers
+	const STRING = 2 * 15; // assume 15 character string
```
ItsMeBrianD, line 57: Does the varchar size get exposed? That might be a little more reliable 

This works if not

ItsMeBrianD (changes requested):

packages/universal-sql/src/calculateScore.js
csjh: You mean something like the max sized string in the column? I don't think so

ItsMeBrianD (commented):
A few comments here; this also has a lot of noise - probably should be split into multiple PRs in instances like this.

packages/query-store/src/QueryStore.ts
```diff
@@ -13,6 +13,8 @@ import { Query, sql, count } from '@uwdata/mosaic-sql';
 import { buildId } from './utils/buildId.js';
 import { handleMaybePromise } from './utils/handleMaybePromise.js';
 import { mutations } from './mutations/index.js';
+// @ts-expect-error can't figure out how to resolve this
```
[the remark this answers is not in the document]
csjh: It couldn't figure out the `.d.ts` even though the export condition has it as `"types": "./src/calculateScore.d.ts"`

csjh:
Yeah, some parts from #1296 snuck in because they were affecting the toasts

packages/core-components/src/lib/unsorted/ui/Toast.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
+	import { scale, fly } from 'svelte/transition';
+
+	export let status = '';
+	export let title;
+	export let message;
+</script>
+
+<div
+	class="print:hidden rounded py-1 px-3 my-4 mx-0 shadow-md text-xs font-[var(--monospace-font-family)] flex justify-between transition-all duration-300 border"
```
csjh, line 10:
```suggestion
	class="print:hidden rounded py-1 px-3 my-4 mx-0 shadow-md text-xs font-mono flex justify-between transition-all duration-300 border"
```

packages/query-store/src/QueryStore.ts
```diff
@@ -324,6 +337,8 @@ export class QueryStore extends AbstractStore<QueryStoreValue> {
 				this.#lengthLoaded = true;
 				this.#lengthLoading = false;
 
+				this.#calculateScore();
```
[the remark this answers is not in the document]
csjh: It has to be called anywhere that length or results are fetched, so I think there has to be at least 5 occurrences (async/sync length/initialData, + data).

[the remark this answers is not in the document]
ItsMeBrianD: But could it be called in the data / length fetching functions?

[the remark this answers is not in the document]
csjh: How do you mean? It's already in `#fetchLength` and `#handleInitialData`. Though I'm looking now and it should probably *also* go in `#fetchData`. Is there a reason why `this.#lengthLoaded` isn't set in `#fetchData`? I think that was my heuristic when I was deciding where to call calculateScore

packages/plugin-connector/src/data-sources/exec-source.js
```diff
@@ -51,8 +51,7 @@ export const execSource = async (source, supportedDbs, outDir) => {
 				});
 			} else result = _r;
 		} catch (e) {
-			if (e instanceof z.ZodError)
-				console.log(/**@type {Error & { format: () => any}}*/ (e).format());
+			if (e instanceof z.ZodError) console.log(e.format());
```
ItsMeBrianD, line 54: We might want to look at `clean-zod-errors.js` for this; but not a huge deal either way

ItsMeBrianD (approved):
:shipit:

PR #1284, opened 2023-10-21
og images for component pages

archiewood:
### Checklist


archiewood:
these work, but are the wrong aspect ratio.

Would be good if there was some way to resize them to 2:1 or regenerate

PR #1298, opened 2023-11-02
Fix multiple queries running on server start

archiewood:
This prevents queries from executing twice when pages load in development.

### Checklist

### DIFF
--- a/.changeset/curly-foxes-battle.md
+++ b/.changeset/curly-foxes-battle.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/components': patch
+---
+
+prevent multiple hits to the data API in the terminal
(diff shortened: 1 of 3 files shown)


archiewood:
@csjh - @hughess and I were looking at this together.

We weren't sure what this comment was about
> has to be cloned to bypass the proxy https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/page/load_data.js#L297

But this _seems_ to work and fix the issue.

csjh:
Ahh that makes sense, the whole point of the proxy is to avoid the two requests. However, it inlines the contents of the fetch into the page, which was causing the issue in #900. So removing it causes each templated page to have a full copy of the data.

To see the difference, add something like this to `example-project/src/pages/templated-pages/[category]/[item]/+page.md`

```orders
SELECT * FROM orders
```

`{orders.length}`

archiewood:
Can we have our cake and eat it here?
Only do the clone when we are `build`-ing

(i feel like this code is poorly written, but hopefully my intent is clear!)

csjh:
Yup, that makes sense to me.

archiewood:
Oh, your explanation is very clear. Makes more sense to me now!

PR #1307, opened 2023-11-06
Add LastRefreshed component

hughess:
### Description
Adds `LastRefreshed` component, which shows the time since the last build took place.

Can see the full date/time by hovering over the text.

```html
<LastRefreshed/>
```

CleanShot 2023-11-05 at 21 14 14@2x

CleanShot 2023-11-05 at 21 14 37@2x

#### Props
- `prefix`: allows you to set a custom prefix (e.g., "Updated" instead of the default "Last Refreshed")

```html
<LastRefreshed prefix="Updated"/>
```

CleanShot 2023-11-05 at 21 16 10@2x

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/ui/LastRefreshed.svelte
+++ b/packages/core-components/src/lib/unsorted/ui/LastRefreshed.svelte
@@ -3,39 +3,42 @@
 </script>
 
 <script>
-    import { version } from '$app/environment';
-    const timestamp = Number(version);
+	import { version } from '$app/environment';
+	const timestamp = Number(version);
 
-    export let prefix = 'Last refreshed';
+	export let prefix = 'Last refreshed';
 
-    function timeAgo(startTimestamp, endTimestamp) {
-        const secondsAgo = Math.floor((endTimestamp - startTimestamp) / 1000);
+	function timeAgo(startTimestamp, endTimestamp) {
+		const secondsAgo = Math.floor((endTimestamp - startTimestamp) / 1000);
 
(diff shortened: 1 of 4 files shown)


packages/core-components/src/lib/unsorted/ui/LastRefreshed.svelte
```diff
@@ -0,0 +1,44 @@
...
+			const yearsAgo = Math.round(secondsAgo / 31536000);
+			return yearsAgo === 1 ? '1 year ago' : `${yearsAgo} years ago`;
+		}
+	}
+
+	const varTimeAgo = timeAgo(timestamp, Date.now());
+</script>
+
+<p class="text-sm py-1">
+	<a href=" " class="cursor-text" title="{new Date(timestamp).toLocaleString()} "
```
csjh, line 41: Why is this a link?

hughess: Ah that was just so I could use the `title` prop to show the full date on hover. But it does result in some unexpected behaviour, like reloading the page when clicked

csjh: Sounds like it can be applied to any element based on the MDN page: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

hughess: Ah, so it can be haha. Thanks! I'll change this to remove the link

PR #1311, opened 2023-11-06
Fix hard refresh on initial project boot

csjh:
### Description

Adds dependencies that vite can't find (plugin-connector inserted `@evidence-dev/core-components` imports, imports added in preprocess)

### Checklist


mcrascal:
LGTM, I get deps optimized on the test-charts page, but I think it's because we're importing genSeries from component utilities, which I'd suspect isn't directly depended on by the component library.

csjh:
Yeah I think that one is an "intentional" Vite optimize deps execution

PR #1315, opened 2023-11-08
Explicitly set font family for chart theme

hughess:
### Description
On some operating systems, the font shown in charts is inconsistent with the UI font. This PR explicitly sets the font family within the echarts theme so the font is more consistent across operating systems.

In the future, we should consider changing the font family from `sans-serif` to `inter`, but on trying it here I noticed spacing issues that would need to be worked out.

#### Windows Before
windows-chartfont-before

#### Windows After
windows-chartfont-after

### Title colours
This also slightly darkens the default chart title and subtitle colour to fit better with the new project styling:

#### Before:
CleanShot 2023-11-07 at 21 37 24@2x

#### After:
CleanShot 2023-11-07 at 21 37 36@2x

Still might be a touch too light, but we can start here

### Checklist

### DIFF
--- a/.changeset/chatty-pugs-sleep.md
+++ b/.changeset/chatty-pugs-sleep.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/component-utilities': patch
+---
+
+Explicitly set font family for chart theme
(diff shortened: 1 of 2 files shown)


mcrascal:
LGTM

PR #1316, opened 2023-11-08
POC of an SSR Loading Skeleton for Charts 

mcrascal:
styling could be improved slightly

CleanShot 2023-11-07 at 22 40 09


hughess:
Added the chart height. A bit jumpy, but pretty close!

CleanShot 2023-11-08 at 11 34 58
CleanShot 2023-11-08 at 11 35 40

hughess:
Made some adjustments to get a bit closer to the actual height and margins. Not sure why they aren't identical, but seems close enough for now

archiewood:
@hughess I think it's because of the download bar spilling out of the chart container slightly?
CleanShot 2023-11-08 at 12 29 45

hughess:
@archiewood I thought I had accounted for that, but the pixels didn't seem to be adding up to what I expected. In any case, your update looks very close!

PR #1346, opened 2023-11-20
More informative errors for HMR errors

hughess:
### Description
Implements a new error overlay to replace the default Vite HMR error overlays.

Catalogs errors in `hmrErrorHandling.js` in `component-utilities`.

The idea here is to receive a Vite error message, evaluate the content it shares with us, then make an inference and supply the user with a more informative message.

This is a manual and inaccurate way to do this, but it's better than the messages our users currently receive from Vite.

### To do
- Tighten up some of the rules around “component was left open” - this is too eager a message at the moment
- Confirm that this approach doesn’t slow down HMR for users
- Check out what’s possible for logging - I don’t want to log the actual errors if we can avoid it…it’s a lot of text to the console and doesn’t contain much helpful info for a user (if any)
   - I’d like to log our custom errors to the console if we can

### Example
CleanShot 2024-02-23 at 09 52 20

### Checklist


hughess:
in the most recent commit, a new line is added to the svelte config. Initially I tried to add this to the config for the svelte vite plugin in vite.config as shown [here](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md) (which feels like a better place for it), but that didn't work.

hughess:
When I tried to trace the path that a warning takes in the svelte vite plugin, I noticed that `svelte-tiny-linked-charts` seems to generate a lot of work/noise for the preprocessor. I don't know for sure, but I suspect this has an impact on performance of the initial page load since it's duplicating a bunch of the workflows

archiewood:
We should deprecate this feature I think - we've talked about it before

sites/example-project/src/pages/+layout.svelte
```diff
@@ -75,13 +75,92 @@
...
+
+			// Apply basic styles for visibility
+			errorDisplay.style.position = 'fixed';
+			errorDisplay.style.top = '10px';
+			errorDisplay.style.left = '50%';
+			errorDisplay.style.transform = 'translateX(-50%)';
+			errorDisplay.style.background = 'var(--red-600)';
+			errorDisplay.style.color = 'white';
+			errorDisplay.style.padding = '10px';
+			errorDisplay.style.zIndex = '1000'; // Ensure it's above other elements
```
ItsMeBrianD, line 150: Use `classList`

sites/example-project/src/pages/+layout.svelte
```diff
@@ -75,13 +75,92 @@
...
+
 	const prefetchStrategy = dev ? 'tap' : 'hover';
 
 	let mobileSidebarOpen = false;
 
 	$: if ($navigating) {
 		mobileSidebarOpen = false;
 	}
+
+	if (import.meta.hot) {
```
ItsMeBrianD, line 87: This can likely be put into an `EvidenceErrorOverlay` component to remove it from the layout

mcrascal: @ItsMeBrianD see: #1364

ItsMeBrianD:
What's the story on this PR currently?

cc: @mcrascal @Winterhart

mcrascal:
@ItsMeBrianD I can take a look. It was paused during the USQL release, but I suspect it's good to go.

mcrascal:
@ItsMeBrianD It's not ready to go -- I'm getting `URL.split is not a function` when trying to boot it up.

hughess:
This one seems to be working on my end

packages/core-components/src/lib/unsorted/ui/ErrorOverlay.svelte
```diff
@@ -0,0 +1,77 @@
...
+				handleExpectedCharacterErrors,
+				handlePropErrors,
+				handleUnexpectedEnd,
+				handleUnexpectedTokens,
+				handleExpectedWhitespace
+			];
+
+			let customErrorMessage = errorHandlers.reduce((acc, handler) => {
+				return acc || handler(err);
+			}, null);
```
ItsMeBrianD, line 33: Are these messages always going to be in order?
e.g. could `handleExpectedWhitespace` override an error from `handleOpenTagErrors`?

If yes, we could consider making this an array and showing all of the found errors / potential solutions

This can also be made `const`

sites/example-project/svelte.config.js
```diff
@@ -4,6 +4,10 @@ import adapter from '@sveltejs/adapter-static';
 import { evidencePlugins } from '@evidence-dev/plugin-connector';
 /** @type {import('@sveltejs/kit').Config} */
 
+function errorHandler(warning) {
+	throw new Error(warning.message);
```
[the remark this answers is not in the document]
ItsMeBrianD: Should also have types / checking, we don't _know_ that `warning` is always going to be an `Error`

sites/example-project/svelte.config.js
```diff
@@ -13,6 +17,9 @@ const config = {
 			postcss: true
 		})
 	], // Modify preprocess to allow for loading of $lib instead of package version of components library
+	onwarn: (warning) => {
+		errorHandler(warning);
+	},
```
ItsMeBrianD, line 22:
```suggestion
	onwarn: errorHandler
```

packages/core-components/src/lib/unsorted/ui/ErrorOverlay.svelte
hughess: There could be multiple of these things going wrong at once, but I'm not sure if vite would supply all of the reasons or only one at a time. I'm assuming one at a time since that's what appears in the current HMR error overlay

sites/example-project/svelte.config.js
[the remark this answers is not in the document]
hughess: it looks like `warning` is an `object`.

It seems to trigger the "non-error" section of the hooks.js files and logs "page threw a non-error" to the console

PR #1354, opened 2023-11-23
Add fuzzy searching via fuse.js

csjh:
### Description

Replaces string filtering (with `contains`) with fuzzy searching to enable similarity search filtering in DataTables

### Checklist


mcrascal:
Seems like a huge improvement. 

I think turning down the default [threshold](https://www.fusejs.io/api/options.html#fuzzy-matching-options) might make sense for our case -- I am finding it a little too "loose".  

Maybe would make more sense as a separate issue, but would it be possible to preserve the original height of the table during search and pagination? It's pretty jarring to have the layout shift on keystrokes. 

Other than that LGTM.

csjh:
re: layout shift, do you just mean the shift when it goes from a full page of results -> less than a full page?

mcrascal:
> re: layout shift, do you just mean the shift when it goes from a full page of results -> less than a full page?

Yes exactly. I think ideally the table container would stay the same height. It's a different issue though, we can tackle it separately.  

New threshold seems great to me -- ship it.

ItsMeBrianD (approved):
This lgtm

PR #1356, opened 2023-11-27
Fix massive chart rendering on iOS

csjh:
### Description

Adds a check of the div size to switch to an svg if canvas would be too big for ios

### Checklist


packages/component-utilities/src/echarts.js
```diff
@@ -12,9 +12,16 @@ import debounce from 'debounce';
 
 /** @type {import("svelte/action").Action<HTMLElement, ActionParams>} */
 export default (node, option) => {
+	// https://github.com/evidence-dev/evidence/issues/1323
+	const useSvg =
+		['iPad Simulator', 'iPhone Simulator', 'iPod Simulator', 'iPad', 'iPhone', 'iPod'].includes(
+			navigator.platform
+			// ios breaks w/ canvas if the canvas is too large
```
mcrascal, line 19: navigator.platform is deprecated? Couldn't we just trigger the SVG whenever the chart was large enough, regardless of the browser we're serving it to?

mcrascal (commented):
LGTM except this navigator.platform check.

packages/component-utilities/src/echarts.js
csjh: The `navigator.platform` bit isn't deprecated as much as it is discouraged afaict. I don't think it would matter too much if we did it on every browser though.

PR #1365, opened 2023-11-30
Universal SQL: DuckDB Connector improvements

ItsMeBrianD:
### Checklist

### DIFF
--- a/packages/duckdb/index.cjs
+++ b/packages/duckdb/index.cjs
@@ -58,13 +58,9 @@ const mapResultsToEvidenceColumnTypes = function (rows) {
 	});
 };
 
-/** @type {import("chalk")} */
-let chalk;
 
 /** @type {import("@evidence-dev/db-commons").RunQuery<DuckDBOptions>} */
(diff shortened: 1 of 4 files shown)


packages/duckdb/index.cjs
```diff
@@ -121,7 +125,13 @@ module.exports.getRunner = async (opts, directory) => {
 module.exports.testConnection = async (opts, directory) => {
 	const r = await runQuery('SELECT 1;', { ...opts, filename: path.join(directory, opts.filename) })
 		.then(() => true)
-		.catch((e) => ({ reason: e.message ?? 'File not found' }));
+		.catch((e) => {
+			if (typeof e === 'string' && Boolean(e)) {
```
csjh, line 129: Is the `Boolean(e)` only meant to catch empty strings? Would be clearer as an `e !== ''` imo

csjh (approved):

packages/duckdb/index.cjs
ItsMeBrianD:
```suggestion
			if (typeof e === 'string' && e !== '') {
```

PR #1366, opened 2023-11-30
Add echarts custom options

hughess:
### Description
Adds 2 props to chart components:
- `echartsOptions`: lets you supply an object containing Echarts config, which gets merged with the config of the chart
- `printEchartsConfig`: (true/false) helper prop for custom chart development - inserts a code block with the current echarts config onto the page so you can see the options used and debug your custom options

### Remaining Issues
- ~~Needs work for following components:~~
   - ~~FunnelChart~~
   - ~~SankeyChart~~
   - ~~USMap~~
- ~~Needs to be incorporated into `echartsCopy.js`, `echartsCanvasDownload.js`, and `echartsMap.js`~~

### Checklist

### DIFF
--- a/sites/docs/docs/components/area-chart.md
+++ b/sites/docs/docs/components/area-chart.md
@@ -146,7 +146,7 @@ hide_table_of_contents: false
 
 <table>
 <tr>	<th class='tleft'>Name</th>	<th class='tleft'>Description</th>	<th>Required?</th>	<th>Options</th>	<th>Default</th>	</tr>
-<tr>	<td>echartsOptions</td>	<td>Custom Echarts options to override the default options. <a href='/components/echarts-options'>See detailed walkthrough</a> for available options.</td>	<td class='tcenter'>-</td>	<td class='tcenter'>{'{{config}}'}</td>	<td class='tcenter'>-</td>	</tr>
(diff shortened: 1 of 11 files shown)


archiewood:
> Do you think we should fix it now?

no need if not easy

> Agreed on the concerns - I was thinking this should be a separate page in the docs, which we can link to from each chart type. That way we can include a bit more of a warning/explanation

sounds very sensible

hughess:
Couple other tricky things about this atm:
- To access the yAxis, you need to use this syntax:
```
echartsOptions={{
   yAxis: [
      { 
         something: 'something'
      }
   ]
}}
```

- To reference a specific series, you need to pass it in the correct order, using empty curly braces for the series you don't want to change. e.g., to change the third series:

```
 echartsOptions={{
        series: [
            {},
            {},
            {symbol:'triangle'}
        ]
    }}
```

- As far as I can tell, it's not possible to change all series at once through this

hughess:
One other gotcha with this - if you're using a horizontal chart `swapXY=true`, the axis you need to make changes on will be flipped in the echarts options. The y-axis in Evidence would be referenced with `xAxis` in ECharts and vice versa

hughess:
Last thing I noticed - it seems to not be possible to change the series type using this approach. I've created an issue with echarts for this:
https://github.com/apache/echarts/issues/19349

archiewood:
should / are we able to error handle the eCharts options if you pass an invalid config?

archiewood:
I am writing some rudimentary docs for this

PR #1371, opened 2023-11-30
Universal SQL: Input syntax and cleanup

### LINKED ISSUE #1355: Universal SQL: Input Syntax and Clean up
## Feature Description

* From -> data 
* Error handling -> no exclamation points, less stack trace 
* On page -> match error state to `value` error states  

```
<Dropdown
  data={trips}
  label="Month Picker"
  value=month
/>
```

csjh:
### Description

Fixes #1355 

image

### Checklist


mcrascal:
I think it might be cleaner + more consistent to replace w/ the error indicator we use on values? e.g. just a span, rather than a re-styled drop-down.

csjh:
Yeah makes sense, I just thought the span -> select change was a bit odd looking

PR #1372, opened 2023-11-30
Fix tests

archiewood:


archiewood:
okay so the tests all pass here. which I assume means i can merge into main and then the tests will also pass in version packages...

PR #1373, opened 2023-11-30
Universal SQL: prevent query clobbering

ItsMeBrianD:
### Checklist


csjh:
Was this an existing issue? I thought sources were executed one-by-one, so I don't understand how this could happen

PR #1375, opened 2023-12-04
feat: add option param for postgres pool

### LINKED ISSUE #1358: Add Options to Postgres Connector?
Hi, this is sorta a question and potentially a connector update request. See below for the answers to the GH issue template, but in short, I'm wondering if it's currently possible to pass postgres connection options?

I'm using the postgres wire protocol, but not an actual postgres server, so it's helpful for configuration to be able to pass options at connection time. E.g. what I do w/ `psycopg2` in python:

```python
conn = psycopg2.connect(
    host=HOST,
    port=5432,
    user=EMAIL,
    password=PASSWORD,
    dbname=DBNAME,
    options="organization=public",
)
```

So basically having the `options` parameter available.

I also quickly tried a URL encoded connection string, but that didn't work. If it's indeed possible, and I'm just dumb, I'll have another look :).

Being able to issue commands at connection-time may also work (e.g. `SET organization ...`).

Thanks!

## What do you use this database for?

Postgres, which is already supported

## Are there any popular and well-maintained clients for this database?

`pg`

## Are you interested in contributing this database connector?

Maybe, if it's easy :)

tshauck:
### Description

Hi, this PR if merged would close #1358 by adding `options` to the postgres settings options.

I did a bit of manual testing, and could see the option pass though certainly happy to update to make things robust.

image

This is what the UI looks like with the new field.

image

### Checklist

Thanks!

### DIFF
--- a/.changeset/happy-games-wink.md
+++ b/.changeset/happy-games-wink.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/postgres': patch
+---
+
+adds options parameter to postgres connection settings
--- a/packages/postgres/index.cjs
+++ b/packages/postgres/index.cjs
@@ -51,7 +51,11 @@ const envMap = {
 		{ key: 'POSTGRES_SCHEMA', deprecated: true },
 		{ key: 'schema', deprecated: true },
 		{ key: 'SCHEMA', deprecated: true }
-	]
+	],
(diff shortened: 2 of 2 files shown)


ItsMeBrianD (approved):
Looks perfect!

PR #1377, opened 2023-12-04
Universal SQL: Misc Fixes

ItsMeBrianD:
### Description

While writing the SQL Console component, I encountered a few issues that were easily resolved

### Checklist

### DIFF
--- a/packages/preprocess/src/process-queries.cjs
+++ b/packages/preprocess/src/process-queries.cjs
@@ -184,26 +184,6 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 		const all_query_stores = valid_ids.map((id) => `$: ${id} = $_${id};`);
 
 		queryDeclarations += `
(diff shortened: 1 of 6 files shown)


packages/preprocess/src/process-queries.cjs
```diff
@@ -247,6 +227,27 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
             }
         });
 
+		import { browser, dev } from "$app/environment";
```
csjh, line 230: Why was this moved outside `queryDeclarations`?

csjh (approved):

packages/preprocess/src/process-queries.cjs
ItsMeBrianD: Because we want it to always be available; I had to put a dummy query on my page with the sql console so that the query deps would actually load onto the page

PR #1379, opened 2023-12-04
Universal SQL: DataTable + Chart Error States

ItsMeBrianD:
### Description

This makes sure that the DataTable and Chart report errors - even if the query is reporting a loading state

### Checklist


packages/core-components/src/lib/unsorted/viz/Chart.svelte
```diff
@@ -821,7 +821,7 @@
 	$: if (data?.error) error = data.error.message;
```
csjh, line 821: Should this be changed to `$: if ($data?.error) error = data.error.message ?? data.error;` as well?

csjh (approved):

PR #1381, opened 2023-12-04
Fix wack inputs requirement

csjh:
### Description

Adds a dummy fetch during SSR to evaluate the page and saturate a map that gets inputs and sql strings from the page

Couldn't get `node-async.js` working so queries are still evaluated synchronously - not a major concern right now, more of a worry when we're doing motherduck querying, and shouldn't be an impossible fix

### Checklist


ItsMeBrianD:
Will this account for nested inputs? (e.g. if I made a filter row component that included dropdowns)

csjh:
Ah forgot nested inputs was a req, shouldn't be too difficult

csjh:
Some very hacky stuff is needed for nesting, we could do with more `Array.isArray` and `instanceof QueryStore` checks around

ItsMeBrianD:
Not a blocker but another thought; will this support nested input components?

e.g. if I have some (reusable) report section that has filters on it, will those filters work;

(this is a component not a page)
```ReportSection.svelte
<Dropdown name="Category" />
<Datepicker name="DayOf" />

<!-- some query -->
<!-- some dataviz -->
```

csjh:
It should work, as long as you're getting the `inputs` store from context properly and everything.

PR #1383, opened 2023-12-05
Universal SQL: Fix `build:sources` regressions

ItsMeBrianD:
### Description

[asciicast](https://asciinema.org/a/Ky1tnnKsCgdadbaZYhZ0lxMWt)

### Checklist


packages/plugin-connector/src/data-sources/build-sources.js
```diff
@@ -93,14 +93,18 @@ export const buildSources = async (
 			},
 			/**
 			 * @param {string} name
+			 * @returns {boolean} true if query is included in filters
 			 */
 			isFiltered: (name) =>
-				Boolean(filters?.queries?.has(name) || filters?.queries?.has(`${source.name}.${name}`)),
+				Boolean(filters?.queries?.has(name) || filters?.queries?.has(`${source.name}.${name}`)) ||
+				!filters?.queries,
```
csjh, line 100: Doesn't this make the `Boolean` part irrelevant? If `filters?.queries` is true, then it doesn't matter if either call of `filters?.queries?.has`  is true

ItsMeBrianD: This should mean:

```
table name generically referenced || table name specifically referenced || no filters were specified

```

csjh: Ah yeah missed the exclamation mark

sites/example-project/package.json
```diff
@@ -4,7 +4,7 @@
 	"scripts": {
 		"dev": "vite dev --port 3000",
 		"build": "vite build",
-		"sources": "evidence sources",
+		"build:sources": "__EXAMPLE_PROJECT=true evidence sources",
```
csjh, line 7:
```suggestion
		"sources": "__EXAMPLE_PROJECT=true evidence sources",
```

csjh (approved):

PR #1382, opened 2023-12-05
Universal SQL: Docs

archiewood:
### Description

Documentation for Universal SQL - #1187 

## Outstanding

- [x] add link to legacy docs using vercel link
- [x] page for plugins @mcrascal 
    - [ ] source plugins page @ItsMeBrianD
- [x] various images (dropdown), 
- [x] image for "Filters" page
- [x] update to "how it works" diagram
- [x] confirm prop names for label in DropdownOptions
- [x] check that the DEPLOYMENT instructions are still correct

maybe

- [x] Some new kind of page on the Evidence Query Engine / USQL? (May not be required)
- [x] more examples on "filters" page

PR #1399, opened 2023-12-08
source plugin docs

mcrascal:
Reorganizes plugins section to accomodate source plugins, adds source plugins docs.


mcrascal:
@ItsMeBrianD will need some help on the creating a source plugin section.

archiewood:
cherry picked into #1382

PR #1409, opened 2023-12-14
Fix example project

archiewood:
### Description

The example project now runs 

```bash
pnpm run sources:example-project
pnpm run dev:core-components
pnpm run dev:example-project
```

## To Do
- fix templated page syntax


archiewood:
This is pretty helpful - Lots of stuff is now working!

There are a few charts that are not working and are hanging in the skeleton loading state.

ItsMeBrianD:
Merging in next may fix those hanging charts, depending on when this was branched off

archiewood:
@ItsMeBrianD it does!

archiewood:
tests are failing - but I think that is because tests on next are failing too

PR #1413, opened 2023-12-15
observe changes in headers for TOC

mcrascal:
### Description

Adds mutation observer to table of contents, and a test case for <Value /> tag in a header to avoid rendering loading... in TOC.

### DIFF
--- a/packages/core-components/src/lib/unsorted/ContentsList.svelte
+++ b/packages/core-components/src/lib/unsorted/ContentsList.svelte
@@ -17,7 +17,9 @@
 			updateLinks();
 		});
 
-		observer.observe(document.body, { subtree: true, characterData: true, childList: true });
+		headers.forEach((header) => {
+			observer.observe(header, { subtree: true, characterData: true, childList: true });
+		});
 
 		return observer;
 	}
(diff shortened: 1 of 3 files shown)


packages/core-components/src/lib/unsorted/ContentsList.svelte
```diff
@@ -1,12 +1,34 @@
...
+	function updateLinks() {
+		headers = Array.from(document.querySelectorAll('h1.markdown, h2.markdown'));
 		headers.forEach((header, i) => {
-			header.id = encodeURIComponent(header.innerText + i);
+			// Headers may contain values that change in response to user input, so we create our anchors as just the position on the page.
+			header.id = encodeURIComponent(i + 1);
 		});
+	}
+
+	function observeDocumentChanges() {
```
ItsMeBrianD, line 15: Is this firing any time the DOM changes anywhere? Can we scope this somehow to avoid running a querySelectorAll each time?

Probably isn't a big deal in the grand scheme of things, but could (stress on could, this is probably a nitpick) cause perf issues because it's blocking on the main thread

ItsMeBrianD (approved):

PR #1423, opened 2023-12-21
Telemetry "parity" for USQL

ud3sh:
### Description

Maps build source queries to old query events.  Doing the profile source control under a different PR.

### Checklist


packages/telemetry/index.cjs
```diff
@@ -46,40 +77,84 @@ const logEvent = async (eventName, dev, settings) => {
...
-module.exports = logEvent;
+/**
+ * Logs an event emiited from source queries
+ * @param {TelemetryEventName} eventName
+ * @param {string | undefined} [databaseName]
+ * @param {string | undefined} [sourceName]
+ * @param {string | undefined} [queryName]
+ */
+const logQueryEvent = async (eventName, databaseName, sourceName, queryName, dev = false) => {
+	//TODO there is no concept of dev mode when running npm run sources
```
ud3sh, line 139: @hughess I don't think we can easily tell apart `npm run sources` commands run by CI vs during development.  What I did for now was to set  `dev=false` for all events emitted as a result of `npm run sources`

@ItsMeBrianD or @csjh any alternative ideas?  I just thought about checking for the existence of evidence.setting.json.

packages/telemetry/index.cjs
```diff
@@ -1,8 +1,20 @@
...
 const { readJSONSync, writeJSONSync, pathExistsSync } = require('fs-extra');
 const wK = 'ydlp5unBbi75doGz89jC3P1Llb4QjYkM';
+const Analytics = require('analytics-node');
+
+/**
+ * @typedef {'usageStatsDisabled' |
+ * 			 'db-plugin-unvailable' |
+ * 			 'db-connection-error' |
+ * 			 'db-connection-success' |
+ * 			 'source-connector-not-found' |
```
ud3sh, line 12: @hughess Added a new event `source-connector-not-found` for detecting when the connector is not found via build sources (I am guessing this happens pretty rarely). Can remove this if it's not that useful.

packages/telemetry/index.cjs
hughess: @ud3sh doesn't the built-in svelte `dev` variable cover that for us?

packages/plugin-connector/src/data-sources/build-sources.js
```diff
@@ -105,10 +109,13 @@ export const buildSources = async (
 			source.sourceDirectory
 		);
 		if (connectionValid !== true) {
+			logQueryEvent('db-connection-error', source.type, source.name);
 			throw new Error(
 				chalk.red(`[!] ${chalk.bold(source.name)} failed to connect; ${connectionValid.reason}`)
 			);
 		}
+		//TODO evidence-1344 and db-connection-error didn't have an equivalent for event in legacy
```
ud3sh, line 117: @hughess in legacy we didn't emit an event when there was a connection error during query runs.  It was only emitted when the connection test failed.
Here I am doing it in both places.  Let me know if this not useful.

packages/telemetry/index.cjs
```diff
@@ -46,40 +77,84 @@ const logEvent = async (eventName, dev, settings) => {
...
+			const projectProfile = await getProfile();
 			var analytics = new Analytics(wK);
-			analytics.track({
+			const payload = {
 				anonymousId: projectProfile.anonymousId,
 				event: eventName,
 				properties: {
 					devMode: dev,
 					repoHash: repo,
 					database: database, // logs database type (postgres, snowflake, etc.)
```
ud3sh, line 112: @hughess storing additional fields that are now available via USQL in hashed form.  Can be removed, unhashed etc, if need be.

archiewood:
I can imagine 4 scenarios where queries are run in USQL:
1. **Developer runs Source SQL** `npm run sources` during development (infrequent) 
2. **Developer runs DuckDB SQL**, either by accessing a page, changing the SQL on a page directly, or by changing an input component
3. **Build or CI process runs Source SQL** `npm run sources` in pipeline
4. **Deployed site runs DuckDB SQL** due to users accessing pages, or changing and input component


number |environment | query type | source
--|--|--|--
1 | dev | source-query | snowflake
2 | dev | wasm-query | n/a
3 | prod | source-query | snowflake
4 | prod | wasm-query |  n/a


are we able to distinguish all these cases?

packages/telemetry/index.cjs
csjh: Not necessarily, since `npm run sources` isn't run while the dev server is running

I think we could check the node_env environment variable to determine source HMR though

hughess: ahhh got it

hughess:
@archiewood it seems like 2 and 4 in your table are covered, but 1 and 3 at the moment we only know that it's source queries, but not whether it was done in dev or during a build

hughess:
@ud3sh does this PR solve the profile json issue, or is that for a future change?

ud3sh:
> @ud3sh does this PR solve the profile json issue, or is that for a future change?

Separate change.

packages/telemetry/index.cjs
ud3sh: I don't recall seeing the `NODE_ENV` var when runnign `npm run sources` (I printed these out to see).  It wouldn't make sense for it to be populated I'd think unless you run `NODE_ENV=development npm run dev`.

The evidence.settings.json wouldn't be available in non-dev environment right?
- it's not source controlled hence when you run in on CI, it won't be available.

Moving this back to draft.  I think this is something we can solve.

hughess: If we're blocked from everything else in the meantime, it might be worth going ahead without the dev mode indicator and adding it later. Not sure what info will be available prior to this PR going in

ud3sh: > If we're blocked from everything else in the meantime, it might be worth going ahead without the dev mode indicator and adding it later. Not sure what info will be available prior to this PR going in

No events are going in post USQL.

hughess (approved):

PR #1425, opened 2023-12-21
make filter docs less confusing

mcrascal:


archiewood:
Edits looks good, 

but the fact this is necessary makes me think we have not named the props correctly

PR #1433, opened 2023-12-27
Allow motherduck and in-memory duckdb connections

archiewood:
### Checklist


packages/duckdb/index.cjs
```diff
@@ -185,17 +185,27 @@ module.exports.getRunner = async (opts, directory) => {
...
+		return runQuery(queryContent, { ...opts, filename: filename }, batchSize);
 	};
 };
 
 /** @type {import("@evidence-dev/db-commons").ConnectionTester<DuckDBOptions>} */
 module.exports.testConnection = async (opts, directory) => {
-	const r = await runQuery('SELECT 1;', { ...opts, filename: path.join(directory, opts.filename) })
+	// Determine filename based on string
+	const filename =
+		opts.filename.startsWith('md:?motherduck_token=') || opts.filename == ':memory:'
```
archiewood, line 201: a bit of copypasta here, should probably define elsewhere?

packages/duckdb/index.cjs
```diff
@@ -185,17 +185,27 @@ module.exports.getRunner = async (opts, directory) => {
...
 		// Filter out non-sql files
 		if (!queryPath.endsWith('.sql')) return null;
-		return runQuery(
-			queryContent,
-			{ ...opts, filename: path.join(directory, opts.filename) },
-			batchSize
-		);
+		// Determine filename based on string
+		const filename =
+			opts.filename.startsWith('md:?motherduck_token=') || opts.filename == ':memory:'
```
archiewood, line 190: I wonder if a more robust way is to assume that it is an absolute path by default, and only revert to relative if certain conditions are met.

https://duckdb.org/docs/connect

It seems there are three possibilities in general
- An existing file
- In memory - `:memory:`
- If no file is specified, one will be created (the duckdb CLI creates an in-memory file , for example)

.changeset/grumpy-dancers-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@evidence-dev/duckdb': major
```
ItsMeBrianD, line 2: patch fix

packages/duckdb/index.cjs
ItsMeBrianD: Is there any reason this can't live in `runQuery`?

ItsMeBrianD (changes requested):
Not a major change, filename can probably be declared in `runQuery`

PR #1438, opened 2023-12-28
issues-1344 Store profile under source controlled customization directory in template

ud3sh:
### Description

* This changes the location of `.profile.json` from `.evidence/template` (not source controlled) to .evidence/customization (source controlled)
* If the file exists in the old location it will be migrated
* In order to ease development, this keep `.profile.json` in example-project so we don't create a new profile during evidence development (i.e it migrates it every time but telemetry events will all share the same old profile id)
* upon cloning the template, the user will see the new .profile.json in their git diff

### Checklist


packages/core-components/src/lib/unsorted/ui/TelemetryOptOut/TelemetrySettingsPanel.svelte
```diff
@@ -31,7 +31,9 @@
 					<p>Each time you run a query, we get the following pieces of information:</p>
 					<ol class="list-decimal px-8">
 						<li>
-							A random identifier that is stored in <code>.evidence/template/.profile.json</code>
+							A random identifier that is stored in <code
+								>.evidence/customization/.profile.json</code
```
archiewood, line 35: This is probably an unintentional change from the formatter

Maybe won't make a difference

archiewood (approved):

hughess (approved):

PR #1444, opened 2024-01-03
Add annotations to databricks and trino connectors

ItsMeBrianD:
### Checklist


archiewood:
Tested this out in example project getting the following error.

I suspect that it's the databricks token, which I have updated

CleanShot 2024-01-03 at 17 43 34@2x

packages/databricks/index.cjs
```diff
@@ -139,7 +139,7 @@ module.exports.testConnection = async (opts) => {
 module.exports.options = {
 	token: {
 		title: 'Personal Access Token',
-		type: 'text',
+		type: 'string',
```
archiewood, line 142: is this update right?

archiewood (commented):
Think this is fine, but let me know if I have misinterpreted the errors

packages/databricks/index.cjs
archiewood: end user has verified

PR #1446, opened 2024-01-04
Schema Explorer UI redesign

archiewood:
### Description

Schema Explorer needed a bit of love. 

Here is an idea for a more traditional source -> table -> column approach.

CleanShot 2024-01-04 at 03 38 41

Feedback very much welcomed!

### Checklist


packages/core-components/src/lib/unsorted/ui/Explore/SchemaExplorer.svelte
```diff
@@ -0,0 +1,108 @@
+<script>
```
ItsMeBrianD, line 1: Please move this component into the Molecule directory rather than unsorted

Issue #1228, opened 2023-09-30
Entering an invalid value for SSL when connecting to postgres crashes evidence


scsmithr:
## Steps To Reproduce

I'm using the template repo here: https://github.com/evidence-dev/template

1. `npm install` & `npm run dev`
2. Go to `127.0.0.1:3000/settings`
3. Select postgres, enter in some information, then enter in an invalid value for the `SSL` field (I entered "require" thinking it mapped to `sslmode`).
4. Observe network error and the following logs:

```
  Connection Test running.../Users/sean/Code/github.com/glaredb/reports/node_modules/pg/lib/connection.js:85
        if ('key' in self.ssl) {
                  ^

TypeError: Cannot use 'in' operator to search for 'key' in require
    at Socket.<anonymous> (/Users/sean/Code/github.com/glaredb/reports/node_modules/pg/lib/connection.js:85:19)
    at Object.onceWrapper (node:events:628:26)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

I am also unable to start up evidence again as I'm met with the following:

```
% npm run dev

> my-evidence-project@0.0.1 dev
> evidence dev --open /


Forced re-optimization of dependencies

  VITE v4.3.9  ready in 481 ms

  ➜  Local:   http://127.0.0.1:3000/
  ➜  Network: use --host to expose
  ➜  press h to show help
  orders_by_month running.../Users/sean/Code/github.com/glaredb/reports/node_modules/pg/lib/connection.js:85
        if ('key' in self.ssl) {
                  ^

TypeError: Cannot use 'in' operator to search for 'key' in require
    at Socket.<anonymous> (/Users/sean/Code/github.com/glaredb/reports/node_modules/pg/lib/connection.js:85:19)
    at Object.onceWrapper (node:events:628:26)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

[...]

PR #1448, opened 2024-01-05
Add 'no-verify' option for pg SSL

archiewood:
### Checklist

### DIFF
--- a/.changeset/curly-hotels-teach.md
+++ b/.changeset/curly-hotels-teach.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/postgres': patch
+---
+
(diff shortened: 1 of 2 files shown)


archiewood:
This doesn't work as no-verify is not an sslmode: option

archiewood:
Prior to USQL we handled this like this: https://github.com/evidence-dev/evidence/blob/1e5678358f8cc169f8c5c02752d558db4f4fe356/packages/postgres/index.cjs

the user could pass the strings: `true`, `false`, or `no-verify` to the `ssl` option

mcrascal:
Related #1228

PR #1477, opened 2024-01-10
1457 Multiselect Dropdown, Date Range & Input Cleanup

mcrascal:
@csjh 

This has the date range migrated from here https://github.com/evidence-dev/shadcn-components/tree/main/plugin, bumps svelte + adds the deps for shadcn. 

None of the components in the plugin above are wired into USQL. The shadcn components I'm putting in atoms/shadcn for now, but feel free to move these if I haven't landed them in the right place.

Here's what I'm thinking: 
- Replace the current `dropdown` w/ the select component in the above 
- Wire the inputs into USQL + inputs store -- @ItsMeBrianD has some ideas on date range 
- Replace the existing accordian, modal, and tabs if time 
- Possibly move all the shadcn base components from the above?


packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -21,15 +36,76 @@
 	/** @type {string} */
 	export let name;
 
+	export let multiple = false;
+
+	export let hideDuringPrint = true;
+
 	/** @type {string} */
-	export let defaultValue = null;
+	export let defaultValue = undefined;
```
archiewood, line 44: This appears now to not work properly - I can't unselect the defaultValue, as the dialog to select flashes and then closes immediately

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -41,43 +117,110 @@
...
+									{/if}
+								</div>
+							{/if}
+						{/if}
+					</Button>
+				</Popover.Trigger>
+				<Popover.Content class="w-[200px] p-0" align="start" side="bottom">
+					<Command.Root shouldFilter={false}>
+						<Command.Input placeholder={title} bind:value={search} />
+						<Command.List>
```
archiewood, line 195: This is a bit "flashy" when typing

CleanShot 2024-02-06 at 18 12 00

packages/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
```diff
@@ -0,0 +1,95 @@
...
+
+	function dateToYYYYMMDD(date: Date) {
+		return date.toISOString().split('T')[0];
+	}
+
+	const inputs: Writable<object> = getContext(INPUTS_CONTEXT_KEY);
+
+	export let name: string;
+	export let title: string | undefined;
+	export let hideDuringPrint = true;
```
archiewood, line 25: false?

packages/core-components/src/lib/atoms/inputs/date-range/_DateRange.svelte
```diff
@@ -0,0 +1,159 @@
...
+
+	const dfShort = new DateFormatter('en-US', {
+		dateStyle: 'short'
+	});
+
+	export let selectedDateRange: DateRange | undefined;
+	export let start: string;
+	export let end: string;
+
+	$: calendarStart = YYYYMMDDToCalendar(start);
```
archiewood, line 34: hard reload managed to get be Jan 1st 1970 again
CleanShot 2024-02-06 at 18 26 22@2x

packages/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
```diff
@@ -43,30 +46,32 @@
...
-		{:else}
-			<slot />
-			{#if hasQuery}
-				{#if $query.error}
-					{$query.error}
-				{:else}
-					{#each $query as { label, value }}
-						<ButtonGroupItem {value} valueLabel={label} />
+<HiddenInPrint enabled={hideDuringPrint}>
+	<div class="inline-flex flex-col mt-2 mb-4 mx-1">
```
archiewood, line 50: might be out of scope but we are not super visually consistent with new dropdown any more
CleanShot 2024-02-06 at 18 29 30@2x

archiewood (changes requested):
Overall very much digging these.

Shadcn styling is a big leap on. 

Seems pretty performant on big data sets, except search which kills dropdown at the mo

Some Docs for Date Range needed

packages/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
csjh: I thought consensus was default to hiding during print `Hide inputs on print, incl. Print option default to false` but I might be misinterpreting

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -21,15 +36,76 @@
 	/** @type {string} */
 	export let name;
 
+	export let multiple = false;
+
+	export let hideDuringPrint = true;
```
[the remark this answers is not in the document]
archiewood: my bad

packages/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
archiewood: yep you are right

archiewood:
I think we also need to improve the keyboard accessibility of this component:

- [x] I can tab to the component
- [x] I can open the dropdown with space or enter - this lands me in the text search
- [ ] I can move the focus from the text search to the options list without using the mouse (e.g. using tab)
- [x] I can select the options using enter

mcrascal:
> I currently cannot complete a selection without using a mouse

@archiewood the `up` and `down` arrows allow you to move the selection in both the multi-select and the select while your cursor remains in the search box. `Enter` adds the highlighted selection to the selected values. 

CleanShot 2024-02-07 at 09 22 28

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -41,43 +117,108 @@
...
+										{/each}
+									{/if}
+								</div>
+							{/if}
+						{/if}
+					</Button>
+				</Popover.Trigger>
+				<Popover.Content class="w-[200px] p-0" align="start" side="bottom">
+					<Command.Root shouldFilter={false}>
+						<Command.Input placeholder={title} bind:value={search} />
```
archiewood, line 192:
```suggestion
       {#if items.length > 5}
						<Command.Input placeholder={title} bind:value={search} />
			 {/if}
```

maybe if the options are short we should hide search input?

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -41,43 +117,108 @@
...
+						<Command.List>
+							<Command.Empty>No results found.</Command.Empty>
+							<Command.Group>
+								<slot />
+
+								{#if hasQuery}
+									{#await $items.fetch() then _}
+										<VirtualList height="160px" items={$items} let:item>
```
archiewood, line 200: This seems too tall if you only have eg three options?

Not sure if an issue?

CleanShot 2024-02-07 at 09 32 14@2x

packages/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
mcrascal: I can clean this up

mcrascal:
Not blocking, but the virtual list on large data sets isn't fully keyboard accessible. You can't force it to scroll with the down or up keys.

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -21,15 +36,76 @@
...
+		if (multiple) {
+			values = $selectedValues.map((x) => x.value);
+			values.toString = () => jsToDuckDB(values);
+		} else {
+			values = $selectedValues[0]?.value ?? null;
+			// the default `toString` method for Dates aren't good for duckdb
+			if (values instanceof Date) values.toString = () => jsToDuckDB(values);
+		}
+
+		$inputs[name] = {
```
archiewood, line 77: This seems to deprecate support for `'${inputs.my_input_name}'` in favour of `'${inputs.my_input_name.label}'`

This is a breaking change

archiewood: hmm I'm not sure. 

I guess an alternative would be some kind of proxy that maps `inputs.my_input_name` -> `inputs.my_input_name.label`?

{}

csjh: That would require proxying the string `inputs.my_input_name.label`, but you can't proxy strings (only objects)

One thing that's possible would be doing ```js
$inputs[name] = {
    // ...
    valueOf: () => label,
    toString: () => label
}
```
which would fix it in `SELECT * FROM somewhere WHERE id = '${inputs.something}'` and `inputs.something == 'blah'`, but *not* `inputs.something === 'blah'`

Comment (csjh) [reply packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte:77]: image

File: packages/core-components/package.json [hunk 10]
```diff
@@ -33,21 +33,31 @@
...
+		"clsx": "^2.0.0",
+		"cmdk-sv": "^0.0.12",
 		"echarts": "5.4.3",
 		"echarts-stat": "1.2.0",
 		"export-to-csv": "0.2.1",
+		"formsnap": "^0.4.2",
 		"fuse.js": "^7.0.0",
 		"lodash.debounce": "^4.0.8",
 		"prismjs": "1.29.0",
+		"radix-icons-svelte": "^1.2.1",
```
ItsMeBrianD, line 52: We should continue to use steeze UI icons for consistency

packages/core-components/package.json
```diff
@@ -33,21 +33,31 @@
...
 		"ssf": "0.11.2",
 		"svelte-tiny-linked-charts": "1.1.5",
+		"sveltekit-superforms": "^1.13.1",
```
ItsMeBrianD, line 55: Does this introduce 2 form libs, if so, are they both needed?

packages/plugin-connector/package.json
```diff
@@ -29,7 +29,7 @@
 		"lodash.debounce": "^4.0.8",
 		"lodash.merge": "^4.6.2",
 		"ora": "^7.0.1",
-		"svelte": "3.55.0",
+		"svelte": "3.56.0",
```
ItsMeBrianD, line 32: Svelte versions need to be consistent across the project

ItsMeBrianD (commented):
Still need to go more in depth on this w/ it pulled down, but there is a lot to do here I think

ItsMeBrianD:
# This PR is now pointing to a feature branch, not `next`, and will be released independently of `next`.

We should be reviewing for functionality, not code quality at the moment. Next week we will merge the feature branch into next, and worry about code-qual then

archiewood:
Weird tests are failing

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
archiewood: That seems reasonable to me

perhaps we can print a warning to the terminal if you use either that it is deprecated and should be replaced with `inputs.something.label`

PR #1485, opened 2024-01-11
feat: query file hmr

ItsMeBrianD:
### Description

This adds file watching and query updating to external query files

### Checklist

### DIFF
--- a/packages/plugin-connector/src/build-plugins/query-directory-hmr.js
+++ b/packages/plugin-connector/src/build-plugins/query-directory-hmr.js
@@ -1,0 +1,35 @@
+import fs from 'fs/promises';
+/** @type {import("vite").Plugin} */
+export const queryDirectoryHmr = {
+	name: 'evidence:query-directory-hmr',
+
+	configureServer(server) {
+		server.watcher.add('../../queries/**.sql');
+		server.watcher.addListener(
+			'change',
+			/**
+			 *
+			 * @param {string} filename
+			 * @returns {Promise<void>}
+			 */
+			async (filename) => {
+				if (!(filename.includes('.evidence/template/queries') && filename.endsWith('.sql'))) return;
+				const content = await fs.readFile(filename, 'utf-8');
+				const queryId = filename
+					.split('.evidence/template/queries/')
+					.pop()
+					?.split('.sql')[0]
+					.replace('/', '_')
+					.replace('\\', '_');
+
+				server.ws.send('evidence:queryChange', {
+					queryId,
+					content
+				});
+			}
+		);
+	},
+	watchChange(f, opts) {
+		console.log({ f, opts });
+	}
+};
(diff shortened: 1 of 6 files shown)


packages/preprocess/src/process-queries.cjs
```diff
@@ -64,6 +64,23 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 				$: _${id}_query_text = \`${duckdbQueries[id].compiledQueryString.replaceAll('`', '\\`')}\`;
 				$: _${id}_has_unresolved = __checkForUnsetInputs\`${duckdbQueries[id].compiledQueryString.replaceAll('`', '\\`')}\`;
 
+				if (import.meta?.hot) {
```
csjh, line 67:
```suggestion
				if (import.meta.hot) {
```

packages/preprocess/src/process-queries.cjs
```diff
@@ -64,6 +64,23 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
+						let errors = []
+						if (!queryId) errors.push("Malformed event: Missing queryId")
+						if (!content) errors.push("Malformed event: Missing content")
+						if (errors.length) {
+							console.warn("Failed to update query on serverside change!", errors.join("\\n"))
+							return
+						}
+
+						if (queryId === "${id}") {
+							_${id}_query_text = content
```
csjh, line 78: Don't query files technically allow interpolated variables right now? I think this is technically a regression (though query files interpolating variables are kinda weird behaviour anyways)

packages/plugin-connector/src/build-plugins/query-directory-hmr.js
```diff
@@ -0,0 +1,35 @@
...
+			'change',
+			/**
+			 *
+			 * @param {string} filename
+			 * @returns {Promise<void>}
+			 */
+			async (filename) => {
+				if (!(filename.includes('.evidence/template/queries') && filename.endsWith('.sql'))) return;
+				const content = await fs.readFile(filename, 'utf-8');
+				const queryId = filename
```
csjh, line 18: Should probably make a regex for this replace/split/pop menace at some point because I think it pops up in a few places

csjh (approved):
LGTM other than minor regression concern

packages/preprocess/src/process-queries.cjs
ItsMeBrianD: Is `import.meta` guaranteed to exist?

packages/preprocess/src/process-queries.cjs
ItsMeBrianD: Ah yeah that's a good point; I think we can probably get away with "don't do that" for now, it is a pretty odd pattern

PR #1486, opened 2024-01-12
Move context.js -> component utilities chartContext.js

### LINKED ISSUE #895: Allow custom components to access existing components and helper functions
## Feature Description
Allow custom components to access other components or helper functions within a project.

I'm building a custom component within a standalone project and found myself reaching for a couple of things which my custom component file doesn't have access to:
1. Evidence components (e.g., if I want to include a `DataTable` in my component)
2. Helper functions (number formatting and title formatting most important here I think)

## Goal of Feature
Reuse existing components and functions to build something for a specific use case, without having to contribute it into the main Evidence project.

## Current Solution
Build the component within the monorepo and contribute it into the open source project.

## Alternative Ideas
There may be a way to import the components I want to use in the script tag. I think ideal solution is to not require an import for things already in the project, but if there's an import statement that can be copy-pasted and then modified, that would work as well.

ItsMeBrianD:
### Description

This makes some context keys that were previously only available to components in `core-components` available to 3rd party components. Useful for leveraging the existing `Chart.svelte` component when building new viz

### Checklist

Resolves #895


mcrascal:
lgtm

csjh (approved):

mcrascal (approved):

PR #1487, opened 2024-01-12
Add xMin and xMax props to ScatterPlot

ycaffee:
### Description

The ScatterPlot component internally sets the `boundaryGap` EChart option which determines the values shown on the axes (Scatter.svelte:207). You can use the `yMin` and `yMax` options  on the ScatterPlot component to change the values shown on the y-axis but there is currently no way to change the x-axis values. Adding an `xMin` and `xMax` prop would allow users to control the values shown on the x-axis. For example, setting `xMin=0` would  prevent negative values on the x-axis. 

- Added xMin and xMax to the ScatterPlot props
- Updated ScatterPlot docs
- Added an example scatter chart using the `xMin` and `xMax` props to the example site

### Checklist

Example of a scatter plot showing negative values on the x-axis even though negative values are not relevant in this context.
scatter_before_changes

After adding the `xMin` prop, I can control the values on the x-axis. 
scatter_after_changes


archiewood:
@ycaffee great addition, thank you!

ItsMeBrianD (approved):
This looks great to me!

PR #1489, opened 2024-01-15
Standardize sankey chart and sankey diagram

### LINKED ISSUE #1225: SankeyChart and SankeyDiagram
We appear to have both a [SankeyChart](https://github.com/evidence-dev/evidence/blob/main/packages/core-components/src/lib/unsorted/viz/SankeyChart.svelte) and a [SankeyDiagram](https://github.com/evidence-dev/evidence/blob/main/packages/core-components/src/lib/unsorted/diagrams/SankeyDiagram.svelte) in the library. 

These have similar, but not identical source code. I suspect an accidental duplication.

We should delete one of them (I suggest SankeyChart), and make sure the other contains all the pertinent config, and update the docs as appropriate.

mcrascal:
- `SankeyDiagram` and `SankeyChart` now use the same source (SankeyDiagram)
- Adds the duration fix from #1093 to `SankeyDiagram`
- Fixes the docs to only reference `SankeyDiagram` 

closes #1225


archiewood:
@mcrascal did this retain the echartsProps option in SankeyDiagram?

mcrascal:
@archiewood if it supported it before, it still does. I can confirm though.

hughess:
One other place to update is the snippet in the VS Code extension which uses SankeyChart now

archiewood:
@mcrascal I think it's possible it got dropped. 

We implemented the echartsoptions in SankeyChart (not SankeyDiagram) in #1366 so I'm not sure it got copied over!

archiewood:
This was helpful for me: It's a diff of the two Sankey Components:

https://www.diffchecker.com/4A2tNPEO/

mcrascal:
Ah, thanks! Looks like Sankey Diagram never supported the echarts options. I can pull them over.

Issue #1459, opened 2024-01-08
Clean up SQL Console Component 


mcrascal:
- Update the dev mode SQL console to leverage the usql-toolbox console

ItsMeBrianD:
@mcrascal are we ok with keeping the `usql-toolbox` package as a plugin, rather than pulling this all into core-components? We can re-export the sql console from core components to prevent needing to install it in the template

mcrascal:
I'm good with it.

PR #1505, opened 2024-01-18
WIP: Add updated SqlConsole 

ItsMeBrianD:
### Description

Need to update the explore page and make sure it is added to navigation; also want to make the query from the SQL Console exposable (possibly via inputs or another similar store).

image

### Checklist


ItsMeBrianD:
Resolves #1459

archiewood:
## UI 
- **Icon** - love it!
CleanShot 2024-01-19 at 10 03 23@2x

- **Naming** - I think probably we could be tighter here, we have two things called "explorer"
   - To me this thing is a "SQL Runner" or "SQL Console"
   - The thing we currently call a "Schema Explorer" is probably more like a "Schema Viewer" or "Table Viewer", since it doesn't allow you to see any data (currently), just sources, tables, columns

- **Autocomplete** - is very eager with lots of keywords I am unlikely to be wanting 
CleanShot 2024-01-19 at 10 10 54@2x
  - I'd consider restricting to a core set of statements (select, from, limit, group by, order by, where, having, left join, join, with)
  - or only having suggestions after two letters are typed


## Accessing data

- Hard to know my tables are called. They don't autocomplete for me, and I don't have a table viewer

## Functionality
On initial load:
- The SQL console width "jumped" from narrow to fullscreen layout in a jerky way
- CleanShot 2024-01-19 at 09 59 42
- Some issue with database? CleanShot 2024-01-19 at 09 58 15@2x
- CleanShot 2024-01-19 at 10 02 48@2x

Couldn't seem to run any queries

archiewood:
Okay got the db working in Chrome

## SQL writing experience
- Tab does not create indents
- functions when autocompleted (e.g. count) do not autocomplete with brackets
- when I have completed a word, it is still shown as a suggestion
- CleanShot 2024-01-19 at 10 32 04@2x

## Chart Builder
This is a nice start!
- The "delete" button doesn't work for me
- What is a "Break" chart? Is it a chart that breaks?
CleanShot 2024-01-19 at 10 34 07@2x

archiewood:
final thought.

Feels like these should available in Prod.

the SQLrunner is then our first "self serve" feature

ItsMeBrianD:
> I'd consider restricting to a core set of statements (select, from, limit, group by, order by, where, having, left join, join, with)

I'll see what we can do here, the language autocomplete is all done with `@codemirror/sql`

> Hard to know my tables are called. They don't autocomplete for me, and I don't have a table viewer

We could add the schema explorer to this page somehow; alternatively using `FROM [schema].` will autocomplete tables, and columns autocomplete when the table is explicitly named (either fully qualified or an aliased name)

> The SQL console width "jumped" from narrow to fullscreen layout in a jerky way

Will take a look here, this is a little hard; the page has to talk to the layout

> Some issue with database? (Firefox 121.0.1)

Is this only happening for the SQL Console, or other areas as well?

> What is a "Break" chart? Is it a chart that breaks?

Test value that leaked, should be gone now

packages/core-components/src/lib/molecules/sql-console/SqlConsole.svelte
```diff
@@ -0,0 +1,135 @@
...
+		data.fetch();
+	}
+
+	/** @type {HTMLElement | undefined} */
+	let editor;
+	/** @type {SqlConsoleArgs | undefined} */
+	let consoleArgs;
+
+	$: if (consoleArgs) consoleArgs.disabled = disabled;
+	onMount(
```
csjh, line 54: Why is this in an `onMount`?

packages/core-components/src/lib/molecules/sql-console/SqlConsole.svelte
```diff
@@ -0,0 +1,135 @@
...
+	 - Try to fix autocomplete
+	 	- non-namespaced?
+		- keyword updates
+	 */
+
+	/** @type {QueryStore} */
+	export let data = buildQuery(currentQuery);
+
+	$: if (currentQuery) {
+		data = buildQuery(currentQuery);
```
csjh, line 44: It would be nice to have something that strips semicolons, it tripped me up for a second

csjh:
Also a small UI note - the mouse doesn't switch to the text input pointer unless you're hovering over the top line

archiewood:
> > Some issue with database? (Firefox 121.0.1)
>
> Is this only happening for the SQL Console, or other areas as well?

Appears to be only happening with the SQL console. However once the SQL console errors out, anything after requiring DuckDB is then broken

CleanShot 2024-01-31 at 10 12 08

ItsMeBrianD:
> Appears to be only happening with the SQL console. However once the SQL console errors out, anything after requiring DuckDB is then broken

Fixed this one I believe

archiewood:
Confirm fixed

archiewood:
I find this auto indenting of line after select annoying but it's a minor UX gripe

i.e. I would rather just maintain the current indent level, unless I chose to adjust it with tab.

CleanShot 2024-02-01 at 16 25 27

archiewood (approved):

PR #1515, opened 2024-01-23
Add freeze column to datatable

hughess:
### Description
Minimal implementation of a `freezeColumn` prop for `DataTable`,  which freezes the first column on the left and allows you to scroll the other columns.

If `rowNumbers` are shown, these are not frozen - they will scroll away, leaving only the first column of the table frozen on the left.

CleanShot 2024-01-23 at 15 33 01

### Checklist

## To Do
- Fix issue with white background in Firefox - overrides the border of the cells
- Fix issue with edges of white background on mobile - interferes with header bottom border
- Test on Windows and multiple browsers

### DIFF
--- a/.changeset/real-parents-smoke.md
+++ b/.changeset/real-parents-smoke.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/core-components': patch
+'@evidence-dev/components': patch
+---
+
+Add freezeColumn to DataTable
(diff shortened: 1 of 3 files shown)


archiewood:
Why is the white background needed?

hughess:
> Why is the white background needed?

when you scroll horizontally, the content from the right columns will pass behind the frozen column - you can see it underneath if you don't set the background

hughess:
Need to revisit this - would be a great feature to have, but it seems like the table needs to be aware of the width of the columns, which will require some thought

PR #1517, opened 2024-01-23
Basic Theme Support

ItsMeBrianD:
### Description

This PR has a simple support for extending / overwriting the built-in evidence tailwind configuration:

image

There are still areas that are not impacted by the tailwind configuration, and there is a pass needed to rename colors (e.g. adding primary, secondary, success, etc.) but the scope of those changes is much larger than this initial basic support and will be included in a follow-up PR

### Checklist


archiewood (commented):
Am I right in inderstading the behaviour is this:
- user can add tailwind.config.js file, which will append / overwrite existing styles

## Overwriting colors
```js
const keppel = {
  50: '#f3faf7',
  100: '#d7f0e7'
  ...
}

module.exports = {
  theme: {
    extend: {
      colors: {
        blue: keppel
      }
    }
  }
};
```
- anywhere there is e.g. `bg-blue-50`, this would be replaced with the `#f3faf7` colour

## Add additional colours
```js
const keppel = {
  50: '#f3faf7',
  100: '#d7f0e7'
  ...
}

module.exports = {
  theme: {
    extend: {
      colors: {
        keppel
      }
    }
  }
};
```
- `bg-keppel-500` is now also a valid tailwind class

sites/example-project/tailwind.config.cjs
```diff
@@ -1,6 +1,23 @@
...
+// Use find so that we can stop iteration
+altConfigFilepaths.find((file) => {
+	if (fs.statSync(file, { throwIfNoEntry: false })) {
+		presets.push(require(file));
+		return true;
+	}
+	return false;
+});
+
+console.log(presets);
```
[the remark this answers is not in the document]
ItsMeBrianD:
```suggestion
```

ItsMeBrianD:
That understanding is correct, we basically put the user's configuration as a preset that takes precedence over our own preset (`@evidence-dev/tailwind`), so there are a few things that they can't change (e.g. the `content` field is integrated with our plugins system so it can't be touched), but other than that they can define their own configuration

sites/example-project/tailwind.config.cjs
```diff
@@ -1,6 +1,22 @@
...
+const fs = require('fs');
+const path = require('path');
+let presets = [evidenceTailwind];
+
+const altConfigFilenames = ['tailwind.config.js', 'tailwind.config.cjs', 'tailwind.config.mjs'];
+const altConfigFilepaths = altConfigFilenames.map((filename) => path.join('..', '..', filename));
+// Use find so that we can stop iteration
+altConfigFilepaths.find((file) => {
+	if (fs.statSync(file, { throwIfNoEntry: false })) {
+		presets.push(require(file));
```
csjh, line 13: Would `require('tailwind.config.mjs')` work? might have to be a dynamic import

ItsMeBrianD: Ah yeah this is a good shout, that will not work

archiewood:
LGTM

archiewood:
And I guess the big PR that follows involves basically renaming all the colours to things like `primary` rather than `blue` across all our components

ItsMeBrianD:
> And I guess the big PR that follows involves basically renaming all the colours to things like primary rather than blue across all our components, and in the default tailwind config

Correct

ItsMeBrianD:
> AFAICT this doesn't work with @apply? I think we need to maintain support for it.

`@apply` gets baked in, so unless it is baking in as CSS Variables we can't touch those values

ud3sh (approved):

packages/core-components/src/lib/unsorted/ui/Sidebar.svelte
```diff
@@ -191,13 +201,3 @@
...
-
-<style>
-	.selected {
-		@apply text-blue-600;
-	}
-
-	.selected:hover {
-		@apply text-blue-600;
-	}
-</style>
```
mcrascal, line 203: Original behaviour is that a selected sidebar link, when hovered, stays the selected colour. This makes a selected sidebar link have the same hover behaviour as an unselected one.

mcrascal (changes requested):
If we can keep the sidebar behaviour the same as before, and there's a path forward to deal with `@apply`, I'm good with this.

That said, I think losing support for `@apply` is a pretty big concern. Is there a way to do this that doesn't break `@apply`, or what would be our migration plan away from it?

ItsMeBrianD:
> Is there a way to do this that doesn't break @apply, or what would be our migration plan away from it?

I think we'd need to look at how `@apply` is baking the values in. If we can get it to bake in css variables rather than hard-coded colors then we should be okay. If not, then we won't be able to use this anymore for anything themeable.

There are a lot of places in the docs where the usage of `@apply` is discouraged though; so this may be a good change overall

hughess:
@ItsMeBrianD I remember you mentioning a challenge to set hover colours as something relative to the primary colour. Not sure if that applies to this PR specifically or themes more generally, but this package @archiewood found might have a solution:

https://github.com/Qix-/color

It includes functions to manipulate colours, so we could potentially have some sort of darkening or lightening threshold for hover states. Though I'm unsure if that will cover your point that with some colours it's actually a shift in hue that's required vs. a change in lightness

mcrascal (approved):

archiewood:
Needs docs eventually. Could be after we do all the renaming of the tokens

PR #1519, opened 2024-01-24
DataTable total rows

archiewood:
### Description

Not specifying any aggregations: Only numeric columns are aggregated, they are summed

```svelte
<DataTable data={orders_by_category} totalRow=true/>
```

CleanShot 2024-01-24 at 13 13 55@2x

Specifying custom aggregations at a column level. Sum is default, but can be overridden

```svelte
<DataTable data={orders_by_category} rowNumbers=true totalRow=true>
  <Column id=month totalAgg="All Months"/>
  <Column id=category totalAgg="All Categories"/>
  <Column id=sales_usd0k contentType=colorscale totalAgg=mean/>
  <Column id=num_orders_num0 contentType=colorscale scaleColor=red totalAgg=mean/>
  <Column id=aov_usd2 contentType=colorscale scaleColor=blue totalAgg={orders_by_category[0].aov_usd2}/>
</DataTable>
```

CleanShot 2024-01-24 at 13 15 04@2x

## Names for props

- totalRow - for DataTable
- totalAgg - for Column
- totalFmt - for Column

## Aggregate options

- sum (default)
- mean
- median
- min
- max
- count
- countDistinct
- Any other string, which will just print as the string

## Formatting of total row

- If its a standard aggregate, follow the formatting in the column
- If it is a count or countDistinct, don’t apply any formatting
- If it’s a custom string, don’t apply any formatting

## Aggregation logic

- currently using tidyJS. Perhaps would be more performant to do in duckdb, or only if someone wants the totals

### Checklist


archiewood:
To fix:
- formatting using shortcodes
- row numbers
- formatting raw values

hughess:
This looks awesome! Tested it with filter inputs and it works great.

Only scenario I don't think is covered is search:
- Should the total update based on the filtered search results?
- Should the total row disappear when the search returns zero results?

archiewood:
Good points

- On investigating a bit, updating totals based on the search results is not simple as the aggregates are computed in the query column metadata, which is not updated during search.
  - I think one path to this is using a textInput for search, which would update the query, and hence the totals
  - I think the easy and non confusing approach is probably to hide the totals when search is used.
- Hiding the totals when there are no results is easy, and the default if we do the above also

packages/component-utilities/src/getColumnExtents.js
```diff
@@ -1,24 +1,39 @@
...
+			? summarize({
+					count: n(columnName),
+					countDistinct: nDistinct(columnName),
+					min: min(columnName),
+					max: max(columnName),
+					median: median(columnName),
+					mean: mean(columnName),
+					sum: sum(columnName)
+			  })
+			: summarize({ count: n(columnName), countDistinct: nDistinct(columnName) })
```
ItsMeBrianD, line 23: Is this going to run all aggs for all numeric columns, regardless if they are actually displayed?

packages/component-utilities/src/tests/columnUnitSummary.spec.js
```diff
@@ -30,11 +34,49 @@ describe('getColumnUnitSummary', () => {
...
+			mean: 3,
+			sum: 9,
 			min: 1,
 			max: 5,
 			median: 3,
 			maxDecimals: 0,
 			unitType: 'number'
 		});
 	});
+	it('should only return count and countDistinct when the data is not numeric, and other values should be undefined', () => {
```
ItsMeBrianD, line 48: Tests 🚀

ItsMeBrianD (approved):
LGTM

packages/component-utilities/src/getColumnExtents.js
archiewood: Yes - it was already calculating the median, min & max for all columns as we use that information for automatically formatting column data.
I've added `count`, `countDistinct`, `mean` & `sum` to the same call.

I'm unclear on performance implications but I assume they will not be significantly more expensive that the existing functions

PR #1521, opened 2024-01-24
Bypass fully null bool column

csjh:
### Description

Converts `Bool` columns to `Float64` columns if they're entirely null to bypass `apache-arrow` serialization bug

### Checklist


archiewood:
LGTM

mcrascal (approved):

PR #1522, opened 2024-01-25
Upgrades for USMap

hughess:
### Description
This PR adds several upgrades to the `USMap` component:
- Adds Hawaii and Alaska to the map
- Expands the size of the map and fixes sizing issues on mobile
- Removes NaNs from tooltips
- Adds support for custom color palettes with the `colorPalette` prop (accepts an array of colors)
- Legend support (use with `legend=true`)
- Filtered legend support (use with `filter=true`)

This example uses CSS color names, but this prop accepts hex codes, RGB, HSL:

```html
<USMap
    data={demoData}
    state=name
    value=population
    colorPalette={['maroon','white','navy']}
    legend=true
    filter=false
/>
```

CleanShot 2024-01-24 at 21 04 45@2x

## Other Potential Changes

- Legend position control
- Support for projections
- Modifications to support arbitrary geoJSONs for other map regions

### Checklist


archiewood:
Do we need to add to docs for the new props?

hughess:
@archiewood we do - I've got it on my list!

packages/core-components/src/lib/unsorted/viz/map/_USMap.svelte
```diff
@@ -29,6 +29,12 @@
 
 	export let fmt = undefined;
 
+	export let filter = true;
```
archiewood, line 32: my sense is that this should be false by default

archiewood (approved):

PR #1523, opened 2024-01-25
Support multi axis for LineChart and BarChart

hughess:
### Description
This PR adds support for multiple y axes for `LineChart` and `BarChart`, including the ability to change the series type of the y2 axis through the `y2SeriesType` prop.

Previously, multiple y axis support was only available for line charts, and both axes had to be lines. Now it is possible to mix and match across lines, bars, and scatter.

### Color Customization
You can pass in a custom `colorPalette`, which will color the series as well as the axes by default. This can be turned off by setting `yAxisColor` and `y2AxisColor` to `false`.

#### Example

```html
<BarChart
    data={orders_by_month} 
    x=month 
    y=sales_usd0k 
    y2=num_orders_num0 
    y2SeriesType=line 
    colorPalette={['rgb(110,117,176,0.8)','rgb(37,91,161)']}
/>
```
CleanShot 2024-01-24 at 22 50 58@2x

```html
<BarChart
    data={orders_by_month} 
    x=month 
    y=sales_usd0k 
    y2=num_orders_num0 
    y2SeriesType=scatter 
    colorPalette={['rgb(110,117,176,0.8)','rgb(37,91,161)']}
/>
```

CleanShot 2024-01-24 at 22 52 26@2x

## To Do
- Test on Windows and multiple browsers
- Check that y2 does not work on charts where `swapXY=true`

### Checklist


archiewood (approved):
Looks great

Confirming that it does not work with swapXY

though it does just show a blank space rather than an error

CleanShot 2024-01-25 at 17 49 12@2x

hughess:
There may be some legitimate use cases for horizontal - like this chart I could see being horizontal as well:
CleanShot 2024-01-25 at 23 23 10@2x

But possibly tricky to figure out at the moment. Might just try to trigger an error instead

PR #1526, opened 2024-01-25
feat: add environment variable substition to source queries

ItsMeBrianD:
### Description

Adds interpolation for variables in source queries using `${}` syntax

"Advanced" Connectors have the option to disable this functionality when reading files (it is done automatically for "simple" connectors)

All variables must be in the environment as `EVIDENCE_VAR__[var name]` to ensure that they were deliberately set.

e.g.

```sql
SELECT * FROM MY_TABLE
WHERE client_id = ${client_id}
```

```bash
EVIDENCE_VAR__client_id=5 pnpm run sources
```

```sql
-- actually runs
SELECT * FROM MY_TABLE
WHERE client_id = 5
```

### Checklist


archiewood:
We may consider scoping this syntax?

For example:


```sql
SELECT * FROM MY_TABLE
WHERE client_id = ${env.client_id}
```

or 

```sql
SELECT * FROM MY_TABLE
WHERE client_id = ${var.client_id}
```


This might make it easier to distinguish variables from other queries if we were to add query chaining in the future?

```sql
SELECT * FROM ${my_other_query}
WHERE client_id = ${env.client_id}
```

ItsMeBrianD:
> This might make it easier to distinguish variables from other queries if we were to add query chaining in the future?

I think this may merit a larger discussion; query chaining will have to be (or rather, really should be I think) implemented at the connector level, rather than by Evidence itself, because the rules for doing so will vary based on the query syntax (e.g. how would query chaining work for yaml-based queries). Simple raw variable interpolation is safer and should be pretty portable.

archiewood:
Sounds good!

packages/plugin-connector/src/data-sources/sub-source-vars.js
```diff
@@ -0,0 +1,39 @@
...
+				const name = k.substring(envPrefix.length);
+				const value = v?.toString();
+				return [name, value];
+			})
+	);
+	let output = queryString;
+	const regex = RegExp(/\$\{(.+?)\}/, 'g');
+	let match;
+	while ((match = regex.exec(queryString)) !== null) {
+		if (match[1] in validVars) output = output.replace(match[0], validVars[match[1]]);
```
archiewood, line 31: This appears to throw
> │ > tsc --noEmit -p tsconfig.json && parcel build
│ src/data-sources/sub-source-vars.js(33,64): error TS2769: No overload matches this call.
│   The last overload gave the following error.
│     Argument of type 'string | undefined' is not assignable to parameter of type '(substring: s…
│       Type 'undefined' is not assignable to type '(substring: string, ...args: any[]) => string…

packages/plugin-connector/src/data-sources/sub-source-vars.js
```diff
@@ -0,0 +1,39 @@
...
+		else
+			console.warn(
+				`Missed substition for ${match[0]}, do you need to set EVIDENCE_VAR__${match[1]}?`
+			);
+	}
+
+	return queryString;
```
archiewood, line 38: Should this return `output`?

ItsMeBrianD:
Added a manual escape for users as well; `${variable}` gets replaced with the variable value, `$${variable}` gets replaced with `${variable}`

archiewood (approved):
Tested, works great for me!

PR #1529, opened 2024-01-26
Update type information to be more in line with USQL

csjh:
### Description

Changes type information from `_evidenceColumnTypes` to type inference based on column's JS type
-> also adds slightly better `fmt` error handling when the value column is string/bool

Doesn't rip out `_evidenceColumnTypes` in its entirety *yet* since it's still used in some places but probably worth expanding on later

### Checklist


ItsMeBrianD (approved):

archiewood:
Am I right in feeling like this should be tested pretty exhaustively?

Issue #1219, opened 2023-09-29
Add a default prop to chart components like 'empty_set=error | warn | pass'


gwenwindflower:
## Feature Description

It would be helpful to have a prop available optionally on all components, that would let you override and customize the `build` vs `build:strict` behavior of that component. I think it would default to the current behavior `error`, but offer new options like `warn` or `pass`.

## Goal of Feature

We often have queries and charts that are fairly specific or only return sparse results over large time ranges, etc -- often in CI or dev these will return no results when running over a limited set of data for cost+speed. So we have some things we expect and want to fail (revenue on the executive dashboard) if empty, which works with `build:strict` now, but we may have several charts that don't fit this model, for instance analyzing experiments that may not be in the test dataset.

## Current Solution / Workarounds

Right now these have to be wrapped in an `#if` block in the template, it would be DRYer, faster, and cleaner I think to have a prop you can pass to achieve this behavior as a standard.

## Alternatives

You could try ensuing the datasets never come up empty with some sort of backwards cacheing logic where Evidence can generate subset fixtures for you that populate the charts for CI and testing.

mcrascal:
Empty states that result from inputs and filters should also be handled so they don't error.

mcrascal:
- During SSR, empty set is treated as an error, unless empty_state = warn | pass, including an empty set that is the result of a filter default during ssr 
- Empty state as the result of an input being changed should not show an error state, rather an empty component state

archiewood:
@gwenwindflower @aaronsteers - shipped!

PR #1531, opened 2024-01-27
Add empty state to components

hughess:
### Description
Proof of concept, but may be close to ready.

Adds an `emptySet` prop to components as described here: #1219 

`emptySet` can be:
- `error`: component renders on page as an error, error is thrown, build:strict will be blocked
- `warn`: component renders to new empty state (appearance TBD) and warning sent to console, builds not blocked
- `pass`: component renders to new empty state (appearance TBD) and no warnings sent

Breaks the empty dataset logic out of `checkInputs.js` and into `isEmptyDataset.js`

Also wraps most component script tag code in a `if(!isEmpty)` conditional.

### Notes
Similar approach to our current error handling in components.

I think checks and error handling in components might be easier if they accepted a string for query name rather than the data object, then used data loading components to retrieve the actual data.

### Work remaining if this approach works
- Create empty state design for each component type
- Add to:
   - FunnelChart
   - SankeyDiagram
   - USMap
   - ReferenceLine
   - ReferenceArea

### Checklist

### DIFF
--- a/packages/component-utilities/src/checkInputs.js
+++ b/packages/component-utilities/src/checkInputs.js
@@ -17,9 +17,7 @@ export default function checkInputs(data, reqCols, optCols) {
 				'}'
 		);
 	} else if (data[0] === undefined || data.length === 0) {
-		throw Error(
-			'Dataset is empty: query ran successfully, but no data was returned from the database'
-		);
+		return;
 	}
 
 	// Check if data warehouse returned an error
@@ -27,7 +25,9 @@ export default function checkInputs(data, reqCols, optCols) {
 		throw Error('SQL Error: ' + data[0]?.error_object?.error?.message);
 	}
 
+	
 	if (reqCols != undefined) {
+		
 		if (!(reqCols instanceof Array)) {
 			throw Error('reqCols must be passed in as an array');
 		}
@@ -82,6 +82,7 @@ export default function checkInputs(data, reqCols, optCols) {
 			}
 		}
 	}
+
 	// IDEAS:
 	// Trigger a function call when error is caught - that function somehow sends us to the Error chart component
 	// rather than letting the rest of the current component file continue running?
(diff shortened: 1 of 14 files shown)


ItsMeBrianD:
I would recommend consolidating the empty logic into `QueryLoad`; that will make this available to all viz components (even if they don't accept the props at the moment)

In `QueryLoad.svelte`
```svelte
<script>
	import { onDestroy } from 'svelte';
	import { Skeleton } from '../skeletons';
	/** @type {import("@evidence-dev/query-store).QueryStore | unknown}*/
	export let data;

	$: if (data?.__isQueryStore) {
		data.fetch(); // Somebody wants this to load. Without this the query builder features don't work
		unsub();
		unsub = data.subscribe((v) => {
			_data = v;
		});
	}

	let unsub = () => {};

	let _data;
	onDestroy(unsub);
</script>

{#if !data || !data?.__isQueryStore}
	{#if !data?.__isQueryStore && isEmptySet(_data) && $$slots.empty }
	    <!-- handle case where data is not a query store but is also empty -->
		<slot name="empty"/>
	{/if}
	<!-- Not a query store, nothing to be done -->
	<slot loaded={data} />
{:else if !_data || (!_data?.loaded && !_data.error)}
	<!-- Data is loading -->
	<slot name="skeleton">
		<div class="w-full h-64">
			<Skeleton />
		</div>
	</slot>
{:else if _data.error && $$slots.error}
	<slot name="error" />
{:else if isEmptySet(_data) && $$slots.empty}
	<slot name="empty"/>
{:else}
	<slot loaded={_data} />
{/if}

```

hughess:
@mcrascal I think you mentioned you have an empty state design for another component - could you share that if you think it makes sense to apply it to charts as well?

mcrascal:
@hughess it's pretty minimal -- happy to adjust dimension grid to match whatever you land on here as well. 

https://github.com/evidence-dev/evidence/blob/a4947e73e656650551fa715cb0b17a6ca08f8ed1/packages/core-components/src/lib/molecules/dimension-grid/DimensionCut.svelte#L124

hughess:
Closing in favour of #1624

PR #1533, opened 2024-01-28
Fix getcompleteddata null treatment

hughess:
### Description
If the `x` column in `getCompletedData` started with a value of `null`, the function would throw an error, breaking the page. This PR adjusts the function to check for the first non-null value in the column. If the column is entirely null, a different error message is displayed.

#### Before
CleanShot 2024-01-28 at 16 23 22@2x

#### After
CleanShot 2024-01-28 at 16 23 34@2x

### Checklist


ItsMeBrianD (commented):
Implementation looks good to me; probably needs a test to cover this case. Happy to help write that if you'd like!

packages/component-utilities/src/tests/getCompletedData.spec.js
```diff
@@ -83,19 +83,19 @@ const testSuite = (opts) => {
...
+			// 		case 'number':
+			// 			expect(typeof data[0][keys.x]).toEqual('number');
+			// 			break;
+			// 		case 'date':
+			// 			expect(data[0][keys.x]).toBeInstanceOf(Date);
+			// 			break;
+			// 		default:
+			// 			expect(typeof data[0][keys.x]).toEqual('string');
+			// 			break;
+			// 	}
```
ItsMeBrianD, line 98:
```suggestion
```

ItsMeBrianD (approved):
LGTM

PR #1534, opened 2024-01-28
Less verbose error messages for charts

hughess:
### Description
Currently if you have errors in a chart on your page, you will see large error messages printed to the terminal. 

This PR shortens the error message, adds the chart type, and adds red font formatting to terminal output.

If we need the additional info, it may be worth adding a `logLevel` flag to components to allow larger messages to be posted.

#### Before
CleanShot 2024-01-28 at 16 37 28@21x

#### After
CleanShot 2024-01-28 at 16 45 17@2x

### Checklist

### DIFF
--- a/packages/core-components/src/lib/unsorted/viz/core/_Chart.svelte
+++ b/packages/core-components/src/lib/unsorted/viz/core/_Chart.svelte
@@ -992,7 +992,7 @@
 			});
 		} catch (e) {
 			error = e.message;
-			console.error(e);
+			console.error('\x1b[31m%s\x1b[0m', `Error in ${chartType}: ${e.message}`);
 			// if the build is in production fail instead of sending the error to the chart
 			if (strictBuild) {
 				throw error;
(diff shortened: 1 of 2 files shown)


packages/core-components/src/lib/unsorted/viz/core/_Chart.svelte
```diff
@@ -992,7 +992,7 @@
 			});
 		} catch (e) {
 			error = e.message;
-			console.error(e);
+			console.error('\x1b[31m%s\x1b[0m', `Error in ${chartType}: ${e.message}`);
```
ItsMeBrianD, line 995:
```suggestion
                         const setTextRed = '\x1b[31m%s\x1b[0m'
			console.error(setTextRed, `Error in ${chartType}: ${e.message}`);
```

The escape sequence here is a fairly "magic" string, if we put it into a variable it's easier to see what is happening at a glance

ud3sh (approved):

PR #1539, opened 2024-01-30
Change file download names to query names

hughess:
### Description
Previously downloads of data or images from components would be called `evidence-download` or `evidence-chart`.

This PR makes the file name the same as the query name in the component and appends the timestamp in the local timezone.

`evidence-download.csv` --> `my_query_2024-01-28T07-35-22.csv`

### Checklist


packages/component-utilities/src/echartsCanvasDownload.js
```diff
@@ -23,7 +23,13 @@ export default (node, option) => {
...
 	});
 
-	download(src, 'evidence-chart.png');
+	const date = new Date();
+	const localISOTime = new Date(date.getTime() - date.getTimezoneOffset() * 60000)
+		.toISOString()
+		.slice(0, 19)
+		.replaceAll(':', '-');
+
+	download(src, (option.queryID ?? 'evidence-chart') + `_${localISOTime}.png`);
```
archiewood, line 32: Potentially cool behaviour would be to name the image after the chart title if it is defined

archiewood (approved):

packages/component-utilities/src/echartsCanvasDownload.js
hughess: Oh that's a great idea

hughess: added that in

PR #1542, opened 2024-01-30
Preview Command

### LINKED ISSUE #259: "Serve" command
**Feature Description**

I know I can run a live preview with the `dev` command and I can build the project with `build`, but is there a `serve` command I can use to just host the build artifacts?

**Goal of Feature**

In cases where we just want to pre-build all pages and then render them in a non-live capacity, this would give an alternative to `dev` which I believe is compiling on-demand.

**Current Solution**

Simply run `dev` and wait for pages to load the first time, or run `build` and find another custom path to host those files.

**Alternative Ideas**

n/a

**Examples/Screenshots**

n/a

**Are you interested in contributing this feature?**

Sure. 😄

archiewood:
### Description

- This adds a preview command to evidence which allows you to view the built version of a site locally
- This should *not* be used as a production server, [see vite docs](https://vitejs.dev/guide/cli.html#vite-preview)

This closes #259

**Note: this will require an update to the template `package.json` to expose to users**

### Checklist


csjh:
I don't think `vite preview` would actually serve the static output directory here, I think it would actually run the app as though it was an `adapter-node`. I usually use `npx serve build` to preview built projects but that has the annoyance of adding a dependency

archiewood:
> I usually use npx serve build to preview built projects but that has the annoyance of adding a dependency
@csjh yeah I wanted to use vite so that we didn't have to do that - I normally use `http-server` but same issue

> I don't think vite preview would actually serve the static output directory here
what makes you think that is the case? this appeared to work for me locally

csjh:
I think it should still work, but it would be more akin to dev mode than a preview of a built project. They might've changed it since, but a quick test would be to delete the build directory and see if it still works - if it does, then it's previewing as a non-static project.

archiewood:
I did an alternative test which was to change the dev project, and re-run `preview`

The served site did not include the changes

csjh:
Yeah that would stay the same because in reality there's a *second* build folder called `.svelte-kit` which contains the "raw" build output which `adapter-static` then adapts into static build output into the `build` folder

packages/evidence/cli.js
```diff
@@ -314,4 +314,27 @@ prog
...
+	.action((args) => {
+		if (args.debug) {
+			process.env.VITE_EVIDENCE_DEBUG = true;
+			delete args.debug;
+		}
+
+		const flatArgs = flattenArguments(args);
+
+		// Run svelte kit dev in the hidden directory
+		const child = spawn('npx vite preview --outDir build --port 3000', flatArgs, {
```
archiewood, line 329: @csjh that's why I'm explicitly passing the --outdir flag

When I delete the `build` directory, this process throws

archiewood:
Perhaps I should handle this error a bit better though, its not exactly a nice error

```
error when starting preview server:
Error: The directory "build" does not exist. Did you build your project?
    at preview (file:///Users/archie/Projects/evidence/node_modules/.pnpm/vite@4.3.9/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:63861:15)
    at async CAC.<anonymous> (file:///Users/archie/Projects/evidence/node_modules/.pnpm/vite@4.3.9/node_modules/vite/dist/node/cli.js:865:24)
```

packages/evidence/cli.js
csjh: Huh, didn't realize vite supported them like that, nice

hughess:
There was a page I had to comment out in test-env to get my local build to work, but the preview command worked well!

hughess (approved):

PR #1544, opened 2024-01-31
Dimension-grid

mcrascal:


packages/core-components/sources/faker-demo-source/flights.yaml
```diff
@@ -0,0 +1,35 @@
...
+    category: airline
+    item: airline 
+    targetField: name
+  fare: 
+    category: number
+    item: float 
+    options:
+      - max: 10000
+      - min: 2
+      - fractionDigits: 2
```
ItsMeBrianD, line 19: If this is supposed to be a single options object the `-` shouldn't be on these lines

archiewood (approved):

PR #1548, opened 2024-02-01
Issue-1547 Enable automatic PR creation by depedabot

ud3sh:
### Description
* This will enable depedanbot to create automatic PR for detected vulnerabilities https://github.com/evidence-dev/evidence/security/dependabot
* This should help with `CVE detection -> Fix` time window as it would reduce the need for devs to scan the vulnerabilities list as well as reduce time to investigate the packages that need to be patched.

### Checklist


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
```
ud3sh, line 3: CleanShot 2024-02-01 at 10 54 16@2x
covered under `npm`

ItsMeBrianD (approved):

PR #1549, opened 2024-02-01
Fix query viewer table column order

hughess:
### Description
When the query viewer received numeric columns, it was displaying content in the wrong order due to a mismatch in the data received by the columns of the query viewer data table, and the cells of that table. This PR gets both to use the same data/order.

@archiewood this solves the problem you ran into with the pivot data

#### Before
CleanShot 2024-02-01 at 10 48 38@2x

#### After
CleanShot 2024-02-01 at 11 09 29@2x

### Checklist


archiewood:
So now the category is being placed as last column in the table. I'm not sure that's what DuckDB does by default?

CleanShot 2024-02-01 at 12 46 07@2x

hughess:
that's the same as it was ordering before, but the query viewer was putting the wrong data in the columns. This change just gets the right data in the columns.

If the column order is wrong, i'm assuming that would be coming from the query store, but I might be wrong - @ItsMeBrianD do you know?

ItsMeBrianD:
> If the column order is wrong, i'm assuming that would be coming from the query store, but I might be wrong - @ItsMeBrianD do you know?

It either comes from `_evidenceColumnTypes` or the metadata query from duckdb, I'm not sure where it would get jumbled though

hughess:
oh wait sorry i think i do know - i'm pretty sure the column summary is where it happens.

I'll confirm, but if that's the case I assume it's not a bad fix

hughess:
Just confirmed it's `getColumnSummary` - when it creates the object, if the properties are numeric, it sorts by them

ud3sh (approved):

PR #1582, opened 2024-02-09
Feat/hive partitions

ItsMeBrianD:
### Description

This adds a `partitions` field to `connection.yaml` that lets users create a map of query names to column names that duckdb will attempt to create partitions for

### Checklist


archiewood:
Very cool, have it working on a test dataset

Couldn't use it for my production dataset:

Needs to escape names of partition keys that could have `/` characters in the string

for example

Person | Band
--|--
Anna | The Beatles
Bob | AC / DC

outputs

```
Invalid Error: Error: ENOENT: no such file or directory, mkdir 'static/data/band_data/favorite_bands/4eb01c8955e8746f3739bbdea53882f4/favorite_bands/band=AC / DC
```

archiewood:
Also the partition source files are no longer cleaned out when you run `npm run sources`

archiewood:
I think this partition with `/` in string is a DuckDB bug - I have reproduced and reported

https://github.com/duckdb/duckdb/issues/10575

In the meantime, perhaps we can tell users to remove any / characters from their columns they want to partition by.

archiewood:
tests are failing - perhaps the new `rm` command is being counted in the overall number?

ItsMeBrianD:
> tests are failing - perhaps the new rm command is being counted in the overall number?

This is the issue

PR #1598, opened 2024-02-14
Additional chart override options

hughess:
### Description

These options can be applied to any standard cartesian charts in Evidence:
- `renderer` - "canvas" (default) or "svg"
```
renderer=svg
```
- `seriesOptions` - ECharts options to be applied to all series in a chart. This will loop through the series to apply the changes (e.g., setting a border around data points)
```
seriesOptions={{
   itemStyle: {
      borderWidth: 4,
      borderColor: '#c8c8c8'
   }
}}
```
This option skips reference lines and reference areas when looping through series - this avoids unintended consequences, but also restricts the ability to edit those items. We may need to offer another way to access them. In the meantime, they are accessible through `echartsOptions` by specifying `series` changes to the correct series index locations.

- `seriesColors` - option to set a specific colour for a specific series, specified by name
```
seriesColors={{
   'Canada': 'red',
   'UK': 'navy'
}}
```

CleanShot 2024-02-13 at 22 57 39@2x

```html
<LineChart 
  data={simpler_bar} 
  x=year 
  y=value 
  series=country
  seriesColors={{
    'Canada': 'red',
    'US': 'navy'
  }}
/>
```

This PR also adjusts the rendering timing in `echarts.js` to avoid reloading of the charts where possible. This is a short-term solution that should be cleaned up in the future.

### Checklist

### To do

- Add docs


hughess:
@ItsMeBrianD @csjh would be great to get your eyes on this one - in particular the changes to `echarts.js` 

I shifted some things around and added a delay on initial load to avoid multiple requests. I initially tried debounce, but that seemed to cause issues with charts hooked up to input components.

packages/component-utilities/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@evidence-dev/component-utilities",
-	"version": "2.2.0",
+	"version": "2.2.1",
 	"description": "",
```
ItsMeBrianD, line 4: Why did versions get updated in this PR?

packages/component-utilities/src/echartsCanvasDownload.js
```diff
@@ -7,14 +7,71 @@ export default (node, option) => {
 
 	const chart = init(node, 'evidence-light', { renderer: 'canvas' });
 
-	option.animation = false;
+	option.config.animation = false;
+
+	chart.setOption(option.config);
+
+	// Series Color override
+	const applySeriesColors = () => {
```
ItsMeBrianD, line 15: Why is this being duplicated? Could these functions be pulled out into a shared file?

ItsMeBrianD (changes requested):
Too much duplicated code here; happy to pair and/or take a stab at pulling it out into a shared set of functions to reduce that.

Also unclear why versions were updated in this package, we likely want to revert those changes to avoid interfering with the release.

Otherwise looks good

ItsMeBrianD (approved):

archiewood:
This is missing docs

hughess:
Ah damn, completely missed that. Will add them!

PR #1601, opened 2024-02-14
Add Heatmap to legacy

hughess:
### Checklist


ItsMeBrianD:
Should this be pointing to legacy branch instead of `next`?

archiewood:
@ItsMeBrianD presumably this needs a changeset too?
Or do we want to manually version legacy changes?

archiewood:
Dev workspace is going to fail because USQL adds build:sources step which does not exist in legacy

ItsMeBrianD:
> Or do we want to manually version legacy changes?

We can try changesets but we might want to fall back to manual versioning to ensure we don't conflict

PR #1605, opened 2024-02-14
use `Date.now()` instead of `query.hash` for file path

### LINKED ISSUE #1437: Universal SQL: Filter interaction not triggering on parquet source
## Steps To Reproduce

When opening [this page](https://mdsinabox.com/nba/predictions/), the filter component does not allow the data set to be filtered.

image

This is hosted on Netlify, so it is possible that there is an issue between the browser and netlify that is causing the filter to fail.

The underlying code can be found [here](https://github.com/matsonj/nba-monte-carlo/blob/master/evidence/pages/nba/predictions/index.md).

### Environment

- Node version (`node -v`): 20.10.0
- npm version (`npm -v`): 10.2.3
- OS: Windows
- Browser: Chrome
- Database: n/a (USQL)

## Expected Behavior

Filter should allow you to select a specific team and view their future games.

## Other notes

Works totally fine in dev mode on local (in a codespace).

csjh:
### Description

Switches to `Date.now()` in an attempt to fix #1437, as leading theory for cause is faulty caching. Post-change, parquet file paths are now immutable, so caching shouldn't be an issue.

### Checklist

### DIFF
--- a/packages/plugin-connector/src/data-sources/get-sources.js
+++ b/packages/plugin-connector/src/data-sources/get-sources.js
@@ -111,7 +111,7 @@ export const getSources = async (sourcesDir) => {
 				// queries: queries
 			};
 		})
-	).then((r) => /** @type {Exclude<typeof r[number], false>[]} */ (r.filter(Boolean)));
+	).then((r) => /** @type {Exclude<typeof r[number], false>[]} */(r.filter(Boolean)));
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/plugin-connector/src/data-sources/get-sources.js
```diff
@@ -210,6 +210,19 @@ export async function cleanParquetFiles(dataDir, hashes) {
 					await fs.rm(path.join(queryPath, resultHash), { recursive: true, force: true });
 				}
 			}
+
+			if (!sourceHashes[queryName]) continue;
+			const queryHashPath = path.join(queryPath, /** @type {string} */ (sourceHashes[queryName]));
+			const timestamps = await fs.readdir(queryHashPath);
+			const latest = Math.max(...timestamps.map((x) => Number(x))).toString();
```
ItsMeBrianD, line 217: This will break if run against a project that still has hashes

csjh: TIL `Math.max(1, 2, NaN)` is `NaN`... thanks Javascript

ItsMeBrianD: https://www.destroyallsoftware.com/talks/wat

ItsMeBrianD (approved):

packages/plugin-connector/src/data-sources/get-sources.js
archiewood: @ItsMeBrianD this was so good

PR #1612, opened 2024-02-16
Fix query scoping in dev workspace

hughess:


ItsMeBrianD:
@hughess why is this change needed?

hughess:
@ItsMeBrianD there are 2 orders tables available in the schema - 1 in needful_things and 1 in ecommerce. It seems to pick up needful things initially, but will switch to ecommerce and error out unpredictably. In another PR, the deployment preview was showing a broken page as a result

hughess:
@ItsMeBrianD this seems to happen on the dropdown page in the dev workspace project as well

ItsMeBrianD (approved):

PR #1613, opened 2024-02-16
Fix echarts staggering

csjh:
Removes the debounce workaround in `echarts.js` in favour of removing debounced calls during component initialization


ItsMeBrianD:
Missing changeset

csjh:
merging into #1598 who already has a changeset for this

ItsMeBrianD (approved):

PR #1624, opened 2024-02-22
Empty state for components

hughess:
### Description
Adds an empty state to components, which can be controlled via the `emptySet` prop as described here: https://github.com/evidence-dev/evidence/issues/1219

`emptySet` can be:
- `error` (default): component renders on page as an error, error is thrown, build:strict will be blocked
- `warn`: component renders to new empty state and warning sent to console, builds not blocked
- `pass`: component renders to new empty state and no warnings sent

When `warn` or `pass`, another prop `emptyMessage` is also available, letting you customize the text that appears for your users when a component receives empty data.

### Behaviour with Interactivity
The `error` empty state only considers the initial data load. When filters are changed on a page, causing a query to re-run, this will not trigger an error in components - it will trigger the new empty state. That means your users don't get errors if they find combinations of filter inputs that lead to empty datasets.

This is currently not configurable - filter input changes will not trigger errors - but can be added as an option if needed.

### Considerations
- We may want the new empty state design to take up similar space to the components on the page to avoid layout shift

### Bonus
- This pulls SQL query errors into component errors again, which should avoid blank components on SQL errors
- New BigValue error format which sits inline with other BigValues
- SankeyDiagram now follows the same "USQL wrapped component" approach as the rest of the components
- BigValue and Value errors now print to the console to be consistent with charts

### Checklist


packages/component-utilities/src/isEmptyDataset.js
```diff
@@ -0,0 +1,7 @@
+export default function isEmptyDataset(data) {
+	if (data === undefined || data[0] === undefined || data.length === 0) {
```
ItsMeBrianD, line 2: We may want to have to guard against nulls or falsey values
```suggestion
	if (!data || !data[0] || !data?.length) {
```

PR #1634, opened 2024-02-24
Add basic server loading page

hughess:
### Description
Adds a loading page immediately after server start, which redirects to the Evidence app once it is ready.

**Note: this branch started from the hmr error branch, so probably easier to start fresh if we decide to implement this**

Uses a proxy server on port `n+1` to achieve this.

E.g., run `npm run dev` in your Evidence project:
 - It starts the dev server on port 3000 (if available, otherwise increment up by 1)
 - It serves a loading page found at `@evidence-dev/evidence/static/loading.html` on port 3001
 - Once the app is ready, it automatically redirects from port 3001 to port 3000

CleanShot 2024-02-24 at 15 47 48

### New dependencies
- `get-port` - checks availability of ports and increments from 3000 up to 3005, then to random available port (rather than relying on vite to do this for us)
- `ora` - if we decide to go for enhanced terminal feedback
- `http-proxy` and `open` - for proxy server

### Concerns
- Potential flakiness of proxy server - it has felt solid in testing, but it's another layer where something could go wrong/get delayed
- Does is slow down the loading process significantly? A slight slowdown may be worth it in order to show loading feedback rather than blank browser window

### To do if this approach is feasible
- Set up loading page design/animation
- Consider a fade-in animation by default for evidence projects in dev mode - can coordiante with fade out animation of loading screen for a smooth transition
- Should the loading page be a skeleton evidence app?
   - Issue here would be needing to respect styling/layout customizations in a user's evidence project so they don't get a stock evidence project as loading screen, then redirect to styled app 
- [nice to have] Enhanced terminal feedback - we should be able to get a flag back when the app is loaded, which we could use to update the terminal logs (e.g., "starting evidence server" --> "evidence server running" 

### Example custom console feedback
Would need to reincorporate our typical console logs as well - this is a rough approach:

CleanShot 2024-02-24 at 17 20 00

### Checklist


archiewood:
I assume this is not ready?

hughess:
Closing in favour of #1685

PR #1637, opened 2024-02-25
Sidebar-sorting & Layout Options

mcrascal:
Adds four frontmatter options to control the sidebar: 

1. `sidebar_position` - lower number shows up first, within their parent directory 
2. `title` - sidebar label matches title, helpful for acronyms (UI, MRR etc.) 
3. `sidebar_link`- controls whether or not the page has a link in the sidebar. If the page is the index of a directory and this option is false, the title of the directory is still shown.  
4. `sidebar_badge` adds a green badge next to the title in the sidebar (want this for docs). 

Treatment of directories:

Setting the `sidebar_position` of the index page in a directory moves the directory relative to other directories. In combination with `sidebar_link: false`, you can use this to move directories that don't have a home page. 

Also adds new layout options, per #1638 

New +layout file:

```
<script>
	import '@evidence-dev/tailwind/fonts.css';
	import '../app.css';
	import { EvidenceDefaultLayout } from '@evidence-dev/core-components';
	export let data;
</script>

<EvidenceDefaultLayout {data}>
	<slot slot="content" />
</EvidenceDefaultLayout>

```

`EvidenceDefaultLayout` has props for:

* Overriding the project logo with a text title
* Overriding the project logo with an image
* Setting the width of the project to full width
* Never showing queries in prod - removes the show queries option from the drop down when deployed (leaves it during dev)
* Hiding the sidebar
* Github Repo
* Slack Community
* X Profile
* Built With Evidence - adds a small link to our homepage below the sidebar
* Algolia docs search settings 

Ultimately, I'd like to expose these props as options in evidence.config.


mcrascal:
closes #416

ItsMeBrianD (approved):
Quite a lot to look through here; seems like it makes sense

mcrascal:
@ItsMeBrianD 

> Quite a lot to look through here; seems like it makes sense

Ah my bad, this was previously two PRs. The sidebar API stuff, and the layout refactor.

archiewood:
Seems like this needs some work still?

archiewood (approved):

PR #1638, opened 2024-02-26
refactor layout and add layout props

mcrascal:
Off of #1637 

Refactors the base layout to a much smaller and easier to work with footprint. 

New +layout file:

```
<script>
	import '@evidence-dev/tailwind/fonts.css';
	import '../app.css';
	import { EvidenceDefaultLayout } from '@evidence-dev/core-components';
	export let data;
</script>

<EvidenceDefaultLayout {data}>
	<slot slot="content" />
</EvidenceDefaultLayout>
```

`EvidenceDefaultLayout` has props for: 
1. Overriding the project logo with a text title 
2. Overriding the project logo with an image 
3. Setting the width of the project to full width
4. Never showing queries in prod - removes the show queries option from the drop down when deployed (leaves it during dev)
5. Hiding the sidebar 
6. Github Repo
7. Slack Community
8. X Profile 
9. Built With Evidence - adds a small link to our homepage below the sidebar 

Ultimately, I'd like to expose these props as options in evidence.config.

CleanShot 2024-02-25 at 20 11 18@2x


archiewood:
Very cool!

If possible, I think any slack and github icons should be left of the kebab page menu.

I can also believe you might want to link to a more generic place.
- Your company marketing site
- Your data dictionary 
- etc

- Could have a more generic "link" icon for this purpose

packages/core-components/src/lib/organisms/layout/sidebar/Sidebar.svelte
```diff
@@ -273,4 +274,14 @@
 			{/each}
 		</div>
 	{/if}
+	{#if builtWithEvidence}
+		<div class="fixed bottom-0 text-xs py-2">
+			<a
+				href="https://www.evidence.dev"
+				class="bg-gradient-to-r inline-block text-gray-950 antialiased font-medium"
+			>
+				Built with Evidence</a
```
archiewood, line 283: I think there's an argument to put this at the bottom of the article, rather than tucked away in the sidebar

mcrascal:
CleanShot 2024-02-25 at 21 04 35@2x

packages/core-components/src/lib/organisms/layout/sidebar/Sidebar.svelte
mcrascal: See gif -- I like that it's visible at all times, though think bottom right might be better?  

CleanShot 2024-02-25 at 21 07 29

hughess:
This is sweet. Only other thing it may be worth considering is bringing a toggle for SQL queries out of the kebab menu - we've had a lot of people ask about how to show/hide their queries.

One other option for that is creating a "dev toolbar" which appears at the top of the page only in dev mode and includes anything that would help people when developing their project - SQL toggle, schema explorer, link to deploy, etc.

archiewood:
dev toolbar is nice, but we'd still have to have a show/hide queries button elsewhere

mcrascal:
I'm also interested in looking at a command k menu for dev controls. There's a few things that will end up in the header though, even in prod (dark mode toggle, arbitrary links etc.) where we may end up needing the kebab as the overflow for smaller screens. If you guys are good, can I get an approval and we can revisit this in another PR?

PR #1653, opened 2024-02-28
Immediate hydration on page load

csjh:
### Description

Adds a `backgroundFetch` function and `longQueryNotifier` to querystore

toast notification shows for 3 seconds after query takes longer than 1 second to execute:
image

### Checklist

### DIFF
--- a/packages/preprocess/src/process-queries.cjs
+++ b/packages/preprocess/src/process-queries.cjs
@@ -130,6 +130,7 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 							'${id}',
 							{
 								scoreNotifier,
+								longQueryNotifier,
 								initialData,
 								initialError,
 								noResolve: false,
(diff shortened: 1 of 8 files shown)


packages/preprocess/src/process-queries.cjs
```diff
@@ -261,6 +267,15 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 			}, 5000);
 		};
 
+		const longQueryNotifier = (info) => {
+			toasts.add({
+				id: Math.random(),
```
archiewood, line 272: is this what we want? Any issue with passing a random id?

archiewood:
This appears to be breaking the windows tests

csjh:
Seems like the tests are already failing in `features/a`

packages/preprocess/src/process-queries.cjs
csjh: Shouldn't be an issue, just meant to have some sort of differentiator between toasts

packages/preprocess/src/process-queries.cjs
```diff
@@ -261,6 +267,15 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
+				title: info.id,
+				message: \`Query \${info.id} is loading...\`,
+				status: 'info'
```
mcrascal, line 275: I think the behaviour we want on the toast isn't several toasts saying "query x is loading..." but rather just a single "loading..." toast that's present until all long-running queries are done.

mcrascal (changes requested):

PR #1668, opened 2024-03-01
Input components 

mcrascal:
Breaks input components out of 1569


ItsMeBrianD (approved):

csjh:
I'm just noticing/remembering I never concluded the [dropdown backwards compatibility fix](https://github.com/evidence-dev/evidence/pull/1477#discussion_r1481621082)

I'm leaning on the side of an `includeLabels` prop since it seems cleaner for everyone involved

archiewood:
how would that work?

csjh:
yup

archiewood:
Sounds like a good solution to me.

What do you think about warning a breaking change in the future?


> `${inputs.my_input}` will be deprecated in later versions of evidence. 
>
> Use `${inputs.my_input.value}` in combination with `includeLabels=true`

csjh:
Yeah makes sense to me

mcrascal:
@archiewood I'm inclined to just make the breaking change and announce it in the changelog

archiewood:
That's fine. We should definitely add a major bump in that case

PR #1685, opened 2024-03-05
Add server loading feedback

hughess:
### Description
Replacement for #1634

Adds a loading page immediately after server start, which redirects to the Evidence app once it is ready.

After 30 seconds, displays additional information.

Uses a proxy server on port n+1 to achieve this.

E.g., run npm run dev in your Evidence project:

It starts the dev server on port 3000 (if available, otherwise increment up by 1)
It serves a loading page found at @evidence-dev/evidence/static/loading.html on port 3001
Once the app is ready, it automatically redirects from port 3001 to port 3000

#### New dependencies:
get-port - checks availability of ports and increments from 3000 up to 3005, then to random available port (rather than relying on vite to do this for us)
ora - if we decide to go for enhanced terminal feedback
http-proxy and open - for proxy server

### Things to check
- This makes part of the dev command async

### To do
- Complete custom console feedback, or remove - at the moment it would require incorporating error messages as they are currently blocked due to the `pipe` setting
- Ensure new deps are installed in the correct places (currently workspace installs at the monorepo root)

### Checklist


packages/evidence/cli.js
```diff
@@ -180,7 +183,31 @@ prog
...
+		// Set a timeout to change the spinner after 1 minute
+		const timeoutId = setTimeout(() => {
+			// Stop the current spinner
+			spinner.stop();
+
+			// Start a new spinner or modify the existing one as per your requirements
+			spinner.text = chalk.cyan(
+				'Evidence is starting - this can take up to 2 minutes the first time.'
+			);
+			spinner.start();
```
ItsMeBrianD, line 203: I don't think you need to stop/start the spinner to update the text

packages/evidence/cli.js
```diff
@@ -180,7 +183,31 @@ prog
...
+		}, 30000); // 30 seconds
+
+		// Use get-port to find an available port starting from 3000
+		const port = await getPort({ port: [3000, 3001, 3002, 3003, 3004, 3005] });
```
ItsMeBrianD, line 207: This feels a little fragile, probably not a huge deal (who is running 6 Vite servers on 3000)

ItsMeBrianD (changes requested):
A few little things;

I'm not sure how I feel about this pattern, but I also can't really think of a better way to do this.
I'd like to think on it for a little bit but I think this will work

packages/evidence/cli.js
hughess: this falls back to a random port if none of those are available. Just ran into another scenario when playing around with this - I had 3001 occupied, but 3000 was open. So getPort picked 3000 and the proxy server picked 3001. I just fixed this up to cover that scenario

ItsMeBrianD (approved):
Dug into this yesterday, didn't come up with any feasible alternatives; I think we're good as is

hughess:
@ItsMeBrianD only other thing on this one is the dependencies - where is the right place to install them? They're in the root of the monorepo as workspace dependencies at the moment

ItsMeBrianD:
> @ItsMeBrianD only other thing on this one is the dependencies - where is the right place to install them? They're in the root of the monorepo as workspace dependencies at the moment

I believe they should be added as dependencies & peer dependencies for the evidence package

PR #1695, opened 2024-03-06
Fix bug navigating away from settings page

hughess:
### Description
There was a bug that would crash the page if you opened a connection's settings in the Settings page, then attempted to navigate back to a normal page.

This was cause by svelte's animation attempting to close the connection settings pane before navigating.

This PR adds `|local` to the animation, which solves the issue.

Also adds a border to the db settings panel

#### Before
CleanShot 2024-03-06 at 15 58 10

#### After
CleanShot 2024-03-06 at 15 59 23

### Checklist

### DIFF
--- a/.changeset/happy-ads-serve.md
+++ b/.changeset/happy-ads-serve.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Fix bug caused by navigating away from settings page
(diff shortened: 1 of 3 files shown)


mcrascal (approved):

csjh:
Definitely

hughess:
@csjh ok sweet I'll add that in here too

hughess:
I'm also getting an error by default when navigating to the sql console - happens in a fresh install as well:
CleanShot 2024-03-06 at 16 41 36@2x

PR #1700, opened 2024-03-07
Remove headless UI

mcrascal:
Removes headless UI dependency, moves the kebab over to shadcn, hides the kebab entirely in production when never show queries is true.


archiewood:
@mcrascal will this prevent people  printing to PDF when never show queries is true?

mcrascal:
It will remove the button that says "print pdf" but it will have no effect on people's ability to print the page. For our docs at least, I don't want to have a dropdown menu with one sort of useless item in it.

PR #1703, opened 2024-03-07
Fix table search when sorting

### LINKED ISSUE #1702: Search broken on DataTable when column is sorted
## Steps To Reproduce
1. Set up a DataTable
2. Search for an item to confirm it's working as expected
3. Refresh and click to sort a column and repeat the search. Search will return the wrong items

CleanShot 2024-03-01 at 15 32 11

## Workarounds
Search on unsorted table

hughess:
### Description
Fixes #1702 

#### Before
CleanShot 2024-03-01 at 15 32 11

#### After
CleanShot 2024-03-07 at 12 49 08

### Checklist

### DIFF
--- a/.changeset/calm-impalas-rescue.md
+++ b/.changeset/calm-impalas-rescue.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Fix table search when sorting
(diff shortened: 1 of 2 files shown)


archiewood (approved):

ItsMeBrianD (approved):
This looks okay to me, but I am not super familiar with Fuse

csjh:
I think only downside here is that fuse has to re-index the whole dataset every time it changes, which won't matter when we switch to using duckdb searching instead, so I think this is fine in the meantime

PR #1709, opened 2024-03-08
remove right margin on header container

mcrascal:
alignment nit


mcrascal:
closing in favour of #1770

PR #1716, opened 2024-03-08
Move SDK into monorepo

ItsMeBrianD:
This PR does (should) not include any functional changes to the code, and should only include a copying of the `sdk` package, along with some minor version tweaks for compatibility.

### Description

The Evidence SDK is the beginning of a rewrite of the "framework" component of Evidence.

It will replace the `preprocess`, `plugin-connector`, `query-store`, `universal-sql`, `db-commons` and `component-utility` packages, bringing everything into one package that is used to power Evidence.

Sections of the code that are "plugins" will remain split out (e.g. `tailwind`, `core-components`, all datasources, and eventually the example project), and have been sorted into folders accordingly).

The plan is to incrementally adopt and improve the `sdk` package to phase out the above packages, replacing them throughout the codebase with their SDK equivalants, which will appear in future PRs.

---

This PR also makes the `@evidence-dev/sdk` package publicly available as a preview, which can be used to use Evidence datasources and universal SQL in custom Sveltekit projects, presented as a vite plugin + svelte preprocessor. Refer to the [SDK Readme](https://github.com/evidence-dev/evidence/tree/integrate-sdk/packages/lib/sdk/README.md) for more information


archiewood:
looks like a sensible re-org conceptually.
The only thing that broke my brain expectations is that db-commons and db-orchestrator are in datasources.

packages/lib/sdk/src/build-dev/vite/virtuals/proxyStack.js
```diff
@@ -0,0 +1,37 @@
...
+	 * @type {Array<[string, Object]>}
+	 */
+	let stack = [];
+	/** @type {import("svelte/store").Writable<Record<string,Object>>} */
+	const write = writable({});
+
+	const flush = () => {
+		write.update((v) => {
+			for (const k of Object.keys(v)) delete v[k];
+			return Object.assign({}, v, ...stack.map((s) => s[1]));
```
csjh, line 18: Probably not a concern but this seems slow

packages/lib/sdk/src/plugins/components/loaders/file-loader.js
```diff
@@ -0,0 +1,125 @@
...
+	// Return relative filepaths
+	return output.map((p) => path.relative('.', p));
+};
+
+/**
+ * Reduce function to run on a svelte AST.
+ * Searches for the evidenceInclude declaration.
+ * @example
+ *  <script context="module">
+ *      export const evidenceInclude = true
```
csjh, line 46: should probably be specified that this is static analysis somewhere

packages/lib/sdk/src/plugins/datasources/SourceResultCache.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ *
+ * @param {string} sourceName
+ * @param {string} queryName
+ * @param {string} content
+ */
+export const addToCache = (sourceName, queryName, content) => {
+	const contentHash = hash(content);
+	if (!knownQueryHashes[sourceName]) knownQueryHashes[sourceName] = {};
```
csjh, line 48:
```suggestion
	knownQueryHashes[sourceName] ??= {};
```

packages/lib/sdk/src/types/virtuals.d.ts
```diff
@@ -0,0 +1,123 @@
...
+	export const ALL_QUERIES_CONTEXT_KEY = '__EVIDENCE_ALL_QUERIES';
+	export const ALL_QUERIES_MUTS_CONTEXT_KEY = '__EVIDENCE_ALL_QUERIES_MUTS';
+}
+
+declare module '$evidence/initUsql' {
+	const initUsqlPromise: Promise<void>;
+	export default initUsqlPromise;
+}
+
+declare module '$evidence/ssrHook.svelte.js' {
```
csjh, line 38: I really don't like the `.svelte.js` prefix when it has nothing to do with runes

packages/lib/sdk/src/types/virtuals.d.ts
```diff
@@ -0,0 +1,123 @@
...
+	export const getQueries: () => import('svelte/store').Readable<GetQueryResult>;
+}
+
+declare module '$evidence/updateQueryContext' {
+	type UpdateQueryContext = () => (layerQueries: Record<string, string>) => void;
+	const fn: UpdateQueryContext;
+	export default fn;
+}
+
+declare module '$evidence/static-assets' {
```
csjh, line 80: kebab case when everything else is camel case is weird

packages/lib/sdk/src/types/virtuals.d.ts
csjh, line 1: IMO doesn't make sense to have this many virtual files when most are just one or two exports

ItsMeBrianD:
@csjh we can address these comments in the future (probably worth logging an issue); this PR is just to migrate the code and restructure the repository

ItsMeBrianD:
@archiewood I would be okay with moving those into `lib`, but would we then want to move `component-utilites` into lib as well, or do those make sense where they are?

csjh:
Yeah just wanted to label some thoughts while going thru it

csjh (approved):

archiewood:
@ItsMeBrianD I take your point - I'm fine as is


I think the alternative would be:
- db-[stuff] -> lib
- tailwind,  component-utils -> lib
- core-components becomes top level (ie there is no ui directory)

ItsMeBrianD:
```
db-[stuff] -> lib
tailwind, component-utils -> lib
core-components becomes top level (ie there is no ui directory)
```

Could instead do:

```
db-[stuff] -> lib
component-utils -> lib
tailwind + core-components are the ui directory
```

This also reflects the intention of all packages in the `lib` directory wanting to get merged into the `sdk` over time

PR #1719, opened 2024-03-09
Fullscreen DataTable

csjh:
### Description

Adds `Fullscreen` component and uses it in `DataTable`

image

Basic `Fullscreen`-wrapped `BarChart`:
image

### Checklist


packages/core-components/src/lib/atoms/fullscreen/Fullscreen.svelte
```diff
@@ -0,0 +1,56 @@
...
+			}
+		};
+	}
+</script>
+
+<dialog class="w-[90vw] rounded-lg overflow-hidden relative" use:popup={open}>
+	<button class="absolute top-2 right-2 text-gray-500 hover:text-gray-700 focus:outline-none"
+		><Icon class="w-6 h-6" src={X} /></button
+	>
+	<slot />
```
archiewood, line 55: The y axis looks pretty tight to the edge of the modal in the screenshot of the barchart

Can we add a bit of x - padding?

archiewood (commented):
Looks great!

UI nit

archiewood (approved):

hughess:
This is really nice!

I just added a commit to include Column components support

2 issues I ran into:
1. If the column height is different than default, it can cut off the table. For now, I removed `overflow-hidden` from the `Fullscreen` class to enable scroll, but I'm not sure if that will cause issues. Here's an example of that situation (which I think will be pretty rare):

   CleanShot 2024-03-21 at 08 47 11

2. I got an error in `test-env` related to importing `Fullscreen`:
   CleanShot 2024-03-21 at 08 45 16@2x

csjh:
Ah yeah `overflow-hidden` should be fine to drop

I can't repro the `Fullscreen` test-env bug, could you try deleting `node_modules/.vite` and try again?

hughess:
@csjh that did the trick!

PR #1724, opened 2024-03-11
Fix: Arrays for Multiselect Dropdowns

archiewood:
### Description

This allows you to use outputs from mutli-dropdowns as DuckDB arrays.

In the case where nothing is selected, it will return an empty array

## From 
causes multiple values to be returned as a LIST type

`${inputs.category.value} → ['Cursed Sporting Goods', 'Sinister Toys']`

which means to filter a query, you need to use

`where list_contains(${inputs.category.value}, category)`

## To 
use an ARRAY type

`${inputs.category.value} → ('Cursed Sporting Goods', 'Sinister Toys')`

which means you could filter with

`where category IN ${inputs.category.value}`

which I believe is the intended behaviour

### Checklist

### DIFF
--- a/packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
+++ b/packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
@@ -67,9 +67,8 @@
 		if (typeof value === 'number' || typeof value === 'bigint' || typeof value === 'boolean')
 			return String(value);
 		if (value instanceof Date) return `'${value.toISOString()}'::TIMESTAMP_MS`;
-		if (Array.isArray(value)) 
-		return value.length === 0 ? `('')` :
-		`(${value.map((x) => jsToDuckDB(x)).join(',')})`;
+		if (Array.isArray(value))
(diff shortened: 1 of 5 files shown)


csjh (approved):
moved some stuff around, LGTM though

packages/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -66,15 +67,18 @@
...
 	}
 
 	function selectedValuesToInput() {
 		let values;
 		if (multiple) {
 			values = $selectedValues.map((x) => x.value);
-			values.toString = () => jsToDuckDB(values);
+			values.toString = () =>
+				values.length === 0
+					? `(select null where 0)`
```
archiewood, line 80: what happens here when there are no selected items in a multiselect?

will it throw in a `where category IN ${inputs.my_input.value}` situation?

csjh: No, I tested it with a few different queries in the CLI and it seemed like `(select null where 0)` was the only that worked consistently
`('')` fails when you're doing something like `select integer_id from table where integer_id in ('')` (from mismatched types)

PR #1736, opened 2024-03-13
Add maxWidth property, align title logo behaviour with docs, hide sidebar button on mobile when hiding sidebar, fix env var listing in settings 

mcrascal:


packages/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
```diff
@@ -56,14 +58,17 @@
...
+		{maxWidth}
+		{hideSidebar}
 		{githubRepo}
 		{slackCommunity}
 		{xProfile}
 		{algolia}
 	/>
 	<div
-		class={(fullWidth ? 'max-w-full ' : 'max-w-7xl ') +
+		class={(fullWidth ? 'max-w-full ' : maxWidth ? '' : ' max-w-7xl ') +
```
archiewood, line 69: is this the most natural way to introduce this prop?

I feel like maxWidth is mutually exclusive from fullWidth

archiewood: ignore this, that is the behaviour you have added

PR #1763, opened 2024-03-19
Add default `ButtonGroup` selected item

csjh:
### Description

Adds a `default` prop to `ButtonGroupItem` for adding a default selected button

### Checklist

### DIFF
--- a/.changeset/red-beans-cheat.md
+++ b/.changeset/red-beans-cheat.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': minor
+---
+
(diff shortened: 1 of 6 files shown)


archiewood (approved):

hughess:
@csjh it looks like the defaults aren't appearing on the button range page in `test-env`, but they're using `defaultValue` - is that the wrong thing to use?

csjh:
Yeah this doesn't add `defaultValue`

hughess:
ok thanks, i'll adjust those pages

PR #1770, opened 2024-03-21
remove right margin from default layout header

mcrascal:
removes the margin on the right side of the kebab


hughess (approved):

mcrascal:
@ItsMeBrianD linter is failing on query store? 🤔

ItsMeBrianD:
Checking this now, not sure why it's just now popping an issue up

PR #1781, opened 2024-03-23
Speed up tests

archiewood:
### Description

This PR aims to reduce the time it takes for our test suite to run.

It seems to cut test times to 25-30% of the original times.

Notable changes:
- Remove the pre `pnpm test` bundling (since this already happens in postinstall)
- Removes the post install rerun of `npm i`
- Removes the test env build completely (this was over 50% of our test time)

### Results

[Pre optimization test run](https://github.com/evidence-dev/evidence/actions/runs/8398470151/job/23003750927?pr=1780) 

**Ubuntu 18:** 6:44 -> 1:33
**Ubuntu 20:** 6:21 -> 1:34
**Mac 18:** 12mins -> 4mins
**Mac 20:** 10mins -> 4mins
**Windows 18:** 19mins -> 6min
**Windows 20:** 19mins -> 6min

### DIFF
--- a/sites/test-env/package.json
+++ b/sites/test-env/package.json
@@ -6,7 +6,6 @@
 		"build": "evidence build",
 		"build:strict": "evidence build:strict",
 		"dev": "evidence dev",
-		"test": "evidence sources && cross-env NODE_OPTIONS=--max-old-space-size=8192 evidence build",
 		"sources": "evidence sources",
 		"preview": "evidence preview"
 	},
--- a/sites/test-env/package.json
+++ b/sites/test-env/package.json
@@ -3,8 +3,8 @@
 	"version": "3.0.22",
 	"private": true,
 	"scripts": {
-		"build": "evidence build",
-		"build:strict": "evidence build:strict",
+		"build": "cross-env NODE_OPTIONS=--max-old-space-size=8192 evidence build",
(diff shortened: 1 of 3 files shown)


csjh:
I'm not too hot on dropping the `test-env` build test, unless it's replaced with a netlify build (which would be an improvement)

archiewood:
Adding a netlify build is a great idea

archiewood:
I added a netlify build but it fails due to OOM errors...

archiewood:
boosted the memory and it worked

sites/test-env/package.json
```diff
@@ -6,7 +6,6 @@
 		"build": "evidence build",
 		"build:strict": "evidence build:strict",
```
mcrascal, line 7: update the build command to get it to go on netlify 
```suggestion
		"build:strict": "cross-env NODE_OPTIONS=--max-old-space-size=8192 vite build",

```

archiewood: I actually inlined that into netlify, and the approach worked also - this seems cleaner though

why use vite build not evidence build? Does it matter

mcrascal: sorry, I just copied and pasted from the example project, but it should be evidence build!

archiewood:
closing in favour of https://github.com/evidence-dev/evidence/pull/1801 in order to trigger netlify build of test-env

PR #1788, opened 2024-03-25
Add connected charts

hughess:
### Description
Adds `connectGroup` prop (string) which connects any charts sharing the same `connectGroup` and provides synchronized tooltip hovering.

https://github.com/evidence-dev/evidence/assets/12602440/3f875873-9c61-49f0-854b-35a1d9ea5926

CleanShot 2024-03-24 at 19 55 59

### To do before merging
- Test connection on Heatmap, CalendarHeatmap
- Look into interaction options on map - highlighting hovered state and showing value in legend would be nice

### Checklist


packages/ui/core-components/src/lib/unsorted/viz/core/ECharts.svelte
```diff
@@ -183,7 +187,6 @@
 	.chart {
 		-moz-user-select: none;
 		-webkit-user-select: none;
-		-ms-user-select: none;
```
csjh, line 186: Why was this removed?

packages/lib/component-utilities/src/echarts.js
csjh, line 1: `echarts.js` and `echartsMap.js` should probably be merged (out of scope of this PR)

csjh (approved):

packages/ui/core-components/src/lib/unsorted/viz/core/ECharts.svelte
hughess: that i'm not sure - wasn't intentional. just added it back

PR #1789, opened 2024-03-25
Add options to layout

hughess:
### Description
Adds options to layout:
- `hideHeader`
- `hideBreadcrumbs`
- `hideTOC`

Also modifies some margin/padding to make this possible, but likely needs a good look to make sure the changes are not brittle.

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
+++ b/packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
@@ -55,7 +55,7 @@
 	$: crumbs = buildCrumbs(pathArray);
 </script>
 
-<div class="flex items-start mt-8 sm:mt-12 whitespace-nowrap overflow-auto">
+<div class="flex items-start mt-[6px] whitespace-nowrap overflow-auto">
 	<div class="inline-flex items-center text-sm capitalize gap-1 text-gray-500 mb-2 sm:mb-4">
 		{#each crumbs as crumb, i}
 			{#if i > 0}
--- a/packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
+++ b/packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
@@ -55,7 +55,7 @@
 	$: crumbs = buildCrumbs(pathArray);
 </script>
 
-<div class="flex items-start mt-[6px] whitespace-nowrap overflow-auto">
+<div class="flex items-start mt-0 whitespace-nowrap overflow-auto">
 	<div class="inline-flex items-center text-sm capitalize gap-1 text-gray-500 mb-2 sm:mb-4">
 		{#each crumbs as crumb, i}
 			{#if i > 0}
--- a/packages/ui/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
+++ b/packages/ui/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
@@ -35,6 +35,13 @@
 	/** @type {string}*/
 	export let maxWidth = undefined;
 
+	/** @type {boolean} */
+	export let hideBreadcrumbs = false;
+	/** @type {boolean} */
+	export let hideHeader = false;
+	/** @type {boolean} */
+	export let hideTOC = false;
+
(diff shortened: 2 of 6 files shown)


packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
```diff
@@ -55,7 +55,7 @@
 	$: crumbs = buildCrumbs(pathArray);
 </script>
 
-<div class="flex items-start mt-8 sm:mt-12 whitespace-nowrap overflow-auto">
+<div class="flex items-start mt-[6px] whitespace-nowrap overflow-auto">
```
mcrascal, line 58: This is a regression, we currently reduce whitespace between the breadcrumbs and the header on mobile. 

```suggestion
<div class="flex items-start whitespace-nowrap overflow-auto">
```

packages/ui/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
```diff
@@ -52,38 +59,51 @@
...
+					{hideHeader}
+				/>
 			</div>
 		{/if}
 		<main
-			class={(!hideSidebar ? 'md:px-8 ' : '') +
-				'flex-grow overflow-x-hidden print:px-0 print:md:px-0 py-8'}
+			class={(!hideSidebar ? 'md:pl-8 ' : '') +
+				(!hideTOC ? 'md:pr-8 ' : '') +
+				(!hideHeader ? 'mt-[74px] ' : 'mt-[26px] ') +
```
mcrascal, line 97: Could we move the existing breadcrumb behaviour here to avoid having to hardcode pixels? 
 
```suggestion
				(!hideHeader ? ' mt-16 sm:mt-20 ' : ' mt-8 sm:mt-12 ') +

```

packages/ui/core-components/src/lib/organisms/layout/header/Header.svelte
```diff
@@ -110,6 +110,7 @@
 
 <style>
 	.active {
-		@apply bg-gray-100;
+		--tw-bg-opacity: 1;
+		background-color: rgb(243 244 246 / var(--tw-bg-opacity));
```
mcrascal, line 114: What does the active class do? afaict it's no longer used, but it's getting changed here.

mcrascal (approved):
LGTM - a couple suggestions in there.

packages/ui/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
hughess: I found that it was just between `mt-20` and `mt-18` to get the page title aligned with the top of the sidebar. This is with `mt-20`:

CleanShot 2024-03-27 at 12 01 22@2x

packages/ui/core-components/src/lib/organisms/layout/header/Header.svelte
hughess: seems like it wasn't needed - I just took it out

packages/ui/core-components/src/lib/organisms/layout/EvidenceDefaultLayout.svelte
hughess: I've moved the breadcrumb logic here - used tailwind sizes wherever possible, and adjusted for mobile. It should match up with what we've got now

hughess:
Added another adjustment for TOC when header is hidden to push it up in line with the top of the sidebar.

Think this is good to go for now.

Other things this has made me think about re: layout control:
- If someone hides the sidebar, will they expect to have the mobile menu available to navigate? Or would they want to be able to configure that? (e.g., something like `hideSidebar={true} showHamburger={true}`)
- We may need to expose control over the margins of the page. Especially for embedded use cases, if someone turns off all the nav options (header, sidebar, breadcrumbs, TOC) I think they will want fine grain control over the edges of the window
- With `hideHeader`, we lose access to page options like show/hide queries, print, settings, etc. 
   - I think there's a need for a dev toolbar for cases like this, and there's probably other stuff we can add to that toolbar to make it a really useful helper for people as they're developing in Evidence

PR #1796, opened 2024-03-26
Fix format object retrieval in Chart and Value components

hughess:
### Description
In Value and BigValue, when `fmt` was applied to nulls, it threw an error. This PR removes the error and passes the value through as "-"

### Notes
I think we can simplify the flow for value formatting. When we set it up, the only way to format values was with format tags in column names, but now the primary method of formatting is using `fmt` props. Creating a format object gets a bit confusing when `fmt` is the method, as we need the value type for the column, which comes from `columnSummary`.

Once we fix the date/timezone issue and rip out the date parsing logic (which also requires `columnSummary`, it would be a good time to reconsider the default formatting approach inside components.

### Checklist

### DIFF
--- a/.changeset/cyan-suits-fry.md
+++ b/.changeset/cyan-suits-fry.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Fix errors caused by formatting null values
--- a/packages/ui/core-components/src/lib/unsorted/viz/core/_Chart.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/core/_Chart.svelte
@@ -452,7 +452,7 @@
 			// Get format codes for axes
 			// ---------------------------------------------------------------------------------------
 			if (xFmt) {
-				xFormat = getFormatObjectFromString(xFmt, columnSummary[x].format.valueType);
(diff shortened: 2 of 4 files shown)


csjh:
> Once we fix the date/timezone issue and rip out the date parsing logic (which also requires columnSummary)

Could you elaborate on that? I don't think the date parsing logic requires `columnSummary`

archiewood (approved):

hughess:
@csjh it's this section in `_Value` I was referring to (not the actual date parsing, but the date overriding):
```
columnSummary = getColumnSummary(data, 'array');

const dateCols = columnSummary
   .filter((d) => d.type === 'date' && !(data[0]?.[d.id] instanceof Date))
   .map((d) => d.id);

for (let i = 0; i < dateCols.length; i++) {
   data = convertColumnToDate(data, dateCols[i]);
}
```

csjh:
I think that part can be dropped altogether in a breaking change(?)

`!(data[0][id] instanceof Date)` isn't true if the data is coming from a duckdb (because all date/timestamp columns become `Date` objects

The only time it would be used is with hardcoded data (or *possibly* string columns? but I think we removed string column inference)

PR #1801, opened 2024-03-27
Test speedup v2

archiewood:
### Description

This PR aims to reduce the time it takes for our test suite to run.

It seems to cut test times to 25-30% of the original times.

Notable changes:
- Remove the pre `pnpm test` bundling (since this already happens in postinstall)
- Removes the post install rerun of `npm i`
- Removes the test-env build completely (this was over 50% of our test time)
- Separately have added added netlify deploy for test-env to replace coverage

### Results

[Pre optimization test run](https://github.com/evidence-dev/evidence/actions/runs/8398470151/job/23003750927?pr=1780) 

**Ubuntu 18:** 6:44 -> 1:33
**Ubuntu 20:** 6:21 -> 1:34
**Mac 18:** 12mins -> 4mins
**Mac 20:** 10mins -> 4mins
**Windows 18:** 19mins -> 6min
**Windows 20:** 19mins -> 6min


ItsMeBrianD:
I think that running the `test-env` build is an important indicator for PRs, we could try to put it into it's own check if we want to parallelize it better, but getting rid of it doesn't seem like the right move

archiewood:
@ItsMeBrianD the test env build is now happening on netlify - does that allay concerns at all?

It's the single most time consuming step in our tests - it adds around 5 minutes to each test - vs about 10 seconds - 1min for the rest of the test suite

ItsMeBrianD:
> does that allay concerns at all?

My biggest concern is that we're going to lose windows testing, which is only a problem because windows uses `\` instead of `/` for paths, so that style bug could slip through undetected.

We could move the build of test-env to a `next`->`main` check to help mitigate the issue.

Ultimately I'm open to going whichever direction here, if we don't think this is worth the headache, then lets carry on like this until given a reason to look at it again

archiewood:
> My biggest concern is that we're going to lose windows testing, which is only a problem because windows uses \ instead of / for paths, so that style bug could slip through undetected.

Gotcha

> We could move the build of test-env to a next->main check to help mitigate the issue.

This seems smart

archiewood:
works

Issue #1352, opened 2023-11-22
Add ID field to generated headers


ItsMeBrianD:
Right now; it is not possible to link to headers on an Evidence page because they do not have an `id` attribute

We can create (or use) a rehype plugin to add these headers

https://github.com/evidence-dev/evidence/blob/next/packages/preprocess/src/add-classes.cjs

https://github.com/rehypejs/rehype-slug

archiewood:
I'm not sure if we are talking about the same thing, but you can use anchor tags to navigate to specific sections?

![CleanShot 2023-11-22 at 10 29 44](https://github.com/evidence-dev/evidence/assets/58074498/4c092ba5-3303-4329-9fa5-df13e2c49c6f)

ItsMeBrianD:
Hmm, this might be related to the digit at the end of the links that are being used;

 `Marketing3` isn't an intuitive ref for the marketing header

archiewood:
Yeah I think we should probably have no number by default and then increment from 2 if there are multiple headers with the same name

mcrascal:
Will also need a corresponding update to the [TOC](https://github.com/evidence-dev/evidence/blob/b76fd5c02cd40580e7b0123a88ce5765810d1802/packages/core-components/src/lib/unsorted/ContentsList.svelte#L19)

PR #1806, opened 2024-03-28
Slugified Anchor Links

archiewood:
### Description

Addresses #1352

- Adds automatic anchor links to headings (H1-H6) in the preprocess stage.
- Puts H1-H3 in the TOC (rather than H1-H2)
- Links are based on slugified forms of the headers, but are made unique in the case of duplication by appending numbers
- Allows external links to navigate to the correct section of the page eg: `my-project.evidence.app/page-1/#heading-on-the-page`

### Issues
- [ ] Issue (svelte kit bug? https://github.com/sveltejs/kit/issues/11326) which prevents navigation between pages in Evidence from working correctly. **Open issue for this**
  - Appears to be client side nav only issue
  - This is not a regression as this did not work before.
- This uses outdated versions of rehype packages that still support CJS. I suspect the long term solution here is to convert this whole module to ESM, but then we need to rip out Prism (code highlighting) which is CJS only.

CleanShot 2024-03-27 at 22 15 54

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/organisms/layout/tableofcontents/ContentsList.svelte
+++ b/packages/ui/core-components/src/lib/organisms/layout/tableofcontents/ContentsList.svelte
@@ -5,11 +5,7 @@
 	let observer;
 
 	function updateLinks() {
-		headers = Array.from(document.querySelectorAll('h1.markdown, h2.markdown'));
-		headers.forEach((header, i) => {
-			// Headers may contain values that change in response to user input, so we create our anchors as just the position on the page.
-			header.id = encodeURIComponent(i + 1);
-		});
+		headers = Array.from(document.querySelectorAll('h1.markdown, h2.markdown, h3.markdown'));
 	}
 
 	function observeDocumentChanges() {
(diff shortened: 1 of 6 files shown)


packages/ui/core-components/src/lib/organisms/layout/tableofcontents/ContentsList.svelte
```diff
@@ -38,12 +34,8 @@
...
 		On this page
 	</span>
-	{#each headers as header, i}
-		<a
-			href={'#' + encodeURIComponent(i + 1)}
-			class={header.nodeName.toLowerCase()}
-			class:first={i === 0}
-		>
+	{#each headers as header}
+		<a href={'#' + header.id} class={header.nodeName.toLowerCase()}>
```
csjh, line 38:
```suggestion
		<a href="#{header.id}" class={header.nodeName.toLowerCase()}>
```

csjh (approved):

sites/example-project/src/pages/links-to-page-headers/+page.md
```diff
@@ -0,0 +1,4 @@
+
+Doesn't currently work
+
+[Deep in the text](/text-and-metrics/text#more-nostra-non-parsque-talia-discussit)
```
mcrascal, line 4: Could we open an issue instead of doing this?

mcrascal (approved):

sites/example-project/src/pages/links-to-page-headers/+page.md
archiewood: good idea, will do upon merge

archiewood: #1871

PR #1807, opened 2024-03-28
Fix grid printing & add landscape print support

hughess:
(Updated)

### Description
- Adds support for landscape printing
- Fixes printing for Grid components, which were previously incorrectly sized and overlapped each other

### Example
Same page shown printed in portrait and landscape

CleanShot 2024-04-03 at 20 21 22@2x
CleanShot 2024-04-03 at 20 21 13@2x

### Introduced in this PR
- Sets a fixed print width for portrait pages at 650px and landscape pages at 841px (assumes 8.5x11 page).
- Sets maximum grid columns at 2 for portrait and 3 for landscape
- Adds context to the Grid component to share with child components - charts are now aware they are inside a Grid, the number of columns in the grid, and the gap width

### Known Issues
- Safari printing is affected by the width of the browser window when print is selected. This does not happen in other browsers
   - Recommendation is to print using another major browser, or adjust your browser window width until the print output meets your needs

### Implementation Notes
- When a chart is in a Grid, the print action creates 2 images for each chart and swaps it out based on the size of the page
   - This relies on hardcoded numbers for gap widths and page widths 

### Checklist

### DIFF
--- a/.changeset/chilled-students-deliver.md
+++ b/.changeset/chilled-students-deliver.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/core-components': patch
+'@evidence-dev/components': patch
+---
+
+Fix printing of charts in grid component
--- a/packages/lib/component-utilities/src/echartsCopy.js
+++ b/packages/lib/component-utilities/src/echartsCopy.js
@@ -4,9 +4,20 @@ import { evidenceThemeLight } from './echartsThemes';
 export default (node, option) => {
 	registerTheme('evidence-light', evidenceThemeLight);
 
-	const { config, ratio, echartsOptions, seriesOptions, seriesColors } = option;
+	const { config, ratio, echartsOptions, seriesOptions, seriesColors, isMap, extraHeight, width } =
+		option;
 
-	const chart = init(node, 'evidence-light', { renderer: 'canvas' });
+	let initOpts = { renderer: 'canvas' };
+	if (isMap) {
+		initOpts.height = width * 0.5 + extraHeight;
+		if (node && node.parentNode) {
+			// node.parentNode refers to the chart's container
+			node.style.height = initOpts.height + 'px';
+			node.parentNode.style.height = initOpts.height + 'px';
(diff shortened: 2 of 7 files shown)


hughess:
Update: have tested on the following browsers and all looks good

- MacOS
   - Chrome
   - Safari
   - Firefox 
- Windows
   - Chrome
   - Firefox
   - Edge

hughess:
Print width of 100% doesn’t work on iOS Safari, I assume because of the same echartsCopy dimension issue

mcrascal:
This causes the width of the browser when the print event is fired to affect what is actually printed. 


Printing from a medium width: 

CleanShot 2024-03-28 at 09 39 21@2x


Printing from a narrow width: 
CleanShot 2024-03-28 at 09 39 28@2x

hughess:
It looks like there's a timing issue where echarts doesn't get the dimensions of the updated chart container for the printed page size. I'm looking into if there's a way to access that info during the print setup

hughess:
I think I have a solution here, at least until we can figure out something better:

- Generate 2 images of each chart on print - 1 for portrait and 1 for landscape
- Assume portrait maps to anything `sm:` and below per tailwind, and landscape anything above that
- Set `hidden` vs. `block` on the divs containing these images based on the screen width

For charts within a Grid:
- Set up a store within each Grid to track the number of columns and the gap width
- Get the grid context in EChartsCopyTarget and use that to set the correct width for the container based on the screen width settings describe above

hughess:
Just added the grid context approach.

Some things to figure out before it's ready:
- Confirm that page widths are correct for portrait and landscape
- Confirm ideal print behaviour for landscape - at the moment it is constrained by the article
- Tune the grid gap width calcs - something seems off, 2-col grid and 3-col grid are not right-aligned on print at the moment

And for testing:
- Test across browsers
- Test across layout page widths (e.g., with sidebar hidden, etc)

mcrascal (approved):

Issue #1735, opened 2024-03-13
Source SQL ending with comment breaks evidence sources


smilingthax:
## Steps To Reproduce

1. Simple Project w/ Postgres source
2. Add `test.sql` to `sources/mypg/`:
```
SELECT 1
-- comment
``` 
3. run `evidence sources --debug`

## Expected Behavior

The command does not fail.

## Actual Behaviour

```
  test ✖ Cannot read properties of undefined (reading 'rows')
Error: Cannot read properties of undefined (reading 'rows')
    at runQuery (/evidence/node_modules/@evidence-dev/postgres/index.cjs:197:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async $18acbd5e40ec20a6$export$53c7a8f067b9e2ba (file:///evidence/node_modules/@evidence-dev/plugin-connector/dist/index.mjs:1350:38)
    at async $acd86e83767fde42$export$76919addde332833 (file:///evidence/node_modules/@evidence-dev/plugin-connector/dist/index.mjs:1581:22)
    at async file:///evidence/node_modules/@evidence-dev/evidence/cli.js:349:3
```

Also, the node command then "hangs", i.e. never returns to the shell.

## Workarounds

Add `;` on a separate line at the end:
```
SELECT 1
-- comment
;
```

## Additional Information

The failing query is here: https://github.com/evidence-dev/evidence/blob/next/packages/postgres/index.cjs#L160  

The culprit is https://github.com/evidence-dev/evidence/blob/next/packages/db-commons/index.cjs#L186 cleanQuery(), calling `.trim()` on the query, which is wrong, because it no longer ends the sql-comment-line.

The hanging command is probably caused by the exception handlers not closing the pg connection(?).

archiewood:
This bug also exists in SQL queries defined in markdown

smilingthax:
A possibly "good enough" solution for the `cleanQuery`-part could be in https://github.com/evidence-dev/evidence/blob/next/packages/db-commons/index.cjs#L189
```
const cleanQuery = (query) => {
	let cleanedString = query.trim();
	if (cleanedString.endsWith(';'))
		cleanedString = cleanedString.substring(0, cleanedString.length - 1);
	return cleanedString;
};
```
to return this instead:
```
	return cleanedString + '\n';
```

archiewood:
Hey @smilingthax

Thanks for point this out. We'd accept a PR with this if you are open to it.

[...]

PR #1819, opened 2024-03-29
source sql ending in a line comment need final newline

smilingthax:
### Description

See https://github.com/evidence-dev/evidence/issues/1735

### Checklist


archiewood:
I believe these tests are failing because forks do not have access to the secrets we need to run e.g. postgres tests

I think this is due to the recent change in test suite that moved from `pull_request_target` to `pull_request`

csjh (approved):
Thanks!

Issue #1750, opened 2024-03-16
Postgres uppercase column broken


smilingthax:
This commit removed `standardizeResult` in postgres/index.cjs `runQuery`:
https://github.com/evidence-dev/evidence/commit/5e77f6c1d10907dd8fa942c69469a5feefb9d5da#diff-9249d8428d1cac33bfdd2f7d97433f4c6df762afd6d1fd7b9dfbc454ab11f658

This commit added it back, but only partially – i.e. not for the first batch / 100000 rows:
https://github.com/evidence-dev/evidence/commit/99b5972ac6bc7e8b2bb817c64cbf07b9ca8c7467#diff-9249d8428d1cac33bfdd2f7d97433f4c6df762afd6d1fd7b9dfbc454ab11f658R215

(There was also https://github.com/evidence-dev/evidence/issues/478 which questioned the need for lowercasing. AFAICT, duckdb is case-insensitive-but-preserving, which seems to cause problems w/ evidence trying to format the result e.g. of `SELECT *` later...)

PR #1822, opened 2024-03-29
standardizeResult should also be applied to first batch(?)

smilingthax:
### Description

Closes https://github.com/evidence-dev/evidence/issues/1750.  
This PR just re-adds `standardizeResult` to the first batch. I have not extensively tested this in all combinations; AFAIU from the commits mentioned in #1750 this seems to be the right thing to do, though...

### Checklist


csjh (approved):
Thanks v4 😁

PR #1825, opened 2024-03-29
Adjust delta neutral symbol

hughess:
### Description
Changes delta neutral symbol from en dash to em dash when `symbolPosition=left`

This helps with BigValue values where positive values could be confused for negative values.

A better long-term solution is needed for this, but putting this in as a short term fix.

#### Before
CleanShot 2024-03-29 at 14 12 56@2x

#### After
CleanShot 2024-03-29 at 14 16 41@2x

### Checklist

### DIFF
--- a/.changeset/new-cows-speak.md
+++ b/.changeset/new-cows-speak.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Adjust delta neutral symbol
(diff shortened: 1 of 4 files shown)


archiewood:
Also could: make it bold **-**

Use a thicker dash: U+2796

archiewood (approved):

hughess:
Added fix for null and 0 values which were erroring out in the Delta component.

0 values will appear as 0 (or however the user chooses to format them) and nulls will appear as `-` to maintain consistency with null treatment elsewhere.

There's likely a better way to handle nulls (N/A, null symbol, dashes of various types), and it might need to be configurable depending on the situation. Leaving it for now, so we can tackle it in the same way across the product.

PR #1851, opened 2024-04-05
Release 04 05 24

mcrascal:


csjh:
Looks like I accidentally axed `export const prerender = true` in the loading pages PR, not sure why netlify caught in this one but not the other

mcrascal:
Good eye -- I wonder if it was falling back to an earlier build in the other PR?

ItsMeBrianD (approved):

PR #1866, opened 2024-04-10
Feat/dark mode

mcrascal:
- adds a small token system based on shadcn, which I think will be a suitable starting point for theming 
- adds a dark mode toggle to the layout


ItsMeBrianD:
@mcrascal what's the play here? I would advocate for closing this and handing the issue off to @zachstence

mcrascal:
Fine....

PR #1887, opened 2024-04-13
fix: remove comments and trailing `;` in sql query

### LINKED ISSUE #733: Remove `;` from the end of query dependencies, if it is there
**Bug Description**
When chaining queries, if a query ends with a `;`, it is not removed when it is chained. This can lead to syntax errors

**Severity**

**Expected Behavior**
A "correct" query that ends with a `;` should not break when it is used in chaining.

If you received any error messages, please post them here. Screenshots are also helpful.
image

**Steps To Reproduce**
Let us know the steps we can take to reproduce the issue you're seeing. Sample data helps speed up the process of reproducing.
On any page; use these queries:

\`\`\`a
SELECT 1 as t;
\`\`\`

\`\`\`b
SELECT t * 5 from ${a};
\`\`\`

**Workarounds**
Remove `;`

**Do you want to contribute a fix for this bug?**
Probably

**Database:**
- [ ] Postgres
- [X] BigQuery
- [ ] Snowflake
- [ ] Redshift
- [ ] MySQL
- [ ] SQLite
- [ ] DuckDB
- [ ] CSV
- [ ] Other (please specify)

### LINKED ISSUE #1450: Queries cannot end with `;`
## Steps To Reproduce

In markdown

````
```sql test
SELECT 1;
```
````

## Expected Behavior

Query Runs

## Actual Behaviour

```
Parser Error: syntax error at or near ";" LINE XX: SELECT 1;) AS "__userQuery"
```

## Workarounds

Don't terminate your Queries with `;`

### LINKED ISSUE #1735: Source SQL ending with comment breaks evidence sources

AyushAgrawal-A2:
### Description

closes #733
closes #1735
closes #1450

Changes:
- Remove single / multiline comments in query.
- Remove trailing `;` in query.
- Completes TODO of `QueryStore` and fixes bug in chained query due to trailing `;`.

Before:
image

After:
image

### Checklist


csjh:
Thanks for the PR!

One concern: seems like queries with string-enclosed "comments" are also caught here, which could be an issue if it's used as a delimiter or such in a string input (i.e. `New York -- NY`)

AyushAgrawal-A2:
@csjh 
Sorry, I missed that edge case. Tackling this requires parsing the query, I don't think it can be done via simple regex replace. I have added a helper function to do so.

This is required in 3 packages. In order to avoid cross package dependencies I have added this in all three packages. But I don't believe duplicate code is good. Please suggest if there is a common place to place this function. I tried placing it in db-commons but it create a cyclic dependency betweet QueryStore and db-commons.

Before:
Screenshot from 2024-04-18 18-47-39
After:
Screenshot from 2024-04-18 18-43-39

ItsMeBrianD:
> This is required in 3 packages. In order to avoid cross package dependencies I have added this in all three packages. But I don't believe duplicate code is good. Please suggest if there is a common place to place this function. I tried placing it in db-commons but it create a cyclic dependency betweet QueryStore and db-commons.

This should only be used in the `query-store` (which is now `sdk/usql/query.js`), and that should cover the impacted area

AyushAgrawal-A2:
> > This is required in 3 packages. In order to avoid cross package dependencies I have added this in all three packages. But I don't believe duplicate code is good. Please suggest if there is a common place to place this function. I tried placing it in db-commons but it create a cyclic dependency betweet QueryStore and db-commons.
> 
> This should only be used in the `query-store` (which is now `sdk/usql/query.js`), and that should cover the impacted area

@ItsMeBrianD 
This is necessary for `preprocess/src/extract-queries/extract-queries.cjs` also so that the dependent queries are cleaned properly.

AyushAgrawal-A2:
> > This is required in 3 packages. In order to avoid cross package dependencies I have added this in all three packages. But I don't believe duplicate code is good. Please suggest if there is a common place to place this function. I tried placing it in db-commons but it create a cyclic dependency betweet QueryStore and db-commons.
> 
> This should only be used in the `query-store` (which is now `sdk/usql/query.js`), and that should cover the impacted area

@ItsMeBrianD 
Done
image

ItsMeBrianD (approved):

ItsMeBrianD:
This is ready to be merged, but going to hold off until after the release so there is a testing window on `next`, thanks for the contribution!

ItsMeBrianD:
This is addressed in #2103

Issue #1876, opened 2024-04-12
`npm run dev` logs Sourcemap for "../node_modules/@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js" points to missing source files


archiewood:
## Steps To Reproduce

Start the Evidence dev server

## Expected Behavior

Terminal prints

```bash
> my-evidence-project@0.0.1 dev
> evidence dev --open /

Starting Evidence development workspace - this can take up to 90 seconds

Cleared query cache

  VITE v5.1.6  ready in 933 ms

  ➜  Local:   http://localhost:3000/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
```

## Actual Behaviour

```bash
> my-evidence-project@0.0.1 dev
> evidence dev --open /

Starting Evidence development workspace - this can take up to 90 seconds

Cleared query cache

  VITE v5.1.6  ready in 933 ms

  ➜  Local:   http://localhost:3002/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
Sourcemap for "/Users/archie/jaffle_shop_duckdb/reports/node_modules/@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js" points to missing source files
```

## Workarounds

Ignore this message - it does not appear to cause issues

archiewood:
Possibly related: https://github.com/vitejs/vite/issues/5438

PR #1888, opened 2024-04-14
Fix: ignore a specific warning during dev server run

phukon:
ref #1876
this should suppress that annoying message


archiewood:
Feel like this is a sensible approach, thanks

archiewood (approved):

PR #1895, opened 2024-04-15
Remove unused deps

archiewood:
### Description

Removes various old versions of duckdb that do not appear to be used, and echarts from the root of the monorepo

### Checklist


archiewood:
@csjh would be great to confirm that this duckdb dep is indeed not required?

csjh (approved):
should be good, I think the ddb dep in usql snuck in early on and overstayed its welcome

PR #1902, opened 2024-04-16
Add conditional formatting based on another column and red negatives

hughess:
### Description
3 improvements to DataTable:
1. Use another column to define a color scale `scaleColumn`
2. Use another column to define value formatting `fmtColumn`
3. Set negative values to show as red `redNegatives`

### Custom Column Conditional Formatting
CleanShot 2024-04-16 at 15 55 55@2x

### Red Negative Values
Use `redNegatives=true` in the `Column` component to set negative values to show as red.

CleanShot 2024-04-16 at 19 30 03@2x

### Checklist


packages/ui/core-components/src/lib/unsorted/viz/table/TableRow.svelte
```diff
@@ -54,19 +54,48 @@
...
+							: uiColours.grey999
+						: ''}
+				{@const bottom_border =
+					i !== displayedData.length - 1 &&
+					rowLines &&
+					column.contentType === 'colorscale' &&
+					is_nonzero &&
+					column.colorPalette
+						? `1px solid ${chroma(cell_color).darken(0.5)}`
+						: ''}
```
ItsMeBrianD, line 98: Not addressable in this PR, but this really is starting to smell like it needs to be it's own component, rather than all of these `{@const}`s

ItsMeBrianD (approved):

PR #1925, opened 2024-04-21
Add syntax highlighting to code blocks

hughess:
### Before
CleanShot 2024-04-21 at 14 17 15@2x

### After
CleanShot 2024-04-21 at 14 16 59@2x

CleanShot 2024-04-21 at 14 10 17@2x

## Potential Issues
- Styling issues - `@apply` wasn't working, so for now I've had to override styles in the html tag
- Uses a slightly modified definition of Svelte syntax from `svelte-prism` rather than importing from that library
- Prism themes not working, stuck on default styling
- Import method for languages may not be ideal

### Checklist

### To do before merge
- [x] Remove `document` usage in reactive statement
- [ ] Confirm import method for languages and Prism
- [x] Clean up styling conflicts
- [ ] Set appropriate margins/padding on code box
- [ ] Add a language option for `evidence` so users can include documentation in their project without having to know about `svelte`

### DIFF
--- a/.changeset/cyan-carrots-worry.md
+++ b/.changeset/cyan-carrots-worry.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/preprocess': patch
+'@evidence-dev/core-components': patch
+---
+
+Add syntax highlighting to code blocks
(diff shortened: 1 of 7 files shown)


ud3sh (approved):

mcrascal:
Not sure why this doesn't build, but I took a look at the `@apply` issues. Apply works in the component, but it seems like it was conflicting with the mix of inlined styles and classes. I've opened a PR (#1955) that just converts it to normal tailwind.

hughess:
Ah it turns out there was a `document` in there breaking the build. Hopefully got a fix for that

PR #1935, opened 2024-04-23
Evidence SPA Mode

csjh:
### Description

Adds SvelteKit SPA compatibility to Evidence

### Checklist


sites/example-project/src/pages/+layout.js
```diff
@@ -64,30 +68,50 @@ async function getPrerenderedQueries({ data: { routeHash, paramsHash }, fetch })
 	return Object.fromEntries(resolved_entries.filter(Boolean));
 }
 
+/** @satisfies {import("./$types").LayoutServerLoad} */
```
ItsMeBrianD, line 71: Add comments describing what this is doing

ItsMeBrianD (changes requested):
Needs comments explaining the magic and then it should be g2g

csjh:
@ItsMeBrianD I updated it to remove `+layout.server.js` like I mentioned in our call earlier - originally wasn't going to, but it makes the hosting process much smoother, and I don't think there's really any consequences, unless you can think of anything

ItsMeBrianD (approved):

PR #1939, opened 2024-04-24
Fix scrollbar appearing when hovering last chart on page

hughess:
### Description
When hovering the last chart on a page, occasionally a scroll would appear on the page.

This adds padding to the article to eliminate the scrollbar.

#### Before
CleanShot 2024-04-24 at 16 01 20

#### After
CleanShot 2024-04-24 at 16 02 11

### Checklist

### DIFF
--- a/.changeset/sour-papayas-act.md
+++ b/.changeset/sour-papayas-act.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Fix scrollbar appearing on last chart on page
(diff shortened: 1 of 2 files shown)


hughess:
@mcrascal this seems to work well, but a couple things I wasn't sure about on this one:
- Is `article` or `main` the right place for the extra padding/margin?
- Should it be padding, margin, or something else?

ud3sh (approved):

mcrascal:
@hughess this seems totally fine, but I think the extra space should be added to the charts themselves. As an example, `datatable` doesn't cause this bug because it includes sufficient space to accommodate the download buttons. 

With this fix, I think this will crop up again when composing charts into wrapper components (like a modal), or when people are building custom components.

PR #1951, opened 2024-05-01
Add `InputMap` component

csjh:
### Description

Adapts `@evidence-dev/labs` `LeafletMap` component into an `InputMap` component

image

### Checklist


Youssef-Harby:
I will be happy if you support also the other geometry types like lines, polygons, multi-geometry.

packages/ui/core-components/src/lib/atoms/inputs/maps/_InputMap.svelte
```diff
@@ -0,0 +1,156 @@
...
+
+		return {
+			destroy() {
+				actionPromise.then((action) => action.destroy());
+			},
+			update(newData) {
+				actionPromise.then((action) => action.update(newData));
+			}
+		};
+	}
```
ItsMeBrianD, line 144: Stylistic choice but I think this should go in it's own file

csjh: I'm not really a fan of it going in a separate file, all the captured variables have to be forwarded in, which gets ugly

hughess:
This is awesome!

I think we should hold this one until we've got a final syntax we want to use for maps, and release it all at once. There's features for choropleths and tooltips I want to pull in from a custom prototype I've got, so I'll bring those changes into here

nemccarthy:
Sad to see this closed. Any idea when the maps stuff will be released?

hughess:
@nemccarthy the maps are out already! Here is one of the map components: https://docs.evidence.dev/components/point-map/

It can be used as an input by setting the `name` prop.

PR #1954, opened 2024-05-01
Additional date range presets

mcrascal:
### Description

Adds several new presets to the defaults included in the date range and a separator between groups.

### DIFF
--- a/.changeset/giant-trainers-invent.md
+++ b/.changeset/giant-trainers-invent.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


mcrascal:
@ItsMeBrianD what's going on w/ the netlify action?

csjh:
I think we might have to specify [archie's forked action](https://github.com/archiewood/wait-for-netlify-action/compare/v1.0...v1.0.2) as `@v1.0.2` instead of `@v1`? Otherwise the output doesn't make any sense to me

csjh (approved):

PR #1988, opened 2024-05-08
Date range default value

### LINKED ISSUE #1984: Default date range for DateRange input
Ability to set a default date range which will take effect on page load.

e.g., defaultValue=“Last 7 days”

Franreno:
### Description

This PR aims to add a defaultValue parameter to the DateRange input component.
The defaultValue has the purpose to define a default range when the component is first loaded.

To use is to simply add the defaultValue as below:
```js
<DateRange 
    data={orders_by_month} 
    dates=month 
    name=range_default 
    defaultValue="Last 90 days"
/>
```
Then, the DateRange will initialize with the defaultValue provided

Screenshot 2024-05-08 at 19 14 02

If the parameter is not provided or does not comply with the labels available on the `preset` variable, then the component's behavior should be the same as previously

Screenshot 2024-05-08 at 19 13 52

Closes #1984

### Checklist

### DIFF
--- a/.changeset/early-crews-compete.md
+++ b/.changeset/early-crews-compete.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/core-components': patch
+'@evidence-dev/components': patch
+---
+
+Add defaulValue to DateRange
--- a/.changeset/early-crews-compete.md
+++ b/.changeset/early-crews-compete.md
@@ -3,4 +3,4 @@
 '@evidence-dev/components': patch
 ---
 
-Add defaulValue to DateRange
+Add defaultValue to DateRange
--- a/packages/ui/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
@@ -35,6 +35,9 @@
 	/** @type {string | undefined} */
 	export let dates;
 
+	/** @type {string | undefined} */
+	export let defaultValue;
+
 	const exec = getQueryFunction();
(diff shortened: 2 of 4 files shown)


.changeset/early-crews-compete.md
```diff
@@ -0,0 +1,6 @@
+---
+'@evidence-dev/core-components': patch
+'@evidence-dev/components': patch
+---
+
+Add defaulValue to DateRange
```
ItsMeBrianD, line 6:
```suggestion
Add defaultValue to DateRange
```

Franreno: Ops, thanks :D

mcrascal (changes requested):
This is awesome, I like this string based API for selecting default date ranges. 

The only thing that seems off to me here is the loading behaviour. It seems that in the process of loading, results from queries that depend on the value here transition through an empty state, causing a no data error to be briefly visible in dependent components. @ItsMeBrianD can probably help identify the issue. 

CleanShot 2024-05-17 at 10 56 17

ItsMeBrianD:
> The only thing that seems off to me here is the loading behaviour. It seems that in the process of loading, results from queries that depend on the value here transition through an empty state, causing a no data error to be briefly visible in dependent components. @ItsMeBrianD can probably help identify the issue.

I think this was fixed in `features/b`, merging next in should resolve the issue

archiewood:
@kwongz was this addressed by your PR? Can we close this?

kwongz:
@archiewood  Yup! Default Values are now available as of the latest updates

PR #1993, opened 2024-05-09
Added test for testing false values

ud3sh:
### Description

It seems that `EVIDENCE_SOURCE__source__ssl__rejectUnauthorized="false"` is picked up incorrectly (always treated as true).  This failing test might be an indication as to why.  Seems to have to do with interpreting strings as boolean in loading source options.

### Checklist

### DIFF
--- a/packages/lib/plugin-connector/src/data-sources/get-sources.spec.js
+++ b/packages/lib/plugin-connector/src/data-sources/get-sources.spec.js
@@ -87,4 +87,10 @@ describe('loadSourceOptions', () => {
 		expect('ssl' in result).toBeTruthy();
 		expect('rejectUnauthorized' in result.ssl).toBeTruthy();
 	});
+	it('should pickup boolean flags', () => {
+		vi.stubEnv('EVIDENCE_SOURCE__source__ssl__rejectUnauthorized', 'false');
+		const result = loadSourceOptions('source');
+		expect('ssl' in result).toBeTruthy();
+		expect('rejectUnauthorized' in result.ssl).toBeFalsy(); // <-- this is failing
+	});
 });


packages/lib/plugin-connector/src/data-sources/get-sources.spec.js
```diff
@@ -87,4 +87,10 @@ describe('loadSourceOptions', () => {
 		expect('ssl' in result).toBeTruthy();
 		expect('rejectUnauthorized' in result.ssl).toBeTruthy();
 	});
+	it('should pickup boolean flags', () => {
+		vi.stubEnv('EVIDENCE_SOURCE__source__ssl__rejectUnauthorized', 'false');
+		const result = loadSourceOptions('source');
+		expect('ssl' in result).toBeTruthy();
+		expect('rejectUnauthorized' in result.ssl).toBeFalsy(); // <-- this is failing
+	});
 });
```
ud3sh, line 96: `{ ssl: { rejectUnauthorized: 'false' } }`

AyushAgrawal-A2: Apart from the value being a string i.e. 'false', there is another issue with this syntax :
- The `key in object` i.e. `'rejectUnauthorized' in result.ssl` syntax checks for the presence of a key in object. Its value is not being read / tested here.

All tests using this syntax are just checking the presence/absence of that key in object.

packages/lib/plugin-connector/src/data-sources/get-sources.spec.js
```diff
@@ -87,4 +87,10 @@ describe('loadSourceOptions', () => {
 		expect('ssl' in result).toBeTruthy();
 		expect('rejectUnauthorized' in result.ssl).toBeTruthy();
 	});
+	it('should pickup boolean flags', () => {
+		vi.stubEnv('EVIDENCE_SOURCE__source__ssl__rejectUnauthorized', 'false');
+		const result = loadSourceOptions('source');
+		expect('ssl' in result).toBeTruthy();
+		expect('rejectUnauthorized' in result.ssl).toBeFalsy(); // <-- this is failing
```
ItsMeBrianD, line 94:
```suggestion
		expect('rejectUnauthorized' in result.ssl).toBeTruthy();
		expect(result.ssl.rejectUnauthorized).toBeFalsey();
```

AyushAgrawal-A2: @ItsMeBrianD 
`expect(result.ssl.rejectUnauthorized).toBeFalsey();` will also fail as `rejectUnauthorized` is being parsed as string here. `"false"` is a truthy value, as @ud3sh mentioned.

ItsMeBrianD: Correct, but this test _should_ be passing

AyushAgrawal-A2: I meant to say, in the suggested changes
line 94 - `expect('rejectUnauthorized' in result.ssl).toBeTruthy();` - this will pass
line 95 - `expect(result.ssl.rejectUnauthorized).toBeFalsey();` - this will not pass

PR #2007, opened 2024-05-14
fix: drop down gap

kwongz:
### Checklist

- [ x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)


packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -269,6 +269,16 @@
 			}
 		);
 	}
+
+	//kwongz created static variable for displayed options, handle height based on number of options
```
ItsMeBrianD, line 273: Remove comment

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -346,7 +356,13 @@
 						<Command.List>
 							<Command.Empty>No results found.</Command.Empty>
 							<Command.Group>
-								<VirtualList height="160px" items={$options} let:item={option}>
+								<!-- kwongz height handle by function, passing Displayed_Options to component for handling overflow (scroll) -->
```
ItsMeBrianD, line 359: Remove Comment

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Virtual.svelte
```diff
@@ -5,11 +5,15 @@
 	export let items;
 	export let height = '100%';
 	export let itemHeight = undefined;
+	// kw added displayedOption value from Dropdown.svelte
```
ItsMeBrianD, line 8: Remove comment

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Virtual.svelte
```diff
@@ -5,11 +5,15 @@
...
+	export let displayedOptions;
 
 	// read-only, but visible to consumers via bind:start
 	export let start = 0;
 	export let end = 0;
 
+	console.log(items);
+
```
ItsMeBrianD, line 16: Remove console.log

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Virtual.svelte
```diff
@@ -135,11 +139,12 @@
 	});
 </script>
 
+<!-- kwongz added tenerary operation for scroll depending on items and displayed options value -->
```
ItsMeBrianD, line 142: Remove comment

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Virtual.svelte
```diff
@@ -135,11 +139,12 @@
...
 <div
 	bind:this={viewport}
 	bind:offsetHeight={viewport_height}
 	on:scroll={handle_scroll}
-	style="height: {height};"
+	style="height: {height}; overflow: {items < displayedOptions ? 'hidden' : 'auto'};"
```
ItsMeBrianD, line 147: We should double check that we need this here - modifying the height (currently 32px) per option to being slightly higher is probably the fix we want here

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -346,7 +356,13 @@
...
+								<VirtualList
+									height={handleDropDownHeight($options, DISPLAYED_OPTIONS)}
+									displayedOptions={DISPLAYED_OPTIONS}
```
ItsMeBrianD, line 362: I don't know if we want VirtualList to be aware of this - it shouldn't be responsible for knowing about this

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -269,6 +269,16 @@
...
+
+	//kwongz created static variable for displayed options, handle height based on number of options
+	const DISPLAYED_OPTIONS = 5;
+	function handleDropDownHeight(options, numberOfDisplayedOptions) {
+		if (options.length < numberOfDisplayedOptions) {
+			return `${options.length * 32}px`;
+		} else {
+			return `${numberOfDisplayedOptions * 32}px`;
+		}
+	}
```
ItsMeBrianD, line 281: :fire:

ItsMeBrianD (changes requested):

ItsMeBrianD (approved):

mcrascal:
@kwongz hitting the ground running, nice! 

A couple things I noticed: 

1. Dropdown items can wrap to multiple lines. Right now, a dropdown with a wrapping item ends up a little cut off. 
2. Check the behaviour when searching, see gif. Ideally, all results would be visible, up to a height maximum, but right now the menu is shrinking below the size of the list. 

@ItsMeBrianD feel free to weigh in also, but would it make sense to have the behaviour be something more like:   

1. More than [5] list items: we render the virtual list, and it has the same fixed height always. 
2. Less than [5] list items: we render a div, and we just let it fit its contents (e.g. no fixed height)? 

I think that would cover the multi-line and search scenarios while also removing the gap. 

CleanShot 2024-05-15 at 18 39 01@2x

CleanShot 2024-05-15 at 18 39 35

mcrascal (changes requested):
See comment

kwongz:
Thanks @mcrascal! Sound good, get working on this!

kwongz:
Added conditional rendering in the Dropdown.svelte file, Renders scroll-able Virtual List with 5 or more options

image

Dropdowns with less then 5 options will render DropdownOptionDisplay component with a regular div container to manage the height of the contents

Example 1.  - 4 options with Line breaking

image

Example 2.  - 2 options

image

packages/ui/core-components/.evidence-queries/sources/hashes.json
mcrascal, line 1: I believe this should be gitignored

mcrascal (approved):
Very excellent. ~No notes~. One note.

packages/ui/core-components/.evidence-queries/sources/hashes.json
kwongz: Thanks! Added the hashes.json to the gitignore, not too sure how it ended up there will speak to Brian about it tomorrow

mcrascal:
@kwongz Massive!

PR #2011, opened 2024-05-16
Interactive map components

hughess:
### Description
Adds interactive map components powered by Leaflet. Includes:
- **BaseMap** - composable map component, with ability to include multiple layers
   ```html
   <BaseMap>
      <Points/>
      <Areas/>
   </BaseMap>
   ``` 
- **AreaMap**
   ```html
	<AreaMap data={la_zip_sales} geoId="zipcode_field_in_geojson" areaCol="zip_code" />
   ```
- **PointMap**
   ```html
	<PointMap data={la_locations} lat="lat" long="long" />
   ```
- **BubbleMap**
   ```html
	<BubbleMap data={la_locations} lat="lat" long="long" size="sales" />
   ```

CleanShot 2024-05-16 at 11 12 00@2x

### Major Features
- **Drill-down links:** supply a `link` column and when the point/area is clicked, it will navigate to the link 
- **Inputs:** passing a `name` to the map will let you treat the map like an input component. When a point/area is clicked, the input will be set to that row of data. An individual value can be access using `{inputs.my_map_name.my_column_name}`
- **Custom tooltips:** the `tooltip` prop accepts an array of tooltip objects. 
   - This functionality closely resembles the DataTable Column component's options
   - Note this will be replaced by a friendlier component syntax in the future
   - E.g.,
       ```js
           tooltip={[
                 {id: 'store_name', showColumnName: false},
                 {id: 'sales', fmt: 'usd'}
          ]}
      ```
- **Links in tooltips:** within `tooltip`, you can use the `contentType` option to specify that a column is a `link`
   - To make the link clickable, you will need to use `tooltipType=click` (default is `hover`, where it's not possible to click the link) 

### Checklist


packages/ui/core-components/.evidence-queries/sources/hashes.json
mcrascal, line 1: @ItsMeBrianD I've seen these crop up in one of Kyle's PRs -- I think something is off in the gitignore settings on core-components.

sites/docs/pages/components/point-map.md
```diff
@@ -0,0 +1,455 @@
...
+    name=my_point_map 
+    height=200
+/>
+
+*Click a point on the map to see the input value get updated:*
+
+#### Selected value for `{inputs.my_point_map}`: 
+  
+{inputs.my_point_map}
+
```
mcrascal, line 219: Think this is leftover code -- just renders [object Object]

mcrascal (approved):

sites/docs/pages/components/point-map.md
hughess: Unfortunately this is the default selection for the input at the moment - since that line doesn't reference a specific property it shows the object. though JSON.stringify shows undefined strangely.

If you select an item on the map it should populate the object with the columns from that row

mcrascal: Ah, I didn't get that it included the whole row from the example. Maybe a couple additional column references would be more clear than the object? 

CleanShot 2024-05-17 at 11 38 06

hughess: thanks - this is different than what i was seeing before, but fixed now thanks to @ItsMeBrianD!
CleanShot 2024-05-17 at 13 12 30

PR #2032, opened 2024-05-27
Chromatic CI

mcrascal:


package.json
```diff
@@ -75,6 +75,7 @@
 		"dev:test-env": "pnpm --filter ./sites/test-env dev",
 		"dev:core-components": "pnpm --filter ./packages/ui/core-components package:watch",
 		"sources:example-project": "pnpm --filter ./sites/example-project sources",
+		"sources:core-components": "pnpm --filter ./packages/ui/coure-components sources",
```
ItsMeBrianD, line 78:
```suggestion
		"sources:core-components": "pnpm --filter ./packages/ui/core-components sources",
```

ItsMeBrianD:
@mcrascal what is currently blocking this?

We should enable the diffs in PRs before merging this as well

mcrascal:
@ItsMeBrianD I think we're almost good to go 

I'm going to pull in the latest on query debugger and see if I can make it fit. 

The other item on my list is the TW config -- I would really like to get that syncing into the storybook before we set a bunch of baselines, but I won't hold it if I can't get that working.

mcrascal:
@ItsMeBrianD I think we're good to go

ItsMeBrianD:
I think there's one more issue here - it looks like chromatic isn't running sources when building storybook. We probably need to have `npm run sources && npx chromatic ...` in the package.json

packages/lib/sdk/src/usql/query/Query.js
```diff
@@ -868,7 +892,9 @@ DESCRIBE ${this.text.trim()}
 			);
 
 		Query.#constructing = true;
+		console.log(opts.id);
```
ItsMeBrianD, line 895:
```suggestion
```

ItsMeBrianD (approved):

PR #2035, opened 2024-05-28
Fix/1833 slider input

kwongz:
### Checklist

image

A few notes, there is a Tick functionality that i think will make for a better user experience. Would like to try tackling to add that feature at a later date.

Would also like to try adding a live sql graph or table to connect the slider to.

- [ x] I have added to the docs where applicable


sites/docs/pages/components/slider.md
```diff
@@ -0,0 +1,103 @@
...
+````markdown
+<Slider
+    title="Months" 
+    name=monthsLarge
+    min=0
+    max=36
+    size=large
+/>
+````
+
```
archiewood, line 98: Would be great to add a proplisting table similar to https://docs.evidence.dev/components/dropdown/

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -0,0 +1,100 @@
...
+		return sizeMap[size.toLowerCase()] || 'w-40';
+	};
+
+	$: sizeClass = renderSize(size);
+</script>
+
+<HiddenInPrint enabled={hideDuringPrint}>
+	<div class={`relative ${sizeClass} mb-8`}>
+		<p class="pb-2 truncate text-xs">
+			{title} : <span class="text-sm">{$inputs[name]}</span>
```
archiewood, line 92: perhaps the title and the number should be the same font size?

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -0,0 +1,100 @@
...
+		</p>
+		<SliderShadcn {min} {max} {step} {ticks} bind:value />
```
archiewood, line 94: The slider position does not appear to currently be reactive to changes in the slider value

CleanShot 2024-05-28 at 15 29 47

I think the linking of two sliders to the same input `monthsLarge` in the docs was probably unintentional, but it does illustrate a reactivity quirk

archiewood (commented):
Loving the slider functionality. 

A few comments inline on UX

sites/docs/pages/components/slider.md
archiewood: (the checkbox is definitely a bit lean on docs content)

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
kwongz: interesting, resolving the linkage and taking a look at the reactivity quirk

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
kwongz: Will do!

sites/docs/pages/components/slider.md
kwongz: ill take a look at the dropdown table thanks!

kwongz: Props Table Added!

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
kwongz: fixed the Large and Full slider connection

Was not able to experience the reactivity quirk on my side, would like to talk more about it when we can.

Thanks!

archiewood: I think you will need to keep the erroneous linkage in order to reproduce the issue.

this was on Firefox, for what it's worth

mcrascal (approved):
This is looking great. 

A couple ideas: 

1. Can we apply the class `select-none` so that when you're dragging around the slider, it doesn't select it like text
2. Can we add support for the `fmt` prop on this (e.g. a percent slider, a $ slider etc)

kwongz:
Updated Slider

https://www.loom.com/share/a3c1eed75187401a87856303b8e8fce5?sid=ed1f5e87-bb49-46c1-8a5a-407e9a79adef

PR #2043, opened 2024-05-30
Remove Docker, Break out Motivation as own page

archiewood:
### Description

- Remove Docker
CleanShot 2024-05-30 at 13 29 43@2x

- Add VSCode marketplace deeplink
CleanShot 2024-05-30 at 13 30 54

- Motivation section to own page
CleanShot 2024-05-30 at 13 28 10@2x

CleanShot 2024-05-30 at 13 29 25@2x

### Checklist

### DIFF
--- a/sites/docs/pages/index.md
+++ b/sites/docs/pages/index.md
@@ -11,21 +11,7 @@ og:
 
 Evidence is an open source framework for building data products with SQL - things like reports, decision-support tools, and embedded dashboards. It's a code-driven alternative to drag-and-drop BI tools.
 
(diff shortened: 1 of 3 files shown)


hughess (approved):
I wouldn't be opposed to deleting that content entirely as well. Not sure how much it adds overall

archiewood:
I often read this kind of page when investigating a new tool, so I think it has a home.

Why another X?

For example it's the first thing I looked for on bits-ui / shadcn / tailwind when I found out about them

PR #2046, opened 2024-05-31
Added customizable Omitgroups prop to Date Range, updated docs: added…

kwongz:
… to Props table + Live example of new prop

### Description

Docs Additions
image

image

*** Question***
{"days, todate"} , {"days"} do they look too confusing?, should they just be normal text in markdown? (No Curlies?) 

Updated Stories

image

### Checklist

- [ x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)
- [x ] I have added to the docs where applicable


sites/docs/pages/components/date-range.md
```diff
@@ -138,6 +149,19 @@ A manually specified end date to use for the range
 Title to display in the Date Range component
 
 </PropListing>
+
+
+<PropListing 
+    name="omitGroup"
+    options="{["days", "months", "last", "todate"]}"
```
archiewood, line 156: I'm not sure this is quite the right UX.
1. I agree that the names are not quite right in lowercase. Either should be String or camelCase.
2. I think this should probably be opt in rather than opt out, if you specify anything.
3. There needs to be a mechanism that allows you to pick within groups. You might want to include last 30 days, but not last 7 days, for example.

something like

<DateRange
   name=name_of_date_range
   presets={["Last 7 Days", "Last 30 Days", "Last 3 Months", "Last Year"]}
   ...
/>

archiewood (changes requested):
Some UX considerations

kwongz:
-- Updated Docs --
image

image

ItsMeBrianD:
@archiewood can you take a look at this when you get a chance before you go OOO?

sites/docs/pages/components/date-range.md
```diff
@@ -138,6 +159,19 @@ A manually specified end date to use for the range
 Title to display in the Date Range component
 
 </PropListing>
+
+
+<PropListing 
+    name="presetRanges"
+    options={['last7Days', 'last30Days', 'last90Days', 'last3Months', 'last6Months', 'last12Months', 'lastMonth', 'lastYear', 'monthToDate', 'yearToDate', 'allTime']}
```
archiewood, line 166: these will not copy as expected as they will not include the curly braces or square braces

packages/ui/core-components/src/lib/atoms/inputs/date-range/DateRange.svelte
```diff
@@ -34,6 +34,8 @@
 	export let data;
 	/** @type {string | undefined} */
 	export let dates;
+	/** @type {[]string | undefined} */
+	export let presetRanges;
```
archiewood, line 38: allowing a single string as well would make it easier in some cases

kwongz:
@ItsMeBrianD 

Just seeing a conflict with next, im not confident it was one of my changes, was this something you updated?

DateRange.Svelte

image

ItsMeBrianD:
That QueryLoad was added in features b to prevent some flickering, I think you'll just want to bring your `{presetRanges}` prop into the lower copy there

Issue #1837, opened 2024-04-02
Upgrade `preview` command to match behaviour of express static server 


mcrascal:
## Feature Description

Current preview command differs from the behaviour of most deployment environments. It would be good to better match production environments when running the preview command.

archiewood:
# Workarounds

Use another preview server e.g.

- `npx http-server build`
- `npx serve build`

PR #2059, opened 2024-06-03
Better preview command

archiewood:
### Description

Fixes https://github.com/evidence-dev/evidence/issues/1837

### Checklist


packages/evidence/cli.js
```diff
@@ -368,8 +368,8 @@ prog
 		const flatArgs = flattenArguments(args);
 
 		logQueryEvent('preview-server-start', undefined, undefined, undefined, true);
-		// Run svelte kit dev in the hidden directory
-		const child = spawn('npx vite preview --outDir build --port 3000', flatArgs, {
+		// We will likely need to modify this for SPA mode previews
+		const child = spawn('npx serve build', flatArgs, {
```
mcrascal, line 372: `serve -s` for SPA mode

mcrascal (approved):

packages/evidence/cli.js
ItsMeBrianD: I would really prefer to just use their [API](https://www.npmjs.com/package/serve#API) instead of forking a child process, but if that's too much on this PR then this seems okay

archiewood: @mcrascal 
`serve -s` is correct for SPA mode, but it breaks non spa mode on navigation, so we will need to read the config option into this file to support it

PR #2061, opened 2024-06-04
updated agg prop to Value component

kwongz:
### Checklist

- [x ] I have added to the docs where applicable

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/viz/core/Value.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/core/Value.svelte
@@ -9,6 +9,8 @@
 	import EmptyChart from './EmptyChart.svelte';
 
 	export let data;
+	export let column;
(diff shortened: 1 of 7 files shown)


packages/ui/core-components/src/lib/unsorted/viz/core/Value.svelte
```diff
@@ -25,6 +27,10 @@
 
 	// Remove any undefined props (e.g. w/o defaults) to prevent them from being passed
 	$: spreadProps = Object.fromEntries(Object.entries($$props).filter(([, v]) => v !== undefined));
+
+	$: if (agg) {
+		data = data.groupBy(undefined).agg({ [agg]: { col: 'fare', as: [column] } });
+	}
```
ItsMeBrianD, line 33: What happens if I want to get the total (or avg) distance travelled by plane, rather than fare?

Issue #1690, opened 2024-03-05
Building sources hangs on last step


andrejohansson:
## Steps To Reproduce

Run `npm run sources` in a node:18 docker container on Windows or Mac.

### Environment

- Node version (`node -v`): v18.19.1
- npm version (`npm -v`): 10.5.0
- OS: node:18 (official docker image)
- Browser: Firefox Developer Edition
- Database: A lot

**package.json**

```json
  "dependencies": {
    "@evidence-dev/core-components": "^3.4.0",
    "@evidence-dev/evidence": "^28.0.0",
    "@evidence-dev/duckdb": "latest",
    "@evidence-dev/bigquery": "latest",
    "@evidence-dev/csv": "latest",
    "@evidence-dev/databricks": "latest",
    "@evidence-dev/mssql": "latest",
    "@evidence-dev/mysql": "latest",
    "@evidence-dev/postgres": "latest",
    "@evidence-dev/snowflake": "latest",
    "@evidence-dev/sqlite": "latest",
    "@evidence-dev/trino": "latest",
    "mermaid": "^10.7.0"
  },
```

**package-lock.json**

[package-lock.json](https://github.com/evidence-dev/evidence/files/14497205/package-lock.json)

## Expected Behavior

To be able to run `npm run sources` and have the command exit after it´s done processing.

## Actual Behaviour

We build our evidence in docker and the `npm run sources` command does not exit, it hangs on the last data source. It happens to be a Postgres DB  this time but it does not seem to matter if we remove it, then the command just hangs on next data source in line. We have tried both on a Windows host (WSL2) and on Mac host.

On our build machine (Linux host) the `npm run sources` command exits correct.

## Workarounds

Pressing CTRL+C to kill the command then running `npm run dev` works, as it seems to be done with all data processing anyway.

archiewood:
When did this start happening?

archiewood:
I see. This feels like an issue that will benefit from a minimal reproduction if you are able to supply one

andrejohansson:
Sure! Here is a minimal repro

[evidence-issue-1690.zip](https://github.com/evidence-dev/evidence/files/14507077/evidence-issue-1690.zip)

1. Created a fresh evidence project from template
1. dockerized it (see steps for running in the README)
1. Hangs with original duckdb source also

So...my suspicion about the proxy and sidecars seems to be irrelevant.

[...]

PR #2065, opened 2024-06-05
Feature: Switch to Datasources CLI provided by SDK

ItsMeBrianD:
### Description

**This PR contains a substantial change**

This moves the responsibility for `npm run sources` from `@evidence-dev/plugin-connector` into `@evidence-dev/sdk`. The interface for these commands is identical, so this should have no noticeable impact on UX or CI/CD

This has no impact on the settings UI (for now), and includes a legacy compatibility layer for using `evidence.plugins.yaml` rather than the goal of `evidence.config.yaml`.

This creates a pattern for incremental switch from `@evidence-dev/evidence`'s CLI to `@evidence-dev/sdk`'s CLI

### Checklist


ItsMeBrianD:
There is an error in this PR:

Source Query HMR is still using plugin-connector, and needs to be moved over to using SDK instead

mcrascal:
@ItsMeBrianD 

This is great 


1. The delay between printing ` ✅ Done!` and actually exiting the process is long enough that I am reaching for `ctrl c`
    1. Would it make sense to do a spinner or something with “wrapping up”, then print done when we are _truly_ done? 
2. `.evidence/meta` will be version controlled in user’s projects, which I think us unintentional. Unless the intent is to version control this, we should consider moving it to .evidence/template so that people don’t need to update their gitignores / accidentally pollute their version history 
3. Confirming changes to an existing datasource (in the settings UI) yields a pretty unclear print to the console (excerpted below for docs) 
    1. Looks like it runs twice 
    2. Skipping: filtered is very unclear 

```
✔ Loading plugins & sources
-----
  [Filtered] series_demo_source
-----
  [Processing] needful_things
  needful_things ⚠ Skipping: Filtered
  orders ⚠ Skipping: Filtered
-----
✔ Loading plugins & sources
-----
  [Filtered] series_demo_source
-----
  [Processing] needful_things
  needful_things ⚠ Skipping: Filtered
  orders ⚠ Skipping: Filtered
-----

```


**Not blocking**
1. This lists the duckdb filename as skipped on every run. Seems like we should be excluding those from the console log. 
2. Probably not in scope, but right now we don’t fail / provide a non zero exit code when a source includes an error. This seems off. 
3. Source HMR is great, but I can imagine for some of our users they’d want it off, or for some sources, they’d want it turned off. E.g. a config “run sources on save”

mcrascal (approved):

ItsMeBrianD:
> The delay between printing  ✅ Done! and actually exiting the process is long enough that I am reaching for ctrl c

This is being caused by the telemetry package - going to note this as out of scope but noted this in #1690

> .evidence/meta will be version controlled in user’s projects, which I think is unintentional. Unless the intent is to version control this, we should consider moving it to .evidence/template so that people don’t need to update their gitignores / accidentally pollute their version history

We should update this in the .gitignore for the template, and possibly ensure that it is being ignored properly as part of our startup process - could go either way on this. There are a number of things that _aren't_ part of the template, and should live in `.evidence` instead

> Confirming changes to an existing data source (in the settings UI) yields a pretty unclear print to the console (excerpted below for docs)
> Looks like it runs twice
Will take a look at this, it is probably running once for `connection.options` and once for `connection.options.yaml`

> Skipping: filtered is very unclear
Can update the verbiage here, I think this matches what was in the previous implementation as well

> This lists the duckdb filename as skipped on every run. We should exclude those from the console log.
Can change

> Source HMR is great, but some users would like it disabled globally or disabled for some sources.
#2124 

> It's probably not in scope, but right now, we don’t fail or provide a nonzero exit code when a source includes an error. This seems off.
This behavior has come up a few times, I am wondering if this should be part of strict mode?

PR #2069, opened 2024-06-05
Add docs components

hughess:


sites/docs/pages/components/accordion.md
```diff
@@ -3,40 +3,69 @@ title: Accordion
...
 ### Accordion
 
-<PropListing name="single" options={['true', 'false']}>
-
-When true, only a single accordian item can be open at once.
-
-</PropListing>
+  <PropListing
+      name="single"
+      description="When true, only a single accordian item can be open at once."
```
mcrascal, line 60: `PropListing` expects description content in a slot. 

I think if you replace line 37 in `PropListing` with: 

`<div id="markdown-slot"><slot > {description} </slot> </div>`

It'll work with both the description prop and the slot.

archiewood:
Not sure it matters, but there are probably a bunch of image files we are no longer using that could be deleted

hughess:
@archiewood ah yes, completely forgot about that. I'll remove those

hughess:
I tried removing an image I know is used to see if it would trigger a build error, but no luck. It seems tricky to figure out which ones are used vs. not

hughess:
nice looks like it worked! seeing no missing images on a check of the deployment

mcrascal (approved):

PR #2105, opened 2024-06-14
Weighted Mean behavior improvements

ItsMeBrianD:
### Description

This makes a very small change to how weightedMeans are calculated, and adds some unit tests to validate the behavior. If @hughess or @archiewood can validate that the unit tests are correct, we should be good to go here

### Checklist

### DIFF
--- a/.changeset/tall-hornets-remain.md
+++ b/.changeset/tall-hornets-remain.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Fix weighted mean behavior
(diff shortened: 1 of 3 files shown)


hughess:
This changes the intended behaviour, but it might actually be better.

**Initial intended behaviour:**
- If you don't pass a `weightCol`, the weight for each row will be set at 1 (even weighting)
- **Pros:** always works, will just calc the simple average if no weight supplied
- **Cons:** it might not be obvious that you need to supply a `weightCol`, so having a real result returned could be confusing (how is it calculating the weighted average without me telling it the column?)
- In the original implementation, it does not consider nulls or zeroes inside the `weightCol`
 
**Revised behaviour:**
- If you don't pass a `weightCol`, result will return 0
- **Pros:** should be immediate feedback that something is wrong
- **Cons:** 0 isn't the best indicator that something's wrong

One other thing we can do is check if `weightCol` is supplied at all, maybe returning `null` before even entering the loop that does the calculation

hughess (approved):

PR #2127, opened 2024-06-19
Datasource CLI Tweaks

ItsMeBrianD:
### Description

This addresses issues noted in #2065 

- The eager loading notification for large files (mostly CSV and DuckDB files) is behind the debug flag, and will not show the warning
- To manage the `no results returned` for some `needful_things` table, we need to modify the duckdb connector to an advanced connector and skip duckdb files, or find some way for simple connectors to indicate that a file is not a query. Either solution (or both) would be okay
- Verbiage has changed for skipped queries / sources
image

- `[Filtered] source name` has been hidden behind the debug flag, the filtered sources are now only presented as a count at the end by default to reduce noise

- `Skipping: Filtered` for queries has been changed to `This query was not run due to filters`

- `Skipping: Cached` for queries has been changed to `This query was not run because it is already cached`

### Checklist

- [ ] ~~I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)~~ (piggybacks the existing datasource changes)

### DIFF
--- a/packages/lib/sdk/src/build-dev/vite/datasource-plugins/source-query-hmr.js
+++ b/packages/lib/sdk/src/build-dev/vite/datasource-plugins/source-query-hmr.js
@@ -4,6 +4,7 @@ import { dataDirectory, metaDirectory, sourcesDirectory } from '../virtuals/node
 import { updateManifest } from '../../../plugins/datasources/updateManifest.js';
 import { ProcessingQueue } from '../../../lib/processing-queue.js';
 import { VITE_EVENTS } from '../constants.js';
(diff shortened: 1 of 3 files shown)


mcrascal (changes requested):
Nice, two small nits: 

1. Can we remove the little warning icon associated with skipping queries? This is expected behaviour. 
2. Can we replace `This query not run due to filters` with `Skipped`; and `This query was not run because it is already cached` with `From Cache`?

mcrascal (approved):

PR #2130, opened 2024-06-19
Upgrades and fixes for vscode extension

hughess:
## Evidence VS Code Extension v1.5.0
Several new features for writing queries and managing large markdown files.

### Autocomplete for SQL queries
CleanShot 2024-06-20 at 08 35 11

### Improved syntax highlighting
- Better handling of nested components and indenting
- Headers stay highlighted when indented

CleanShot 2024-06-19 at 12 46 13

### Background shading for blocks
When you add any of the below elements to your markdown page, the extension will shade the background to make it clear where that element starts and stops:
- Frontmatter
- SQL queries
- Each blocks
- If blocks

This behaviour can be turned off in the Evidence extension settings in the VS Code settings menu.

CleanShot 2024-06-19 at 13 00 09

### Collapsible blocks
You should now see an arrow at the top of large blocks of code on your page - you can click this to collapse that section, making your page easier to read and navigate.  

CleanShot 2024-06-19 at 12 51 59

This also works well for nested elements:

CleanShot 2024-06-19 at 12 54 28

### Schema Viewer
The built-in schema viewer now lets you copy the column names from your schema and paste them into your markdown

CleanShot 2024-06-19 at 12 58 39

### Fixes
- Fixes issue with templated page command on Windows
- Adds missing components to the slash command list

### DIFF
--- a/packages/extension/language-configuration.json
+++ b/packages/extension/language-configuration.json
@@ -20,12 +20,11 @@
 		["\"", "\""]
 	],
 	"folding": {
-		"offSide": true,
 		"markers": {
-			"start": "^\\s*<!--\\s*#?region\\b.*-->",
-			"end": "^\\s*<!--\\s*#?endregion\\b.*-->"
+			"start": "^\\s*(```\\w+|{#if\\b|<s*\\b|<s*\\b|{#each\\b).*$",
+			"end": "^\\s*(```|{/if}|/>|<\/\\S*>|{/each})$"
 		}
-	},
+	},  
 	"wordPattern": {
 		"pattern": "(\\p{Alphabetic}|\\p{Number}|\\p{Nonspacing_Mark})(((\\p{Alphabetic}|\\p{Number}|\\p{Nonspacing_Mark})|[_])?(\\p{Alphabetic}|\\p{Number}|\\p{Nonspacing_Mark}))*",
 		"flags": "ug"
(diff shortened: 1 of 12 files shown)


mcrascal (approved):

archiewood:
this is awesome

i like the copy command on the columns.

it makes me think of an interesting thing you could add to the table:

perhaps there is an insert button next to tables which if you click, inserts the following at the current cursor position

```
needful_things
`--orders   insert
```

````markdown
```sql orders
select * from [needful_things].orders
```
````

or it could copy it, if inserting is not practical

hughess:
That’s an interesting idea for inserting - we’d be able to do that.

i actually just added in the copying of tables earlier! It copies in `source.table`

hughess:
Added that command for tables and columns - pretty nice!

CleanShot 2024-06-20 at 07 47 06

This is also making me think that we could use the list of columns/tables in the schema viewer to create a crude version of intellisense - give suggestions from that list while typing. if you had a lot of tables it could be hard, but we might be able to put the table name next to each column so you could see if you're in the correct one

hughess:
We are able to use the list to give suggestions inside SQL:

CleanShot 2024-06-20 at 08 35 11

hughess:
At the moment, the autocomplete suggestions trigger on `.`, `(`, and spaces. Would be nice to have it trigger on letters as well, but haven't been able to figure that out yet.

Issue #1889, opened 2024-04-15
Link support in BigValue


justin0mcateer:
## Feature Description

I would like to be able to define a link to a more detailed page that would be accessed by clicking on a BigValue control.

## Goal of Feature

This would allow users to have a dashboard with a high-level set of KPIs then 'dig in' to understand the 'why'.

## Current Solution / Workarounds

I believe the best current solution will be to use a LinkButton in combination with the Group element that is part of the experimental Grid control. In this case each BigValue will have a 'Details' button immediately below which provides the 'link' functionality.

## Alternatives

Any control that supports linking could be used in place of LinkButton, but it would be much cleaner to just allow clicking the BigValue itself to 'drill down'.

PR #2134, opened 2024-06-20
BigValue url prop

kwongz:
### Checklist

Added url prop to BigValue to link to another page
image

on hover
image

image


kwongz:
Updated prop from url to href to match other docs

OLD:
image

UPDATED:
image

hughess:
I feel pretty strongly that we should use `url` instead of `href` when we refer to links, as i don't think `href` is a known term among data analysts who have not already got some experience in web development

kwongz:
@hughess sounds good, reverting to url.

kwongz:
Updated Docs
image

image

kwongz:
resolves #1889

ItsMeBrianD (approved):
This LGTM pending @hughess

hughess:
Sorry I hadn’t noticed the context of this feature when I looked - that was just my knee jerk reaction to href…

for consistency with other components which have click through features, I think the prop should be called “link” and the big value itself should become clickable instead of adding the details underneath.

to show it’s clickable I would do a change in color on hover with the pointer cursor

kwongz:
1. @hughess Sounds good, one potential issue with making the entire BigValue clickable is we lose access to the small graph in BigValue. There is probably a work around, just want to ask if keeping the details button is a deal breaker? 

I think it is a nice add-on in addition to all the other add-ons that can be attached to BigValue.  @ItsMeBrianD thoughts?

BigValue Clickable(wrapped in a tag): can't access small graph inside
image

Current BigValue in docs
image

--------------------------------------------------------------------------------------------------------------------------------------

2. with your permission I would also like to standardize the "link' prop in BigLink and LinkButton

image
image

hughess:
I would probably only wrap the `<Value/>` component with the clickable element. To match with our other viz components, the user experience is to click the thing you want more info on - so if you saw 120 orders in a big value, you'd be able to click the 120 to see which ones they are (or get taken to a page that has more info for you)

kwongz:
I like the change! 

This is ready to merge if you like it @hughess 

LGTM
value on hover:
image

still access to other viz components
image

p.s
I'm wondering if we should add wrappable anchor wrappable options for the other viz components. It would be a pretty verbose prop library for each viz component

kwongz:
@ItsMeBrianD Not too sure why Netlify tests are failing.

kwongz:
image


image

hughess (approved):
Looks good! We might want to adjust the styling on hover later (e.g., maybe make the font blue on hover rather than changing the background), but we can decide that when we implement the same thing for the Value component.

PR #2137, opened 2024-06-20
Exclude base map story from chromatic snapshots

mcrascal:
### Description

Excludes the base map story from chromatic snapshots. 

The base map story alternates between plotting the orange la locations at the top of the layers and at the bottom, which generate spurious diffs. 

CleanShot 2024-06-20 at 10 29 16


ItsMeBrianD (approved):
FWIW this seems like an actual bug

mcrascal:
@ItsMeBrianD opened #2140

PR #2139, opened 2024-06-20
updated docs with selectAllByDefault

kwongz:
### Checklist

image

image

### DIFF
--- a/sites/docs/pages/components/dropdown.md
+++ b/sites/docs/pages/components/dropdown.md
@@ -216,7 +216,7 @@ Selected: {inputs.category_multi_default.value}
     name=name_of_dropdown
     value=column_name
     multiple=true
-	defaultValue={['value1', 'value2']}
+	defaultValue={['Sinister Toys', 'Mysterious Apparel']}
 />
 ````
 
--- a/sites/docs/pages/components/dropdown.md
+++ b/sites/docs/pages/components/dropdown.md
@@ -64,6 +64,7 @@ Selected: {inputs.category3.value}
     name=category4
     value=category_name
     title="Select a Category"
+    defaultValue="%"
 >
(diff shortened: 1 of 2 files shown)


sites/docs/pages/components/dropdown.md
```diff
@@ -76,14 +77,14 @@ Selected: {inputs.category4.value}
     name=category4
     value=category_name
     title="Select a Category"
+    defaultValue="%"
```
ItsMeBrianD, line 80: Why are these additions needed?

kwongz:
I also added explicit targeting of "%" by applying the defaultValue prop. The previous example selected the "%" without using the defaultValue prop. 

But if we are trying to suggest to the user using the defaultValue to be applied automatically I can remove it.

OLD:
image

NEW:
image

kwongz:
Removed sql filter query from unrelated multiselect sections

Removed in Multi-select:
image

Removed in Multiple defaultValues section
image

sites/docs/pages/components/dropdown.md
kwongz: Not needed, but my initial reaction to seeing the defaultValue section, is to see the use of it in the example.

Currently, we have other examples of how it can be used in other sections of dropdown, but the actual defaultValue example in the docs does not show the prop being used. 

I was just curious if this would be a more explicit example for new users.

OLD:
image

NEW:
image

archiewood: @kwongz this example using the `<DropdownOption value=%>` is a workaround from before we had the multiselect dropdown, which had the "select all" button and a corresponding property.

I would consider removing the slot from the example and just use the prop

kwongz:
New Default Value section with changes @archiewood @ItsMeBrianD 

image

archiewood (approved):

PR #2163, opened 2024-06-26
[WIP] Timezones

zachstence:
### Description

This PR will have several changes to make timezones easier to understand and work with in Evidence.

This has been cut off as a feature branch while we await [a new tagged version of DuckDB WASM](https://github.com/duckdblabs/evidence/issues/3), which will support the ICU extension

### Next steps

@archiewood Has made a diagram for the eventual timezone behavior we should work toward (may change as we get deeper into the woods)
image

- @archiewood @hughess QA how this change behaves with different use cases and different databases
- @archiewood Document what changes the user should make to their data/sources to accommodate this breaking change
- @archiewood Create a page in the evidence docs for Timezones as a whole

### DIFF
--- a/.changeset/afraid-parrots-hammer.md
+++ b/.changeset/afraid-parrots-hammer.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/sdk': patch
+'@evidence-dev/universal-sql': patch
+---
+
+Add buildOptions.batchSize to connection.yaml spec
--- a/.changeset/brave-students-joke.md
+++ b/.changeset/brave-students-joke.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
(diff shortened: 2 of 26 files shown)


archiewood:
I've started adding tests for each data source.
- [x] Snowflake
- [x] BigQuery
- [x] Postgres
- [x] DuckDB
- [ ] Databricks
- [ ] MySQL
- [ ] MS SQL
- [ ] Redshift
- [ ] Trino

These are helpful as they also allow you to confirm the syntax a user needs to use to convert their timezones to UTC in source queries.

I've done a v1 docs page, but want to add the user conversion syntax too

ItsMeBrianD:
@zachstence @archiewood what's the status of this?

archiewood:
awaiting tagged release of DuckDB wasm so we can add ICU support

Issue #2073, opened 2024-06-06
[Feature]: `ReferencePoint`


archiewood:
### Description

![CleanShot 2024-06-06 at 13 01 55@2x](https://github.com/evidence-dev/evidence/assets/58074498/81927c04-e1b7-4095-9394-9baed4408ed3)

https://jsfiddle.net/ow9zdakf/53/

### How should it work?

```svelte
<BarChart
   data={data}
   x=day
   y=sales
>
    <Callout 
        x='Tue'
        label='Sales really ripped..."
   />
</BarChart>
```

PR #2164, opened 2024-06-26
Add `ReferencePoint` and `Callout` components

zachstence:
### Description

Adds `ReferencePoint` and `Callout` components with stories and docs

Resolves https://github.com/evidence-dev/evidence/issues/2073

### Checklist


zachstence:
`ReferencePoint`
image

`Callout`
image

archiewood:
lovely!

so the solution if you have a short string and you want it to not have extra space in the callout is to do `width=0`?

zachstence:
> lovely!
> 
> so the solution if you have a short string and you want it to not have extra space in the callout is to do `width=0`?

@archiewood Hmm actually it would be `width=undefined`, which isn't great... Maybe we should default to no width for the Callout too?

zachstence:
@archiewood Or it could be `width=fit`

hughess:
This is really nice!

It looks like there might be an issue with the default yellow color - getting some flickering on hover. In the past, this has been caused by echarts not knowing how to adjust the color for the hover state (can't remember exactly what triggered it, but something to do with the color code):

CleanShot 2024-07-02 at 12 36 45

hughess:
Only other point of feedback I have on the styling is: I think the grey of the label border is too different vs the grey of the text. 

Might be best to line up with the grey we use for the tooltip border, or maybe a shade darker (+ maybe match the same border radius, slightly less than the current callout radius). That way the greys are on the same scale

archiewood:
Overall like. Good design

QA
- Still getting flicker on the yellow point in the docs
   - Weird hover stuff?
  CleanShot 2024-07-02 at 19 35 52

- maybe it's just a browser thing but I'm seeing almost a drop shadow top-left behind the callout (macos, firefox)
CleanShot 2024-07-02 at 19 32 54@2x
 
- Do i need to be able to pass an array to labelposition to fix stuff like this
CleanShot 2024-07-02 at 19 36 29@2x

- Cursor should probably not be pointer as no action on click
CleanShot 2024-07-02 at 19 43 52

zachstence:
@archiewood Thank you for the feedback!

> Weird hover stuff?
> Cursor should probably not be pointer as no action on click

Fixed by making reference points not hoverable (this is in line with reference area/line functionality)

> maybe it's just a browser thing but I'm seeing almost a drop shadow top-left behind the callout (macos, firefox)

It is likely just something with how ECharts is rendering the one pixel border, it may have been on a pixel border on your monitor resulting in it looking a little thicker. I don't think there's anything I can do to control that.

image

> Do i need to be able to pass an array to labelposition to fix stuff like this for just one of my labels?

I don't know... I feel like if you're using data you would have a dynamic number of items and therefore configuring via a prop wouldn't make much sense. Options I see here:
- Have `labelPosition` accept a column from `data` to use for positioning each point
- The user just has to use `width`/`labelPosition`/add line breaks to try to get the label to not overflow in any spot

I wish ECharts had some way to detect/prevent overflow, but I don't think they do

archiewood:
> Fixed by making reference points not hoverable (this is in line with reference area/line functionality)

it still greys out the Callouts and ReferencePoints when you hover over the line, which is different to the ReferenceLine & Area behaviour

CleanShot 2024-07-03 at 11 28 34

PR #2184, opened 2024-07-03
Refactor `ReferenceLine`/`ReferenceArea` to use a store, prop parity with `ReferencePoint`

zachstence:
### Checklist


packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceArea/ReferenceArea.svelte
```diff
@@ -3,36 +3,173 @@
...
+			labelBackgroundColor,
+			labelBorderColor,
+			labelBorderWidth: toNumber(labelBorderWidth),
+			labelBorderRadius: toNumber(labelBorderRadius),
+			labelBorderType,
+			fontSize: toNumber(fontSize),
+			align,
+			bold: toBoolean(bold),
+			italic: toBoolean(italic)
+		});
```
ItsMeBrianD, line 159: Similarly, we haven't overloaded the `set` function in the other stores, but have had more specific functions that describe what the behavior is

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceArea/reference-area.d.ts
```diff
@@ -0,0 +1,61 @@
+// @ts-check
+
+import { Writable, Readable } from 'svelte/store';
+import type { MarkAreaComponentOption } from 'echarts';
+
+import type { Color } from '../colors.js';
```
ItsMeBrianD, line 6: Not sure about just calling this `Color` - it isn't specific here, maybe `ReferenceColor`?

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceArea/reference-area.store.js
```diff
@@ -0,0 +1,215 @@
...
+import { COLORS } from './constants.js';
+import { isPresetColor } from '../types.js';
+import checkInputs from '@evidence-dev/component-utilities/checkInputs';
+
+/**
+ * @param {import('svelte/store').Writable<any>} propsStore
+ * @param {import('svelte/store').Writable<any>} configStore
+ * @returns {import('./reference-area.js').ReferenceAreaStore}
+ */
+export const createReferenceAreaStore = (propsStore, configStore) => {
```
ItsMeBrianD, line 14: See above comments

packages/ui/core-components/src/lib/utils.js
```diff
@@ -62,10 +62,20 @@ export const flyAndScale = (node, params = { y: -8, x: 0, start: 0.95, duration:
...
 };
+
+/**
+ * @param {unknown} value
+ * @returns {boolean | undefined}
+ */
+export const toBoolean = (value) => {
+	if (typeof value === 'undefined') return undefined;
+	if (value === 'true') return true;
+	if (value === 'false') return false;
```
ItsMeBrianD, line 79: Should this be case-insensitive

zachstence: [4dc833b](https://github.com/evidence-dev/evidence/pull/2184/commits/4dc833b7648890167e6b79f52caa0767d55aa3c4)

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceArea/reference-area.d.ts
zachstence: [6e5ddf9](https://github.com/evidence-dev/evidence/pull/2184/commits/6e5ddf96090176b0f3aabae9bf971072ea90ceff)

zachstence:
@ItsMeBrianD Let's have a conversation about what we want to do with the stores. I'm not sure that converting them to classes (or 1 class) like `EvidenceMap.js` is the way to go:

- There is no parent reference component that manages children (like `Map` and `Point`, `MapArea`, etc). This isn't a technical limitation, but moreso a pattern I don't think fits here
- The map markers aren't reactive, and the references are. We need to update their position/styling when the incoming props change. This would require some additional thought of how to handle that with a class/classes.
- Seems unnecessarily complex, all we're really doing is converting from our props to ECharts for the references, whereas map does a lot more

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceLine/_ReferenceLine.svelte
```diff
@@ -1,271 +0,0 @@
...
-	export let label = undefined;
-
-	export let color = undefined;
-	export let lineColor = undefined;
-	export let labelColor = undefined;
-	export let lineWidth = undefined;
-	export let lineType = 'dashed'; // solid, dashed, or dotted
-	// Override arrow symbol with custom svg path which is a nice aspect triangle
-	let arrowPath = 'path://M0,10 L5,0 L10,10 z';
-	export let symbol = 'none';
```
archiewood, line 30: This prop has been removed which is a breaking change - it was a symbol appearing at the end of the line

(The default arrow was nasty so I replaced it, open to changing this, but we should definitely still have some kind of support for the symbol)

archiewood (changes requested):
Overall works well aside from one unintentional breaking change

We will also need to update the docs reference for RefLine/Area where additional props are available

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceLine/_ReferenceLine.svelte
zachstence: I had removed this prop because it wasn't working properly and I thought it was just some leftover code

https://github.com/evidence-dev/evidence/blob/reference-line-symbol-doesnt-work/packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceLine/ReferenceLine.stories.svelte#L52

No symbol is rendered anywhere

image

I can re-add and fix this functionality

zachstence:
> We will also need to update the docs reference for RefLine/Area where additional props are available

@archiewood This is done, see here: https://github.com/evidence-dev/evidence/pull/2184/files#diff-7febc53116f025c7fc408b69f95158907bf9d25613d5591c380975df9ff2c178

packages/ui/core-components/src/lib/unsorted/viz/references/ReferenceLine/_ReferenceLine.svelte
archiewood: The symbol is only rendered at the end of the line when you specify a y2 point - allows you to use as an arrow

zachstence: Added support and docs for `symbol`, `symbolStart`, and `symbolEnd` here
[c91ba01](https://github.com/evidence-dev/evidence/pull/2184/commits/c91ba010a41088b72265f4f4a3c01b40030566e3)

ItsMeBrianD (approved):

archiewood (approved):
Looks great. Easy review with storybook

PR #2194, opened 2024-07-08
Conditonal rendering value

kwongz:
### Checklist

Stories
image

Docs:
image

image

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/viz/core/_Value.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/core/_Value.svelte
@@ -33,13 +33,18 @@
 	let selected_value;
 	let error;
 
-	// conditonal font color
+	// Value Styling Props:
 	export let color = undefined;
-	// export let condition = undefined;
-	let valueColor = undefined;
+	export let redNegative = undefined;
+	let fontColor = '';
+	// export let valueClass = undefined;
 
-	$: if (color) {
-		valueColor = `text-${color}-500`;
+	$: if (redNegative || color) {
+		if (redNegative && selected_value < 0) {
+			fontColor = 'text-red-500';
+		} else if (color) {
+			fontColor = color;
+		}
 	}
 
 	let columnSummary;
@@ -110,7 +115,7 @@
 		>[{placeholder}]<span class="error-msg">Placeholder: no data currently referenced.</span></span
 	>
 {:else if !error}
-	<span class={valueColor}>
+	<span class={fontColor} style="color: {fontColor}">
 		{formatValue(selected_value, format_object)}
 	</span>
 {:else}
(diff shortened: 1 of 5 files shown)


hughess:
This looks great! A couple things to check:

- Is the red used for `redNegatives` the same as the one in `DataTable`?
- I would change the docs to match with the description of other color props in the product (there's a few in the `DataTable` docs)

kwongz:
Updated red color to same in dataTable --> tableRow.svelte
image

@hughess in terms of the color props, the most similar prop that behaves the same as Value's would be the color prop in annotations.svelte

_annotations color prop:_
image
image

The Value Color Prop, can take in the same color options "options="CSS name | hexademical | RGB | HSL"

I'm also wondering if we should keep the title "**Color**" instead of "**Colour**" since the css and out prop is in American Spelling.

Let me know what you think and i can update the Value prop, and the title spelling in annotations if necessary

kwongz:
@hughess I noticed Red Negative is listed in the data table docs, but missing from its prop list. Can we use Value's redNegatives prop list in Data Table

_Datatable proplist_
image

_Value prop list:_
image

hughess:
Color is good for consistency

Ah good catch, yes you can use the same one as Value. But it should go under the `Column` section of the `DataTable` page.

Also looks like there might be a hanging closing tag on the DataTable page under the prop listing before `scaleColumn`

kwongz:
Red negative added to column section in data tables
image

Removed /> above scaleColumn 
image

packages/ui/core-components/src/lib/unsorted/viz/core/_Value.svelte
```diff
@@ -101,7 +115,7 @@
 		>[{placeholder}]<span class="error-msg">Placeholder: no data currently referenced.</span></span
 	>
 {:else if !error}
-	<span>
+	<span class={fontColor} style="color: {fontColor}">
```
ItsMeBrianD, line 118: Why is `fontColor` being used as both a class and a style?

kwongz: thanks, originally we were trying to use tailwind styles, but then we decided to go with inline styling, will remove the class

kwongz:
Value Docs:
image

DataTable Docs: 
image

hughess:
Last thing - I think the rednegatives prop needs to be updated to accept strings for “true” and “false” for consistency with other components. The Column component rednegatives prop should already work like that.

goal with that is to remove the need for user to type curly braces to get Boolean values in the syntax - looks more approachable to data analyst, faster to type.

We should make this consistent across the product where Booleans are used in props

zachstence:
> We should make this consistent across the product where Booleans are used in props

I've got this on my ToDo list for a general improvement to all core components

ItsMeBrianD (approved):
If the behavior here is good to go (@hughess @archiewood) then I think we're set, the code looks good to me

hughess (approved):
Works great!

PR #2195, opened 2024-07-08
Updated Dimension grid take in queries with spaces

### LINKED ISSUE #2192: [Bug]: Dimension Grid cannot handle spaces in column names
### Describe the bug

Dimension Grids will error out in columns have spaces in 

CleanShot 2024-07-05 at 11 42 34@2x

### Steps to Reproduce

````markdown
```sql test_data
select 'United States' as "country name", 100 as population union all
select 'Canada', 50
```

<DataTable data={test_data}/>

<DimensionGrid data={test_data}/>

````

https://github.com/archiewood/dimension-grid-bug/blob/main/pages/index.md?plain=1

### Logs

_No response_

### System Info

_No response_

### Severity

You can workaround by aliasing columns to "column_name"

### LINKED ISSUE #2193: [Feature]: Support selecting multiple dimensions in dimension grid
### Description

Support selecting multiple of the same dimension in dimension grid

### What problem would this solve?

I want to select three provinces as a group and include them in a filter selection

### How should it work?
UI:
Either `shift click` enables selecting multiple dimensions or just clicking a second one is supported. 

API: 
boolean `multiple` prop defaulted to false, same as drop down

input store: 
multiple province dimensions selected: `(province = 'ontario' or province = 'quebec') and language = 'en'`

### PR DESCRIPTION
**Pull request by kwongz**
### Description

fix #2192 

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionGrid.stories.svelte
+++ b/packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionGrid.stories.svelte
@@ -100,3 +100,15 @@
 		})}
 	/>
 </Story>
+
+<Story name="string query with spaces">
+	<DimensionGrid
+		data={Query.create(
+			`SELECT "fare" as "fare price", "airline" as "air line" from flights limit 10`,
+			query,
+			{
+				disableCache: true
+			}
+		)}
+	/>
+</Story>
--- a/packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionGrid.stories.svelte
+++ b/packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionGrid.stories.svelte
@@ -104,7 +104,7 @@
 <Story name="string query with spaces">
 	<DimensionGrid
 		data={Query.create(
-			`SELECT "fare" as "fare price", "airline" as "air line" from flights limit 10`,
+			`SELECT "fare" as "fare price", "airline" as "Airline Name" from flights limit 10`,
 			query,
 			{
 				disableCache: true
--- a/packages/ui/core-components/src/lib/molecules/dimension-grid/_DimensionGrid.svelte
+++ b/packages/ui/core-components/src/lib/molecules/dimension-grid/_DimensionGrid.svelte
@@ -17,7 +17,7 @@
 	/** @type {string} */
 	export let name;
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by kwongz** [comment]
_Evidence project example_
image

_Storybook Example_
image

File: packages/ui/core-components/src/lib/molecules/dimension-grid/_DimensionGrid.svelte [hunk 1]
```diff
@@ -17,7 +17,7 @@
 	/** @type {string} */
 	export let name;
 
-	$: dimensions = data?.columns?.filter((col) => col.column_type === 'VARCHAR');
+	$: dimensions = data?.columns?.filter((col) => col.column_type === 'VARCHAR' || 'DOUBLE');
```
mcrascal, line 20: Filtering for varchar columns is the intended behaviour.

mcrascal (changes requested):

packages/ui/core-components/src/lib/molecules/dimension-grid/_DimensionGrid.svelte
kwongz: Thank You, reverted back to filter VARCHAR only.

Clarification:
So in the dimension grid example above with Fare Price and Airline, without the "DOUBLE" included filtering, the desired result is only to display Air line?

mcrascal: Yeah exactly

packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionGrid.stories.svelte
```diff
@@ -100,3 +100,15 @@
 		})}
 	/>
 </Story>
+
+<Story name="string query with spaces">
+	<DimensionGrid
+		data={Query.create(
+			`SELECT "fare" as "fare price", "airline" as "air line" from flights limit 10`,
```
mcrascal, line 107: Minor, but "air line" looks like a typo -- could we change to some other column name that also includes a space?  

```suggestion
			`SELECT "fare" as "fare price", "airline" as "Airline Name" from flights limit 10`,
```

kwongz: Nice, I like this better!

mcrascal (approved):

Issue #2141, opened 2024-06-20
[Bug]: Snowflake Connector does not pass proxyHost & proxyPort specified in connector options


jtshafer:
### Describe the bug

Our Snowflake instance, for security reasons, requires user/pass auth to go through a proxy. These configuration params are currently supported by the Snowflake node.js driver but are ignored and not passed when supplied in the options block in connection.yaml

https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-options#label-nodejs-connection-options
https://github.com/evidence-dev/evidence/blob/next/packages/datasources/snowflake/index.cjs#L202-L219 

We will to fix this in order to proceed

### Steps to Reproduce

Try to connect to a Snowflake instance (that requires a proxy) via user/pass 

### Logs

<img width="822" alt="Screenshot 2024-06-20 at 12 20 35 PM" src="https://github.com/evidence-dev/evidence/assets/9361420/c519cdd9-5a3c-4e60-9a87-7c4e9e9d028d">

### Severity

blocking production usage of Evidence

### Additional Information, or Workarounds

Locally, I am using externalbrowser as the authenticator, for now.

PR #2200, opened 2024-07-10
Add Proxy config section to snowflake connector

jtshafer:
### Description

- adds a new section to snowflake connection options for a proxy as specified here https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-connect#label-nodejs-proxy-connection. The 5 fields are stored as a nested block in `connection.yaml` but need to be passed into the snowflake driver as top level fields.

Setting proxy.username and proxy.password to `secret` so they are stored separately in `connection.options.yaml`

Screenshot 2024-07-12 at 9 11 17 AM

Screenshot 2024-07-10 at 1 56 52 PM

Screenshot 2024-07-10 at 1 57 01 PM

### Checklist

Fixes issue - https://github.com/evidence-dev/evidence/issues/2141


packages/datasources/snowflake/index.cjs
```diff
@@ -153,6 +153,10 @@ const getCredentials = (database = {}) => {
 	const role = database.role;
 	const schema = database.schema;
 
+	// optional
```
jtshafer, line 156: https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-connect#label-nodejs-proxy-connection

packages/datasources/snowflake/index.cjs
```diff
@@ -381,5 +382,47 @@ module.exports.options = {
...
+					title: 'Host',
+					type: 'string',
+					secret: false,
+					required: true
+				},
+				port: {
+					title: 'Port',
+					type: 'number',
+					secret: false,
+					required: false
```
jtshafer, line 404: maybe this should be required, but I would assume there is a default port.

ItsMeBrianD: I'm not entirely sure I follow, but there isn't a default specified here, does the snowflake library provide a default when this is used?

jtshafer: ahh nvm if you pass hostname port is also required. This is the error log they spit out `[ ! ] Error connecting to datasource data_warehouse: A proxyPort must be specified.`

jtshafer: here is the connector code for checking proxy params https://github.com/snowflakedb/snowflake-connector-nodejs/blob/cba00c1848f875b8e2054cf485bd9f507a7eee9c/lib/connection/connection_config.js#L213-L271

archiewood (approved):
Thanks for the contribution!

I haven't tested via a Snowflake authenticated proxy as I don't have access to one, but this UI seems sensible and compatible with existing Snowflake connections

jtshafer:
> Thanks for the contribution!
> 
> I haven't tested via a Snowflake authenticated proxy as I don't have access to one, but this UI seems sensible and compatible with existing Snowflake connections

Yea not sure how often this will be required (though likely more-so for enterprise users). Our instance requires proxy & host only for now so that's what I tested with

ItsMeBrianD (approved):
Looks great!

archiewood:
Thanks @jtshafer - this will get shipped with our next release. 
(provisionally end of this week)

PR #2224, opened 2024-07-16
Area, Point, Bubble Maps Drill Down + Invisible Links

kwongz:
### Checklist


kwongz:
merged feature branch in, still getting error 

image

Updated branch with pnpm install --no-frozen-lockfile, but still getting the error in the picture above

@ItsMeBrianD @zachstence

archiewood:
@kwongz if you haven't already tried, `pnpm i` and merging the result sometimes fixes this

zachstence:
@kwongz Looks like its working now. I just did `pnpm i --ignore-scripts` and committed/pushed the updated `pnpm-lock.yaml`.

kwongz:
Thanks All!

packages/ui/core-components/src/lib/unsorted/viz/map/BaseMap.svelte
```diff
@@ -74,6 +87,18 @@
 		</div>
 	</div>
 {/if}
+{#if link}
+	<QueryLoad {data} let:loaded>
+		<svelte:fragment slot="skelton"></svelte:fragment>
```
zachstence, line 92: Is `skelton` supposed to be `skeleton`?

packages/ui/core-components/src/lib/unsorted/viz/map/BaseMap.svelte
```diff
@@ -74,6 +87,18 @@
...
+		<InvisibleLinks data={loaded} {link} />
+	</QueryLoad>
+{/if}
+{#if linkCol}
+	<QueryLoad {data} let:loaded>
+		<svelte:fragment slot="skelton"></svelte:fragment>
```
zachstence, line 98: Is `skelton` supposed to be `skeleton`?

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -199,8 +199,12 @@ export class EvidenceMap {
 						this.lastSelectedLayer = layer;
 					}
 					onclick(feature);
+				});
+				layer.on('dblclick', () => {
+					//handles map drilldowns
 					if (link) {
 						window.location.href = link;
+						this.map.doubleClickZoom.disable();
```
zachstence, line 207: It seems a little weird to disable the `doubleClickZoom` only after the user has double clicked the map. Maybe this should happen when the map is created if they provide links?

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -264,9 +268,12 @@ export class EvidenceMap {
 				onclick(marker);
 			}
 			marker.bringToFront();
-
+		});
+		marker.on('dblclick', () => {
+			//handles map drilldowns
 			if (link) {
 				window.location.href = link;
+				this.map.doubleClickZoom.disable();
```
zachstence, line 276: See other comment

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -301,6 +308,7 @@ export class EvidenceMap {
 	 */
 	buildTooltip(item, tooltip) {
 		let tooltipCode = '';
+		console.log(tooltip);
```
zachstence, line 311:
```suggestion
```

zachstence (changes requested):

zachstence:
@ItsMeBrianD Do we care about this Vercel deployment? Not sure what it is or why its failing

hughess:
@kwongz the area map click target is different than I would have expected - it seems like you need to click the specific line point rather than being able to click the area itself.

I know there's some echarts limitations with that, but is there a way to enable the clicks for the area?

kwongz:
Sorry @hughess I think i'm misunderstanding, but on my end i can double click the areas in Area Map to navigate to the link.
I'm unclear on line point vs area? 

https://www.loom.com/share/f19742b9003b4a8eb3f2069cd6ad55c8

hughess:
Sorry @kwongz i meant to say area _chart_!

area map is working for me

kwongz:
@hughess, I understand now, yeah that would be a nice fix. I'll make a new issue for this since Area Charts was in another PR thats been merged in the the feature

Thanks!

hughess:
@kwongz makes sense.

The same issue exists on the line chart, where you have to click the specific point rather than being able to click the line itself. For some line charts with a small # of points, this will cause issues for users who won't be over a point when they try to click.

Ideal would be that the click target for the line chart is the full line, and for the area it's the full area

PR #2226, opened 2024-07-17
Dropdown can have option with null `value`/`label`

### LINKED ISSUE #2179: [Bug]: Dropdown containing `null` will not load
### Describe the bug

If the query feeding a dropdown component contains a null value, the input will not be set, even if a default value is supplied.

### Steps to Reproduce

Reproducible example below. If `null` is included in the `selection` query, the DataTable will not load. If you change that to "def", it will load.

````
```sql selection
select 'abc' as option
union all
select null as option
union all
select 'ghi' as option
```

<Dropdown data={selection} name=withnulls value=option defaultValue=ghi/>

```sql filtered
select * from (
select 'abc' as option, 100 as sales
union all
select 'def' as option, 200 as sales
union all
select 'ghi' as option, 300 as sales
)
where option = '${inputs.withnulls.value}'
```

<DataTable data={filtered}/>
````

### Logs

_No response_

### System Info

_No response_

### Severity

annoyance

Workaround by filtering nulls out of the query feeding the dropdown

### PR DESCRIPTION
**Pull request by zachstence**
### Description

Fixes #2179 

The root cause was this error:

image

I fixed that specific error by adding handling for `null`/`undefined` `value`/`label`.

I also enabled `ts-check` for `dropdownOptionStore.js` to catch any other related issues. Tweaked some types to make everything happy.

### Checklist

~~- [ ] I have added to the docs where applicable~~
~~- [ ] I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable~~

### DIFF
--- a/.changeset/bright-toes-march.md
+++ b/.changeset/bright-toes-march.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Dropdown can have option with null value/label
--- a/.changeset/clean-frogs-flash.md
+++ b/.changeset/clean-frogs-flash.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/sdk': patch
+---
+
+Export sharedPromise from utils/index.d.ts to get type when importing
--- a/packages/lib/sdk/src/lib/sharedPromise.js
+++ b/packages/lib/sdk/src/lib/sharedPromise.js
@@ -10,7 +10,7 @@
  **/
 
 /**
- * @template [ResolveType=unknown]
+ * @template [ResolveType=void]
  * @param {() => void} [stateChangeHook]
  * @returns {SharedPromise<ResolveType>}
  */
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Review by hughess — COMMENTED** [review]
This looks like it's working well - dropdown works for me with null included!

There is only one case I'm not sure about: filtering for legitimate nulls.

For example, if you had a table with counts by `operating_system`, but in some cases you don't have the `operating_system` so those records show `null`.

If you wanted to set up a filtered query you would do:
```sql
where operating_system = '${inputs.mydropdown.value}'
```

Which would work for any OS that is filled in, but would not work if that resolved to `null` like this:
```sql
where operating_system = 'null'
```

You could:
- Add logic to your SQL query to handle the null vs. string situation (feels complicated)
- Cast your nulls to "null" in your SQL (probably better, but still unexpected)

I don't know that there's a good solution to this. If not, we should add a note in the docs to explain our recommended practice for handling nulls in a dropdown/filtered query

ItsMeBrianD (approved):

zachstence:
@hughess I talked with @ItsMeBrianD and we agreed that the concern about filtering by `null` or `"null"` is a larger question we should consider with regard to all of our inputs.

For now, I think users can either handle `"null"`<->`null` in their query, or make use of the `rawValues` field on the input
image

@hughess Would you mind creating a new issue to track this future work?

PR #2231, opened 2024-07-17
Chart drilldowns heatmaps funnelchart

kwongz:
### Checklist

- ~~For UI or styling changes, I have added a screenshot or gif showing before & after~~


hughess:
@kwongz these are looking good.

My only concern is the discoverability of the feature for a viewer - it’s tricky to indicate that double click is required without having text to explain it on a chart, and that could become a burden for the developer, or could lead to viewers needing training.

I think we should do one more review of this behaviour in combination with the chart input + emphasis work so we have a smooth experience for the permutations people will get into. E.g.,:
- Link only - double click to navigate
- Input only - single click to set emphasis & input value
- Link and input combo - single click to set input, double click to navigate

Some ways we could increase the discoverability:
- Pointer cursor - show non-pointer cursor by default on charts, change to pointer when element is clickable
   - This would help the single click case, but only partially help the double click case
- Custom cursors - have custom cursor styles to indicate what's possible when hovering over a chart element 
- Tooltip on click - e.g,. click a bar and see something like:
   - "Filter for 'United Kingdom'"
   - "Filter for 1998"
   - "Go to drill page for 'United Kingdom'"
-  General chart info tooltip - e.g., an `i` icon at the top right of a chart, when hovered includes info that you can click or double click to achieve something
   - If we had this, we could auto-generate the content based on the props supplied but allow a full override (e.g., if `link` supplied, include "Double click to see more detail" in the tooltip by default)
   - Would also need a way to turn this off completely, and decide whether it should be on or off by default
- Some other visual indicator that a chart includes link drilldown - e.g., we have icons for link drill + input, which show up somewhere on the chart to clue users into the functions available to them on that particular chart
   - "If you see the <something> icon, it means you can double click on the chart to go to a drilldown page" 

We don't need to go for a full long-term solution on this just yet, but I want to make sure we're doing whatever easy things we can do to make this more obvious for people before we release it

kwongz:
Thanks @hughess 

Spoke to Brian about ensuring the single click (chart inputs) and double click (drilldowns) will be tested in the final feature branch.

We also plan to implement your suggestions for discoverability in another PR focused on discoverability.

If everything else looks good, can we merge these chart drills downs into feature/chart-inputs-emphasis?

Thanks

hughess (approved):

PR #2239, opened 2024-07-19
Main -> Next (after Release 2024-07-19)

zachstence:


ItsMeBrianD (approved):

zachstence:
There is a type error in `sdk` causing the `prepack` script to fail, and the NPM publish to fail. @ItsMeBrianD is fixing.

PR #2256, opened 2024-07-22
Clarified `tooltipType=hover` for Area, Bubble, Point Map Docs

kwongz:
### Checklist

~~[ ] For UI or styling changes, I have added a screenshot or gif showing before & after~~
~~[ ] I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable~~

Before:
image

After:
image

Considering the tooltipType=hover, it would be beneficial to add general text indicating that the point can be clicked for input store updates now that `contentType: 'link', linkLabel: 'Click here'` has been removed on hover. This might be addressed when we implement our legend for single-click and double-click actions for chart inputs and drilldowns.

Additionally consideration: Add another contentType for more use cases instead of just a link.

image


.changeset/twenty-meals-fetch.md
zachstence, line 1: Since nothing was changed in `core-components`, we don't need a changeset. Can you delete this file?

zachstence (changes requested):

.changeset/twenty-meals-fetch.md
kwongz: Will do

zachstence (approved):

PR #2283, opened 2024-07-29
Feature/2193 dimension grid multi select

kwongz:
### Checklist


mcrascal (changes requested):
Love it 

1. This introduces a regression in initial loading where the grid passes through an error state. Can be seen in the docs preview pretty reliably. 
2. Can we add a story for selecting a null value from a column when the grid functions as a multi-select. (see the `Null Dimension Values` story), and confirm that this case also generates correct SQL? 

CleanShot 2024-07-30 at 11 30 52

kwongz:
Looking into this now

kwongz:
@ItsMeBrianD The Playwright tests are all failing on my local repo. Do you know if these need to pass before merging?

mcrascal (changes requested):
There is a regression the interaction, which I think was introduced since my last review.

See the jank / fliker on interaction in the GIF below. I suspect this is related to the animation on the height, but I am not sure. 

CleanShot 2024-08-01 at 12 19 10

kwongz:
I think i know what this is, taking a look now

kwongz:
https://www.loom.com/share/07778b48ea5b4b1ca98b3ed3c1ecdb91?sid=627b8da9-127b-4311-b684-abdfb461a789

Issue: re-rendering of entire dimension grid occurs on row click
Fix: conditionally render dimension grid on initial load only,

mcrascal (changes requested):
Nice, the interaction feels really smooth now. 

For de-selecting values in the multi-select case, I think there is some additional logic required to prevent users from arriving in an irrecoverable state: 

CleanShot 2024-08-01 at 14 24 27

mcrascal:
@kwongz I think the behaviour you want here is to always show the options which have been selected in a given column, even if they are now zeros due to selections / de-selections in other columns. In the example above, sinister toys and mysterious apparel have both been selected, and should always be visible in the category column despite the mystic pendant item only being in the mysterious apparel category.

kwongz:
@mcrascal 

https://www.loom.com/share/d8c5e7a165da492abec9bff12a5086be

- When Selected Rows do no exist in query return, appear as de-selectable red row with missing values
- (End of Video) I'm noticing some lag on multi row selection, curious to know if you experience it as well

kwongz:
@mcrascal 
**Removed**:
-  Red background color

**Keep:**
-  100% color fill for row, 
- if (row value === NaN) = 'Missing'

image

kwongz:
Grey bg

image

kwongz:
**Future Suggestion:**

- It would be easier for users if the html example code container's height fit its content. Currently, containers seem to keep a fixed height, possibly for a cleaner and uniform look across pages. Curious about your thoughts?

image

archiewood:
I 100% agree and it used to be like that - I think this is a regression, possibly introduced when we added code highlighting to non-SQL code blocks

mcrascal:
@kwongz this is excellent. It feels really good to use. 

Two final items: 

1. Can we get a story that documents the new `missing` state on the multi-select - you may need to generate some new data for this one. Grab me if you need a hand. 
2. Can we move the dimension grid _above_ the linecharts in the docs. It'll help reduce the feeling of jank when you hit an empty state.

zachstence (approved):

kwongz:
Docs: Updated component format (Looks much cleaner!)
image

image

Stories: Added Column Combo w/ Null row values
image

kwongz:
Need to look back into height handling in multi select, discovered when columns break into multiple rows, height handling breaks

image

mcrascal:
@kwongz some tweaks to the list animation in the above

kwongz:
https://www.loom.com/share/d3e3fd66f72d4ba8801f64460c768f1a?sid=d1f75de7-3bb2-4372-b8ab-d17284f82fa9

Fixed: selected rows with null values not appearing in columns
Restructured: Missing Values logic

kwongz:
fix #2193

mcrascal (approved):
LGTM tremendous effort

PR #2301, opened 2024-08-06
Rewrite Dropdown

ItsMeBrianD:
### Description
- Dropdown has all logic moved into dropdown option store
- Dropdown option store is rewritten to simplify behavior
- Dropdown option store has 100% test coverage, with sane tests
- DropdownOption now respects the dom order -

- Lodash packages removed, use main lodash
- Debounces all moved to perfect-debounce
- Vitest is upgrade to latest
- Query search threshold changed to 1 - (1/n) where n is search term length

### Checklist

- [ ] ~~For UI or styling changes, I have added a screenshot or gif showing before & after~~ No display logic was modified
- [ ] ~~I have added to the docs where applicable~~ Author interface remains untouched


mcrascal (changes requested):
See: docs

packages/ui/core-components/src/lib/atoms/inputs/dropdown/dropdownOptionStore.js
```diff
@@ -12,7 +12,7 @@ import { z } from 'zod';
 
 const DropdownOptionSchema = z.object({
 	label: z.union([z.string(), z.number()]).optional().nullable(),
-	value: z.union([z.string(), z.number()]).optional().nullable(),
+	value: z.union([z.string(), z.number()]),
```
zachstence, line 15: Why is this change necessary?

I think at the very least we want `value` to be `.nullable()` so that a `null` value returned from a query can be used

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -164,16 +164,17 @@
...
 	$: open ? pauseSorting() : resumeSorting();
+	const toKey = (queryOpt) => queryOpt[value]?.toString() + queryOpt[label]?.toString();
 </script>
 
 <slot />
 
 <QueryLoad data={finalQuery} let:loaded>
 	<div slot="skeleton"></div>
-	{#each loaded ?? [] as queryOpt (queryOpt.value?.toString() + queryOpt.label?.toString())}
+	{#each loaded ?? [] as queryOpt (isNaN(toKey(queryOpt)) ? Math.random().toString() : toKey(queryOpt))}
```
zachstence, line 174: `isNaN` here will always be `false` since `toKey` returns a string

zachstence (changes requested):

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
ItsMeBrianD: It actually won't if value and label are undefined (which is the case during HMR for some reason)

packages/ui/core-components/src/lib/atoms/inputs/dropdown/dropdownOptionStore.js
zachstence: [bf2c3c6](https://github.com/evidence-dev/evidence/pull/2301/commits/bf2c3c618dd3ca460b1d6a7a3ef1800a9fcf0bf8)

zachstence (approved):

ItsMeBrianD:
Looking at Chromatic with @archiewood 

Issues:
- Sorting seems to have changed in an unexpected way (See: `open`)
- Dropdown's no longer have an unset state

Notes:
- DOM Sort order is enforced now

mcrascal (changes requested):
[Using Dropdowns that interact with eachother's queries](https://www.chromatic.com/component?appId=645d58d2e049d64800e0ee15&csfId=atoms-inputs-dropdown&specName=Using%20Dropdowns%20that%20interact%20with%20eachother%27s%20queries&buildNumber=745) seems a little off

mcrascal (changes requested):
Sort orders

packages/lib/component-utilities/src/buildQuery.js
```diff
@@ -46,6 +46,7 @@ export const buildReactiveInputQuery = (queryProps, id, initialData) => {
 		results: derived(internal, (v) => v),
 		update: async (queryProps) => {
 			const { hasQuery, query } = buildInputQuery(queryProps, id);
+			console.log(queryProps);
```
archiewood, line 49: missed console log

mcrascal (changes requested):
Dropdowns that interact story -- we expect default value to be set for first dropdown?

PR #2302, opened 2024-08-06
Add skeleton loading state to Area Map, Bubble Map, and Point Map

zachstence:
### Description

https://github.com/user-attachments/assets/f31e8b48-8d33-4ddf-9807-e57b02126960

### Checklist

~~- I have added to the docs where applicable~~

### DIFF
--- a/.changeset/brave-toys-press.md
+++ b/.changeset/brave-toys-press.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Add skeleton loading state for Area Map, Bubble Map, and Point Map
--- a/.prettierignore
+++ b/.prettierignore
@@ -26,6 +26,7 @@ yarn.lock
 **/sites/docs/pages/**/*
 **/sites/test-env/pages/**/*
 **/static/data/*
+**/static/geo-json/*
 
 **/node_modules
 **/**/node_modules
--- a/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
+++ b/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
@@ -19,6 +19,8 @@
 
 	let _data;
 	onDestroy(unsub);
+
+	$: console.log({ dataLoaded: _data?.dataLoaded, error: _data?.error });
 </script>
 
 {#if !data}
--- a/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
+++ b/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
@@ -19,8 +19,6 @@
 
 	let _data;
 	onDestroy(unsub);
-
-	$: console.log({ dataLoaded: _data?.dataLoaded, error: _data?.error });
 </script>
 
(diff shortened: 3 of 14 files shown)


zachstence:
Big improvement to the loading experience on the docs:

Before:


https://github.com/user-attachments/assets/df848e26-1b9f-406b-b1f5-119df3ae5c74



After:



https://github.com/user-attachments/assets/a3407683-87ce-4ee4-9a8a-c37c1d22025b

zachstence:
Seems like there still is a period of time where the map renders but has no points/areas/bubbles on it. I think this happens when the SQL query has finished but we are downloading GeoJSON data. I'm not sure if/how we can continue to show the skeleton while we are loading GeoJSON. @ItsMeBrianD any thoughts here?

hughess:
@zachstence this is a huge improvement!

Only other thing I think we should do is adjust the height to match the height of the rendered map. I think there are 2 pieces to that:

### 1. Adjust skeleton height when `height` prop is used

Here I've set `height=550`, but the skeleton stays the same height:

CleanShot 2024-08-07 at 06 29 00

### 2. Adjust skeleton height when `title` prop is used

When the user uses the `title` prop since it will add vertical space to the component. Here's an example - we get a bit of layout shift on load:

CleanShot 2024-08-07 at 06 26 37

For this one I think we have 2 options:
1. Add height to the skeleton if `title` is not `undefined`
2. Remove the title from the skeleton, so the title would immediately render, with the map loading underneath it
   - This would be inconsistent with the charts' behaviour though

hughess:
There also appears to have been a regression in zoom behaviour - the original "initial zoom" level was set to be quite high to account for cases where the data takes a while to load. It's a bit hacky but it does help:

### Without hacky zoom fix
CleanShot 2024-08-07 at 06 48 46

### With hacky zoom fix
CleanShot 2024-08-07 at 06 48 27

hughess:
Just pushed a small commit to change the starting zoom from 1 to 5, which gets us the benefit of the hacky zoom fix in cases where the user hasn't supplied starting coordinates

zachstence:
I added a skeleton that shows while GeoJSON is loading. See the updated demo videos below to compare before/after

There is a slight blip when we swap between skeletons for loading SQL query and loading GeoJSON, but this is more difficult to fix and would require a larger refactor to our maps components. I think this is fine to leave as is for now.

_Demo videos have a 2 second delay added while loading GeoJSON to simulate network latency / large GeoJSON files_

`AreaMap`

Before:


https://github.com/user-attachments/assets/0d587a0f-fe79-474f-912f-7acdf46ffdf4



After:


https://github.com/user-attachments/assets/53f8ab57-ebae-4c0a-80bc-22acf94e422d



`BaseMap` with `Areas`

Before:


https://github.com/user-attachments/assets/157105be-9b6c-4e65-b86b-275fe148bb98



After:


https://github.com/user-attachments/assets/d96c7f38-cc2c-447c-b858-651a506b10e5

hughess:
That looks great! Agreed on the blip, seems okay - worth it to have the loading state carry over to the geojson

hughess (approved):

PR #2303, opened 2024-08-07
Add smooth zoom to leaflet

hughess:
## Smooth Zoom
Adds a smooth zoom plugin to leaflet, adapted from https://github.com/mutsuyuki/Leaflet.SmoothWheelZoom

## Before & After
This isn't a perfect comparison, but the smooth zoom is a significantly faster experience

### Before
`zoomSnap: 0.25`
CleanShot 2024-08-07 at 08 02 25

### After
`scrollWheelZoom: false`
`smoothWheelZoom: true`
`smoothSensitivity: 5`
CleanShot 2024-08-07 at 08 03 06

## Considerations
- Will this work on all OS/browser combos?
   - Tested MacOS + Windows Chrome, Safari, Firefox and all worked pretty well
- How well does this work with high volumes of data/shapes? Will this slow down/get choppy if a page contains a lot of high intensity viz components (e.g., a very long horizontal bar chart)?  
   - Appears okay from brief testing, but something to keep an eye on 
- There seems to be a transition issue from the initial coords/zoom to the loaded data at specific zoom settings
   - In the LA area map example, when default starting zoom is set to 5, you see a very zoomed in "LOS ANGELES" label before the data loads. When zoom is 6 you don't see that. To be investigated. 

### Checklist


hughess:
@zachstence one thing I noticed on this branch after merging in the loading changes - when there's an error with the geojson load, it seems to stay in an infinite loading state rather than display the error:

CleanShot 2024-08-08 at 11 03 37@2x

zachstence:
> @zachstence one thing I noticed on this branch after merging in the loading changes - when there's an error with the geojson load, it seems to stay in an infinite loading state rather than display the error:
> 
> CleanShot 2024-08-08 at 11 03 37@2x

Taking a look

zachstence:
@hughess The error is now being properly surfaced to the user

image

hughess:
@zachstence amazing, that was fast!

zachstence (approved):

Issue #2144, opened 2024-06-21
[Bug]: App Scrolls to Top of Page Instead of Staying Static When Fullscreen View Is Clicked


mdp1125:
### Describe the bug

When clicking 'show fullscreen' on a data table, it automatically scrolls up to the top of the page behind the popup modal. The desired behavior should be to keep the page static behind it so the user does not have to scroll back to where they were.

### Steps to Reproduce

Click fullscreen on a data table :) 

### Severity

annoyance

### Additional Information, or Workarounds

Minor inconvenience but would help keep the user in flow when using the app. Important for external facing users like prospects who are deciding whether to buy our product

ItsMeBrianD:
@kwongz proposed solution is:

- Rewrite fullscreen from the ground up, maintain the same prop interface for backwards compatibility.
- I would look into using `Portal.svelte` and/or `Modal.svelte` to achieve the desired behavior.
- You'll probably want to use 2 elements (1 that is fullscreen, flex center an inner element that contains the slot), look at the props used by the current Fullscreen component for the desired size of the content div

mdp1125:
hey @kwongz @ItsMeBrianD - wanted to post this again on the issue for visibility https://github.com/evidence-dev/evidence/pull/2308#issuecomment-2293487950

^ the latest change is great in that it solves the scroll to top problem but in my instance i am seeing that the page navigation arrows are cut off when in fullscreen mode which is difficult for our users, they open the full result set and then page through it in full screen mode. fyi I am using Chrome browser.

Screenshot from opening a 90page result set in fullscreen mode, you can see the page arrows are cut off
![image](https://github.com/user-attachments/assets/efc173c9-d4cc-4bcb-84ff-74db33a883b3)

archiewood:
Hey @mdp1125 can you open a new issue for this?

archiewood:
![CleanShot 2024-08-16 at 09 22 50](https://github.com/user-attachments/assets/52560738-1f74-4fd1-be4c-847a62d8d6fb)

mdp1125:
> Hey @mdp1125 can you open a new issue for this?
> 
> Can you also include details of your browser, os, device etc, as I can't immediately replicate this

opened a new issue here: https://github.com/evidence-dev/evidence/issues/2447

hughess:
Closing this issue along with our release of the new Evidence Core: https://docs.evidence.dev/

Feel free to open a new issue for this if it is blocking you!

PR #2308, opened 2024-08-07
removed scroll to top and fixed esc close for fullscreen mode

kwongz:
### Checklist

1. Removed Scroll to top
2. Removed Scroll when modal is open
3. Fixed esc key behaviour that locked user out from fullscreen mode

https://www.loom.com/share/127253117d0840f3af0650cae7ff0699

### DIFF
--- a/.changeset/little-plums-repeat.md
+++ b/.changeset/little-plums-repeat.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+removed scroll to top and fixed esc close for fullscreen mode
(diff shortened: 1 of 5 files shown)


mcrascal (approved):

kwongz:
vertical scroll removed
before:
image

after:
image

focus appears on dialog modal open in firefox as well
image

kwongz:
fix #2144

mdp1125:
hey @kwongz thanks for pushing these changes. However, I noticed that now when clicking fullscreen on result sets where the full screen should have multiple pages to click through, the < [page number] > section is cut off. Was that intentional? It is a really necessary use case for our users, they click open fullscreen and then will page through the results in the fullscreen mode

mdp1125:
> hey @kwongz thanks for pushing these changes. However, I noticed that now when clicking fullscreen on result sets where the full screen should have multiple pages to click through, the < [page number] > section is cut off. Was that intentional? It is a really necessary use case for our users, they click open fullscreen and then will page through the results in the fullscreen mode

See new issue here: https://github.com/evidence-dev/evidence/issues/2447

PR #2317, opened 2024-08-09
Clean up stories to fully enable chromatic

zachstence:
### Description

- Use `export const meta` instead of deprecated `<Meta />`
- Use `date.between` instead of `date.recent` to prevent faker data from shifting over time
- Use freshly seeded faker instance for each faker table
- Get rid of `fakerSeries`, write out queries in stories instead
- Use Inter font in ECharts with proper `font-feature-settings` (open 4s and tabular numbers)
- Add Chromatic storybook addon
- Fix `BoxPlot` rendering in storybook
- Fix `CalendarHeatmap` rendering in storybook
- Split `Map.stories.svelte` into separate files for `BaseMap`, `AreaMap`, `BubbleMap`, `PointMap`
- `ReferenceArea`, `ReferenceLine` and `ReferencePoint` stories no longer depend on `numeric_series` series names

Inter font with open 4s and tabular numbers
image

### Checklist

- [x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)

~~- For UI or styling changes, I have added a screenshot or gif showing before & after~~
~~- I have added to the docs where applicable~~
~~- I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable~~

PR #2436, opened 2024-08-15
Clean up stories to fully enable chromatic (2/2)

zachstence:
### Description

Continuation of the work here https://github.com/evidence-dev/evidence/pull/2317


zachstence:
Was able to see the problem with `CalendarHeatmap` locally. Two runs of `pnpm sources` in `core-components` 3 minutes apart (3:58pm -> 4:01pm) results in _very slightly_ different `departure_date` timestamps, with everything else the same

image

archiewood:
Awesome, so should we turn chromatic on as a required test?

zachstence:
> Awesome, so should we turn chromatic on as a required test?

Not quite yet, still need to clean up the differences that are happening

zachstence:
Just fixed a painful mistake with the faker yaml options (should fix issue with dates not generating the same every time) and batch accepted all changes.

zachstence:
Seems that the label background for references isn't sized correctly sometimes. Not sure why

image

ItsMeBrianD (approved):

Issue #2447, opened 2024-08-16
[Bug]: Fullscreen cuts off page nav in deployed environment


mdp1125:
### Describe the bug

Seems to be related to the resolution of issue https://github.com/evidence-dev/evidence/issues/2144 from this PR https://github.com/evidence-dev/evidence/pull/2308 in v39.1.4.

The PR and issue resolution works great in local (results dynamically size and page nav bar remains set at bottom).

However, in deployed environment (app.supplyscope.ai), it does not function the same way.

When opening the fullscreen, the results can dynamically size as the window changes but when the results exceed some set number, the page nav bar is cut off, which disrupts our user workflow.

My browser user agent is this using Google Chrome
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36

Im on a Macbook Pro 14 Inc Apple M3 Pro chip on MacOS Sonoma 14.1. I have provided the system info from my local below, however it seems to work great when in the local dev mode.

I can't show much on this issue as the data is proprietary but this is what the bottom of the fullscreen mode looks like when there's over 800 results. Nav page is nowhere to be seen
![image](https://github.com/user-attachments/assets/ab03d11a-1f0f-4e9f-8c7e-b0b2da56dc51)

### Steps to Reproduce

Log in to our deployed environment and see if you have the same issue?

### System Info

```Shell
System:
    OS: macOS 14.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 718.98 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.0 - /usr/local/bin/node
    npm: 10.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 127.0.6533.120
    Safari: 17.1
  npmPackages:
    @evidence-dev/bigquery: ^2.0.4 => 2.0.5 
    @evidence-dev/core-components: ^4.7.5 => 4.7.5 
    @evidence-dev/csv: ^1.0.7 => 1.0.9 
    @evidence-dev/databricks: ^1.0.3 => 1.0.5 
    @evidence-dev/duckdb: ^1.0.7 => 1.0.9 
    @evidence-dev/evidence: ^39.1.4 => 39.1.4 
    @evidence-dev/mssql: ^1.0.5 => 1.0.7 
[...]

PR #2448, opened 2024-08-16
Fix/2447 dynamic datatable fullscreen height

kwongz:
### Checklist

- ~~For UI or styling changes, I have added a screenshot or gif showing before & after~~

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
@@ -435,11 +435,13 @@
 
 {#if !isFullPage && innerHeight !== undefined}
 	<Fullscreen bind:open={fullscreen} {search}>
-		<!-- header and last row are 22.5+22.5 = 45px, middle rows are 23, when compact middle rows are 17.5 -->
+		<!-- when compact middle rows are 17.5, middle rows are 23 -->
 		{@const ROW_HEIGHT = compact ? 17.5 : 23}
+		<!-- header and last row are 22.5+22.5 = 45px -->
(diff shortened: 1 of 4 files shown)


kwongz:
fix #2447

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -437,7 +437,8 @@
 	<Fullscreen bind:open={fullscreen} {search}>
 		<!-- header and last row are 22.5+22.5 = 45px, middle rows are 23 -->
 		{@const ROW_HEIGHT = 23}
```
archiewood, line 439: will this work if someone uses `<DataTable compact=true/>`

kwongz: No issues in Chrome and Firefox, User is experiencing issue in Chrome

Chrome:
image

image

kwongz:
@archiewood Have not been able to reproduce to the degree the user was experiencing, but i've left a note in slack to get their markdown code of the `DataTable` and asked about any custom CSS styling they may have used.

Currently, on our Docs site, when the search bar is present, we do experience a bit of overflow (similar, but not to the debilitating degree our user has flagged)

(when search bar is present, slight overflow is preset indicated by y scroll bar)
image

This PR aims to take the search bar's additional height into its total `DataTable`'s height, which will remove the slight overflow we are experiencing.

After speaking with the user, returning the overflow-y property to the Fullscreen DataTable will enable the user to access the fullscreen datable pagination. The User confirmed that this was what they desired to return.

kwongz:
Added Comments/Variables to clarify calculations in height calc

mcrascal (approved):

PR #2453, opened 2024-08-19
Fix/2323 tab styling button groups

### LINKED ISSUE #2323: [Feature]: Tab Styling for Button Group
### Description

Sometimes you want the functionality of a ButtonGroup (ie an input) but with the styling of a Tabs component from a UI perspective

### What problem would this solve?

_No response_

### How should it work?

```svelte
<ButtonGroup 
   appearance=tabs
   ...
>

### PR DESCRIPTION
**Pull request by kwongz**
### Checklist

image

@archiewood Should `<ButtonGroup {display="tabs">` have a inherit default value, similar to the tabs component?

Tabs Component (default value):

image

ButtonGroup styled as Tabs (No Default Value): 

image

fix #2323

### DIFF
--- a/packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
@@ -6,7 +6,7 @@
 	import { presets, setButtonGroupContext } from './lib.js';
 	import { writable, readonly } from 'svelte/store';
 	import { INPUTS_CONTEXT_KEY } from '@evidence-dev/component-utilities/globalContexts';
-	import { getContext } from 'svelte';
+	import { getContext, setContext } from 'svelte';
 	import { buildReactiveInputQuery } from '@evidence-dev/component-utilities/buildQuery';
 	import ButtonGroupItem from './ButtonGroupItem.svelte';
 	import { page } from '$app/stores';
@@ -21,6 +21,13 @@
 
 	/** @type {keyof typeof presets | undefined} */
 	export let preset = undefined;
+
+	// for Tabs styling
+	export let display;
+
+	setContext('button-display', display);
+
+	export let color = 'hsla(207, 65%, 39%, 1)';
 
 	const inputs = getContext(INPUTS_CONTEXT_KEY);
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte [hunk 1]
```diff
@@ -49,29 +56,36 @@
 </script>
 
 <HiddenInPrint enabled={hideDuringPrint}>
-	<div class="inline-flex w-fit max-w-full flex-col mt-2 mb-4 ml-0 mr-2">
+	<div
+		class={display === 'tabs' ? '' : 'inline-flex w-fit max-w-full flex-col mt-2 mb-4 ml-0 mr-2'}
+	>
```
ItsMeBrianD, line 61: Having all of these ternaries that depend on display here is a smell to me, do we need to have other display elements or similar?

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroupItem.svelte
```diff
@@ -20,16 +25,26 @@
...
+		id={value}
+		label={valueLabel}
+		{color}
+		on:click={() => update({ valueLabel, value })}
+		activeId={$currentValue?.value}
+	/>
+{:else}
+	<button
+		type="button"
+		class=" flex-none py-1 font-medium h-8 px-3 text-xs truncate
```
archiewood, line 39: is there a reason we can't do the tab styling with just classes?

rather than a new component for the tab display

feels like it would be easier to maintain this way

archiewood (changes requested):
Regarding your top comment, I think there needs to be a way to choose the selected tab (/button) when the page loads. eg 

```svelte
<ButtonGroup
  data={categories}
  display=tabs
  defaultValue="Sinister Toys"
/>
```
This is a sensible prop for ButtonGroup even if not displayed as tabs


I'm unclear on whether this should automatically be the first value or not
- That would be consistent with Tabs
- But inconsistent with buttongroup current behviour

My inclination would be to stick with buttonGroup behaviour, so no default unless specified

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroupItem.svelte
kwongz: By having a display tabs component we can maintain consistency between tabs and buttons styled as tabs in one component. Without that, the two styles could diverge over time.

kwongz:
@archiewood I think this is something we can do in the future, I'll add a issue for this to be resolved in another PR.

kwongz:
Default Value Issue:
https://github.com/evidence-dev/evidence/discussions/2468

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroupItem.svelte
archiewood: I am more concerned about divergence of _functionality_ between buttonGroup styled as tabs and ButtonGroup styled as buttons

zachstence (approved):

archiewood:
This appears to have broken tabs across evidence

See [docs.evidence.dev](https://docs.evidence.dev/components/tabs/)

CleanShot 2024-08-30 at 11 05 13

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
kwongz: Sorry i don't follow, but i agree that this may be a bit smelly

PR #2481, opened 2024-08-30
Fix/display tabs breaks tabs

kwongz:
### Checklist


kwongz:
Hi @archiewood, This PR reverts Tab Display as buttons back into the repo with the additional fix of Tabs onClick functionality.

I believe separating the Tab's display logic into another component is the best approach (despite our issue this morning). This ensures if there are any future changes to Tabs styling, all changes can occur in the Tabs Display Component.

The Tabs onclick bug we experienced today was a lack of foresight from me, and I think that your storybook test will help us avoid this issue in the future. 

Thanks again!

preview docs:

https://github.com/user-attachments/assets/42ac7ce8-e8e6-4c27-9a6a-88c9b5fd6a06

archiewood:
hey @kwongz can you resolve this merge conflict?

kwongz:
merge resolved @archiewood, Thanks!

sites/docs/pages/components/button-group.md
```diff
@@ -223,6 +267,11 @@ group by all
     description="SQL where fragment to filter options by (e.g., where sales > 40000)"
     options="SQL where clause"
 />
+<PropListing 
+    name="display"
+    description="Displays tabs with button functionality"
+    options="tabs"
```
archiewood, line 273: `defaultValue="buttons"`?

I know you have not actually set a default value but it's confusing to only have one value, and it not be set

(otherwise we should have done something like `isTabs=true`, default false)

archiewood (approved):
Looks good. Minor docs nits

Also maybe we can make the docs propListing option clickable by using an array (check some of the other components)

Feel free to merge when ready

sites/docs/pages/components/button-group.md
kwongz: In https://github.com/evidence-dev/evidence/pull/2483, we introduce `defaultValue="buttons"` in button groups. But I Agree, we need to add it to the docs. I will add it to the docs for 2483 since it is more relevant.

will also add the clickable option!

Thanks

Issue #2315, opened 2024-08-09
[Bug]: During build: Failed to pre-render columns Parser Error: syntax error at or near "SELECT" DESCRIBE SELECT


archiewood:
### Describe the bug

During build process, Evidence spits out error messages to do with the pre-rendering of columns

### Steps to Reproduce

1. Clean install on  "@evidence-dev/evidence": "^39.1.3",
2. `npm run build`
3. Observe build logs

### Logs

```Shell
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 22.66s
Failed to pre-render columns Parser Error: syntax error at or near "SELECT"
LINE 2: DESCRIBE SELECT * FROM (select
                 ^
Failed to pre-render columns Parser Error: syntax error at or near "SELECT"
LINE 2: DESCRIBE SELECT * FROM (SELECT DISTINCT category AS "value", category AS "label" FROM (SELECT * FROM (select
      category
  from needful_things.orders
  group by category
) AS "inputQuery-jV")
) AS "inputQuery-5v"
...
                 ^
Failed to pre-render columns Parser Error: syntax error at or near "SELECT"
LINE 2: DESCRIBE SELECT * FROM (select
                 ^
Failed to pre-render columns Parser Error: syntax error at or near "SELECT"
LINE 2: DESCRIBE SELECT * FROM (SELECT DISTINCT category AS "value", category AS "label" FROM (SELECT * FROM (select
      category
  from needful_things.orders
  group by category
) AS "inputQuery-Yw")
) AS "inputQuery-Up"
...
```

### System Info

```Shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 84.97 MB / 16.00 GB
    Shell: 5.1.16 - /opt/homebrew/bin/bash
  Binaries:
    Node: 20.15.0 - ~/.nvm/versions/node/v20.15.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.15.0/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v20.15.0/bin/npm
    pnpm: 8.6.9 - /opt/homebrew/bin/pnpm
[...]

PR #2482, opened 2024-08-30
Clean up column pre-rendering

zachstence:
Resolves https://github.com/evidence-dev/evidence/issues/2315

### Description

Our [previous implementation](https://github.com/evidence-dev/evidence/pull/2261) of column pre-rendering worked by adding a new query suffixed with `__DESCRIBE` that would run during SSR to generate column metadata for all queries.

This inadvertently caused us to run unnecessary queries to describe the `Query` `#fetchColumns` query, resulting in a nested describe query (e.g. `describe describe select ...`). `duckdb-wasm` currently has a bug where nested describes cause an error, which resulted in very noisy logs during build with no actual issues.

This PR removes the `__DESCRIBE` query and instead uses the `_columns` query that is already run by `Query`.

I've also added a debug log to indicate when pre-rendered `knownColumns` are being used, similar to the logs we have in place for `initialData` and result length.

### Demos

Columns information is pre-rendered and no queries run after the first page load

https://github.com/user-attachments/assets/2f223b36-ad4e-46ae-95f4-f1bed7994a7d

Debug log for `knownColumns`

known columns debug log

### Checklist

~~- I have added to the docs where applicable~~


sites/example-project/src/pages/+layout.js
```diff
@@ -131,19 +131,6 @@ export const load = async ({ fetch, route, params, url }) => {
...
-			usqlQuery(`DESCRIBE ${sql}`, {
-				route_hash: routeHash,
-				additional_hash: paramsHash,
-				query_name: `${query_name}__DESCRIBE`,
-				prerendering: building
-			});
-		} catch (e) {
-			console.debug('Failed to pre-render columns', e instanceof Error ? e.message : e);
-		}
-
```
ItsMeBrianD, line 146: Doesn't this disable column pre-rendering completely?

zachstence: No. Adding the `__DESCRIBE` query was actually unnecessary (as far as I can tell). Using the existing `_columns` query from `Query` is sufficient. Otherwise, we end up running a nested `DESCRIBE DESCRIBE SELECT ...` when we try to pre-render the columns for the columns query.

All that was necessary for column pre-rendering is using the proper query in `process-queries`
https://github.com/evidence-dev/evidence/blob/7ae11a804e4b93b9325b157c1d31085223928eb0/packages/lib/preprocess/src/process-queries.cjs#L74

ItsMeBrianD (approved):

PR #2487, opened 2024-09-03
Fix preprocess tsconfig

csjh:
### Description

Fixes preprocess tsconfig to include `*.cjs` and `*.js`, and fixes all type errors that popped up as a result

### Checklist

### DIFF
--- a/packages/lib/preprocess/src/extract-queries/extract-queries.cjs
+++ b/packages/lib/preprocess/src/extract-queries/extract-queries.cjs
@@ -5,15 +5,16 @@ const fs = require('fs');
 const getPrismLangs = require('../utils/get-prism-langs.cjs');
 const { parseFrontmatter } = require('../frontmatter/parse-frontmatter.cjs');
 const chalk = require('chalk');
-/** @typedef {{id: string, compileError: string, compiledQueryString: string, inputQueryString: string, compiled: boolean, inline: boolean}} Query */
+/** @typedef {{ id: string, compileError?: string, compiledQueryString: string, inputQueryString: string, compiled: boolean, inline: boolean }} Query */
 
+/** @type {Record<string, boolean>} */
 const warnedExternalQueries = {};
 
 /**
  *
  * @param {string} externalQuery
  * @param {string} id
- * @returns {Query}
+ * @returns {Query | undefined}
  */
 const readFileToQuery = (externalQuery, id) => {
 	try {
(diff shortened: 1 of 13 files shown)


csjh:
Not 100% sure about one of the changes here, `prismjs` -> `prismjs/components` (base `prismjs` doesn't have the `alias` field)

@ItsMeBrianD do you remember why [this change](https://github.com/evidence-dev/evidence/commit/04618d84e0132b29fb8da91b14906b0807b115e7#diff-22d1d9031151b0fb01e1853325afcb96542db330dce58d128e190001614ded0e) was made before?

packages/lib/preprocess/src/extract-queries/extract-queries.cjs
```diff
@@ -135,13 +138,15 @@ const strictBuild = process.env.VITE_BUILD_STRICT === 'true';
...
+ * @param {string} content
  * @returns {Query[]}
  */
 const extractQueries = (content) => {
+	/** @type {Query[]} */
 	const queries = [];
 
-	queries.push(...extractExternalQueries(content));
+	// todo: second parameter is filename but we don't have that here?
+	queries.push(...extractExternalQueries(content, ''));
```
zachstence, line 149: You can make the second argument `filename` optional in the JSDoc comment on `extractExternalQueries`
https://stackoverflow.com/a/29441515/9080819

packages/lib/preprocess/src/extract-queries/extract-queries.cjs
```diff
@@ -73,10 +78,7 @@ const extractExternalQueries = (content, filename) => {
 		return true;
 	};
 
-	/**
-	 * @type Query[]
-	 */
-	return frontmatter.queries
+	return /** @type {string[]} */ (frontmatter.queries)
```
zachstence, line 81: There's a `.map` statement below with two branches, one for `string` and `object`, so I don't think this type is correct: We don't expect everything in this array to be strings.

zachstence (changes requested):

ItsMeBrianD (dismissed):
Zach's comments

packages/lib/preprocess/src/extract-queries/extract-queries.cjs
csjh: I'm more wondering if we should be passing a file name into `extractQueries`. Not sure if we used to pass a file name in for the better error in `extractExternalQueries`

zachstence: Maybe, maybe not. But for now we should just maintain the argument being optional with the added types

zachstence (approved):

zachstence:
Needs approval from @ItsMeBrianD

ItsMeBrianD (approved):

PR #2503, opened 2024-09-05
Multiline string connector options + Snowflake Key-Pair Fix

archiewood:
### Description

image

Some connection options are by nature "multi-line" but Evidence only allowed users to enter a single line option. 

This was causing Snowflake Keypair auth to fail as the newlines in the Private Key were being erroneously stripped during entry into the input field.

This PR adds support for multi-line strings with a textarea input field and updates the snowflake private key field to use it

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/organisms/source-config/SourceConfigFormField.svelte
+++ b/packages/ui/core-components/src/lib/organisms/source-config/SourceConfigFormField.svelte
@@ -139,7 +139,7 @@
 </script>
 
 <div class="w-full">
-	<label class="flex justify-between w-full h-11 items-start">
+	<label class="flex justify-between w-full items-start {spec.type !== 'multiline' ? 'h-11' : 'h-auto'}" >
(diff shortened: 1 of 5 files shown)


packages/ui/core-components/src/lib/organisms/source-config/SourceConfigFormField.svelte
```diff
@@ -139,7 +139,9 @@
 </script>
 
 <div class="w-full">
-	<label class="flex justify-between w-full h-11 items-start">
+	<label
+		class="flex justify-between w-full items-start {spec.type !== 'multiline' ? 'h-11' : 'h-auto'}"
+	>
```
ItsMeBrianD, line 144:
```suggestion
	<label
		class="flex justify-between w-full items-start"
		class:h-11={spec.type !== "multiline"}
		class:h-auto={spec.type === "multiline"}
	>
```

ItsMeBrianD (approved):

PR #2511, opened 2024-09-06
Fix/2507 dependent dropdowns

### LINKED ISSUE #2507: [Bug]: Dropdown selections are not removed when the query changes to exclude them
### Describe the bug

Reported in [Slack](https://evidencedev.slack.com/archives/C023LRB9Z40/p1725495069231369)

Dependent Dropdown does not deselect the dropdown selections when the query data powering it removes those rows from the selection

https://github.com/user-attachments/assets/70e84260-72a9-49cf-a532-db0c5e235324

### Expected behavior
When selecting a store type, the second dropdown should deselect the stores that do not match the selected store type.

### Actual behavior
The second dropdown filters the options list correctly, but any previously selected options are not deselected.

### Steps to Reproduce

https://github.com/archiewood/tabs-bug

### Logs

_No response_

### System Info

_No response_

### Severity

_No response_

kwongz:
### Description

Fixes #2507 

Fixes dropdown selection not updating when the filtered list changes. Previously, selected items would remain even if the filtered list became empty. Dropdown now updates selections based on available data.

https://github.com/user-attachments/assets/c8fc6390-f193-4306-8f73-5ffd413de83d

### Checklist


packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -121,6 +122,13 @@
 
 	$: hasHadSelection = hasHadSelection || $selectedOptions.length > 0;
 	$: if ($selectedOptions && hasHadSelection) {
+		// console.log(label, opts.length, $selectedOptions);
+		// console.log($options);
+		// values should be $selectedOptions that are currently available in the opts array
+
+		if ($selectedOptions.length > opts.length) {
+			updateSelectedOptions(opts);
```
kwongz, line 130: I'm a little concerned that i'm passing in opts here, which contains the updated query array. Just feels weird to me

packages/ui/core-components/src/lib/atoms/inputs/dropdown/dropdownOptionStore.js
```diff
@@ -232,7 +232,12 @@ export const dropdownOptionStore = (opts = {}) => {
 			if (destroyed) return;
 			options.set(hygiene(get(options)));
 		},
-		destroy: () => (destroyed = true)
+		destroy: () => (destroyed = true),
+		updateSelectedOptions: (/** @type {import("@evidence-dev/sdk/usql").QueryValue} */ data) => {
```
kwongz, line 236: is this a correct type?

packages/ui/core-components/src/lib/atoms/inputs/dropdown/stories/Dropdown.stories.svelte
```diff
@@ -12,8 +12,12 @@
 	import { query } from '@evidence-dev/universal-sql/client-duckdb';
 	import DropdownOption from '../helpers/DropdownOption.svelte';
 	import DependentDropdowns from './DependentDropdowns.story.svelte';
-
+	import DataTable from '../../../../unsorted/viz/table/DataTable.svelte';
```
kwongz, line 15: should I remove this from dropdown stories? i personally like having datatables to visually see the data being used/filtered

zachstence: Yes, please remove it. The stories should generally have only the component the story is for.

packages/ui/core-components/src/lib/atoms/inputs/dropdown/dropdownOptionStore.js
zachstence: Yep!

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
zachstence: @ItsMeBrianD do you have any thoughts here? Doesn't seem bad to me, but there may be a cleaner way

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -121,6 +122,9 @@
 
 	$: hasHadSelection = hasHadSelection || $selectedOptions.length > 0;
 	$: if ($selectedOptions && hasHadSelection) {
+		if ($selectedOptions.length > opts.length) {
```
zachstence, line 125: Is this condition right? I played around with it a bit and it seems to be working correctly, but the condition just seems a little off. Would it make more sense to check if there is a selected option that no longer exists in `opts`?

```js
const selectedOptionsNoLongerInOpts = $selectedOptions.filter(
    (x) => !opts.some((y) => y.value === x.value)
);
```

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
zachstence: @ItsMeBrianD Do you have any thoughts here? I feel like there may be a cleaner implementation, but I can't put my finger on it

zachstence (approved):

zachstence:
Just noticed that this PR has caused a regression. Dropdowns no longer select the first option automatically.

For example: https://www.chromatic.com/test?appId=645d58d2e049d64800e0ee15&id=66eb3164c2b55e6f7a8c8026
The baseline shows "All" as its the first option and selected automatically
The new build shows "Test" which is the name of the Dropdown

There are several other Dropdown stories with the same kind of difference

I'm going to revert this PR and we can revisit

Should be easier to prevent future regressions with Chromatic (nearly) all cleaned up over here: https://github.com/evidence-dev/evidence/pull/2436

PR #2525, opened 2024-09-11
Order columns outside of DataTable helper components

csjh:
### Description

Cleans up column ordering code so future changes to `<Column />` and column order logic is easier

Not passing in components to `<DataTable {data} />` now does this

User writes
```svelte
<DataTable {data}>
```

We turn into
```svelte
<DataTable {data}>
   {#each column_in_data as id}
       <Column id={id} />
   {/each}
</DataTable>
```

### Checklist

~~- [ ] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)~~


archiewood:
> **Not** passing in components to <DataTable {data} /> now explicitly becomes

can you clarify what this statement means?

csjh:
One thing to note here is that this approach isn't friendly if we want to add more user-facing components that are used in the slot of `DataTable` - stuff like ```svelte
<DataTable>
    <SearchBar />
</DataTable>
```
but afaik we don't have plans for anything like that (and conventionally we've done props for those sorts of things)

zachstence:
@csjh Looks like the DataTable Reactive Columns story has an `undefined` column, is that expected?

image

csjh:
Yeah, that’s because of a lack of a `defaultValue` on the button

zachstence (approved):

PR #2528, opened 2024-09-13
Fix query compilation with special `$` sequences

### LINKED ISSUE #2455: [Bug]: Dollar sign ("$") in regex causes incorrect compiled SQL in chained query
### Describe the bug

When you have a query containing a regular expression ending with `$` then reference this query as part of another query, the `$` character and closing quotation are removed resulting in incorrect SQL code. 

### Steps to Reproduce

Say you have the query `all_companies`

```sql
SELECT * FROM companies WHERE name SIMILAR TO 'evi.+e$'
```
When you reference this query in another one say `some_companies` , the compiled SQL is not correct
```sql
SELECT * from ${all_companies}
```
results in the compiled SQL
```sql
SELECT * FROM (
    SELECT * FROM companies WHERE name SIMILAR TO 'evi.+e
)
```
which fails with the error `Parser Error: unterminated quoted string at or near "'^evi_.+ ...`

### Logs

_No response_

### System Info

_No response_

### Severity

Rewriting the regular expression to omit `$` prevents this problem

csjh:
### Description

Fixes #2455

### Checklist

### DIFF
--- a/.changeset/perfect-starfishes-relate.md
+++ b/.changeset/perfect-starfishes-relate.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/preprocess': patch
+---
+
+fix query compilation for queries that include String.prototype.replace special strings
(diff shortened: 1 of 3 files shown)


zachstence (changes requested):
@csjh Can you add some unit tests for this?

csjh:
added

zachstence (approved):

PR #2529, opened 2024-09-13
Fix frontmatter on home page

### LINKED ISSUE #2326: [Bug]: sidebar frontmatter control not working
### Describe the bug

The `sidebar` option in frontmatter does not seem to be having any effect. Should be able to set it to `hide` or `never` and have the appearance change.

### Steps to Reproduce

1. Create a `.md` page. 
1. In the frontmatter, include `sidebar: hide`
1. Check if the sidebar is hidden

### Logs

_No response_

### System Info

_No response_

### Severity

annoyance

Sidebar visibility can be set on a project-wide basis using a custom `+layout.svelte` file as described here: https://docs.evidence.dev/reference/themes-and-layouts/

csjh:
### Description

Fixes #2326

`src/pages/+page.md` was (accidentally?) being skipped in the `import.meta.glob` that generates the page manifest which was causing the home page to not get any of those options

### Checklist

### DIFF
--- a/.changeset/tidy-timers-smell.md
+++ b/.changeset/tidy-timers-smell.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/evidence': patch
+---
+
+fix frontmatter on home page
--- a/e2e/basic/pages/index.md
+++ b/e2e/basic/pages/index.md
@@ -1,5 +1,6 @@
 ---
 title: Welcome to Evidence
+sidebar: hide
 ---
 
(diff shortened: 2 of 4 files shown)


zachstence (changes requested):
@csjh Let's add an E2E test for this in `e2e/basic`. I can help you do that next week or you can take a stab at it (there's some E2E other tests you can use as examples)

zachstence (approved):

PR #2547, opened 2024-09-16
Override `dset` version to resolve security vulnerability

zachstence:
Resolves https://github.com/evidence-dev/evidence/security/dependabot/163


zachstence:
Branch name is incorrect, opening a new PR

PR #1048, opened 2023-07-18
Make page param comparison lowercase

csjh:
### Description

Fixes Netlify normalizing URLs to lowercase

PR #2553, opened 2024-09-18
Fix initial state not showing for some queries

csjh:
### Description

Fixes prerendered arrow files not being used as initial data when inputs are used in their corresponding queries

### Checklist


csjh:
3 failing tests - 2 related to batchup (expected, batchup logic was changed to avoid debounce on the server) and 1 related to initialData vs. noResolve priority change

packages/lib/preprocess/src/process-queries.cjs
```diff
@@ -71,35 +71,34 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
 
 				if (browser) {
 					// Data came from SSR
-					if (data.${id}) {
-						if (data.${id} instanceof Error) {
-							${id}InitialStates.initialError = data.${id}
+					// todo: combine _data + _columns + _length?
```
ItsMeBrianD, line 74:
```suggestion
```

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
```diff
@@ -53,7 +53,7 @@
 	const { results, update } = buildReactiveInputQuery(
 		{ value, data, label, order, where },
 		`ButtonGroup-${name}`,
-		$page?.data?.data[`ButtonGroup-${name}`]
+		$page?.data?.data[`ButtonGroup-${name}_data`]
```
ItsMeBrianD, line 56: I'm not super thrilled about this needing to have `_data` on here, but can fix that later

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -167,23 +175,26 @@
...
+				}, searchQuery.fetch());
 				// await tick();
 			}
 		} else {
 			finalQuery = query ?? data;
 		}
-	}, 100);
+	};
+
+	const updateQuery = browser ? debounce(_updateQuery, 100) : _updateQuery;
```
ItsMeBrianD, line 197: We need to make a browserDebounce function in the SDK lib that can be used in these cases - too much repetition of the same pattern for the same reason

packages/ui/core-components/src/lib/atoms/inputs/dropdown/helpers/DropdownOption.svelte
```diff
@@ -23,5 +24,8 @@
 
 	/** @type {import("../constants.js").EvidenceDropdownContext} */
 	const dropdownContext = getContext(DropdownContext);
+	if (!browser) {
+		onDestroy(dropdownContext.registerOption({ value, label: valueLabel, idx, __auto }));
```
ItsMeBrianD, line 28: Is onDestroy run on the server?

ItsMeBrianD (changes requested):

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
csjh: Maybe we add something like `static setSSRData` on `Query`?

packages/ui/core-components/src/lib/atoms/inputs/dropdown/helpers/DropdownOption.svelte
csjh: yes

mcrascal:
@csjh right now in the test case: https://deploy-preview-2553--evidence-test-env.netlify.app/categories/cursed%20sporting%20goods/ 

The default for the dropdown isn't being set on initial load

csjh:
Ah pretty sure that's netlify lowercase annoyingness ala #1048

packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
ItsMeBrianD: I think we leave it for now, but I was speaking specifically about `InputName_data` instead of `InputName`

zachstence (changes requested):
`core-components` needs to be included in your changeset

packages/lib/sdk/src/usql/query/Query.spec.js
```diff
@@ -589,16 +589,16 @@ describe('Query', () => {
 			expect(q.lengthLoaded).toBe(true);
 			expect(q.length).toBe(0);
 		});
-		it('should not execute the query, and ignore initial data if noResolve is set', () => {
+		it('should not execute the query and use initial data, even if noResolve is set', () => {
```
zachstence, line 592: This test name needs some revision, and the test itself could be improved too

**`should not execute the query`**
We need an assertion for this. Ideally expecting some function to not be called, or expecting some state to indicate the query hasn't been executed.

**`should not ... and use initial data`**
Can you clarify, does this mean we should use initial data, or should not? Either way, we need an assertion for this.

**`even if noResolve is set`**
This phrasing makes me think that `noResolve` isn't important here, and the outcome of the test will be the same *`even`* if `noResolve` is `false`. Is this the case? If so, we don't need to manually set `noResolve` in the test. If not, this test name should use *`if`* not *`even`*, like it was before.

packages/lib/sdk/src/utils/batchUp.js
```diff
@@ -10,7 +10,8 @@ export const batchUp = (fn, timeout = 200) => {
 	/** @type {Input[]} */
 	const collected = [];
 
-	const finalize = debounce(() => {
+	// todo: possibly algorithmic nightmare at high n on server?
```
zachstence, line 13: What about this is algorithmically problematic with a high number of collected items?

zachstence (dismissed):

packages/lib/sdk/src/utils/batchUp.js
csjh: iirc there was a codepath that ran O(n) work n times for n = number of options that was worked around via `batchUp`, not sure if that's still around though

zachstence: Sounds like the function being used in `batchUp` could potentially be problematic, but not `batchUp` itself?

I think we should remove this log or relocate to the specific codepath you were thinking of

zachstence: [f7ffb48](https://github.com/evidence-dev/evidence/pull/2553/commits/f7ffb4860fcfa4d1d6e8dcc9dc30b942ec0b52dd)

ItsMeBrianD (approved):

zachstence:
@mcrascal has confirmed this is working as intended

zachstence:
@mcrascal has QAed and confirmed this is working as intended

PR #2574, opened 2024-09-23
Hotfix | Prerendering

ItsMeBrianD:
### Checklist


packages/lib/preprocess/src/process-queries.cjs
```diff
@@ -42,79 +42,81 @@ const createDefaultProps = function (filename, componentDevelopmentMode, duckdbQ
...
+                let ${id};
+
+                $: __${id}HasUnresolved = hasUnsetValues\`${duckdbQueries[id].compiledQueryString.replaceAll('`', '\\`')}\`;
+                $: __${id}Text = \`${duckdbQueries[id].compiledQueryString.replaceAll('`', '\\`')}\`
+
+                // keep initial state around until after the query has resolved once
+                let __${id}InitialFactory = false;
+                $: if (__${id}HasUnresolved || !__${id}InitialFactory) {    
+                    if (!__${id}HasUnresolved) {
+                        __${id}Factory(__${id}Text, { noResolve: __${id}HasUnresolved, ...${id}InitialStates });
```
csjh, line 97: not 100% sure this part was required for the fix, need to check

packages/lib/sdk/src/usql/query/Query.js
```diff
@@ -367,8 +367,12 @@ ${this.text.trim()}
...
 		return resolved;
 	};
-	fetch = async () => {
-		return Promise.allSettled([this.#fetchColumns(), this.#fetchData()]).then(() => this.value);
+	fetch = () => {
+		const cols = this.#fetchColumns();
+		if (cols instanceof Promise) {
+			return Promise.allSettled([this.#fetchColumns(), this.#fetchData()]).then(() => this.value);
+		}
+		this.#fetchData();
```
csjh, line 375:
```suggestion
		this.#fetchData();
		return Promise.resolve();
```
I think we should return `Promise.resolve()` so people can assume the return value is thenable

packages/lib/sdk/src/usql/query/Query.js
```diff
@@ -367,8 +367,12 @@ ${this.text.trim()}
 		);
 		return resolved;
 	};
-	fetch = async () => {
-		return Promise.allSettled([this.#fetchColumns(), this.#fetchData()]).then(() => this.value);
+	fetch = () => {
+		const cols = this.#fetchColumns();
+		if (cols instanceof Promise) {
+			return Promise.allSettled([this.#fetchColumns(), this.#fetchData()]).then(() => this.value);
```
csjh, line 373:
```suggestion
			return Promise.allSettled([this.#fetchColumns(), this.#fetchData()]);
```
either return `this.value` in the sync version or stop returning `this.value` in either path

packages/lib/sdk/src/usql/query/Query.js
ItsMeBrianD: This means that `instanceof Promise` will always be true though, so the current way we check for async code could no longer wrok

csjh (approved):

PR #2578, opened 2024-09-24
Map legends

kwongz:
### Description

Category and Scalar Legend for maps, Users select a column using the value prop and can create a scalar or categorical legend, customizing colors using the colorPallete.

https://github.com/user-attachments/assets/f5c2646c-d441-4a95-a33f-77dcb041d92f

image

### Checklist


kwongz:
image

packages/ui/core-components/src/lib/unsorted/viz/core/ErrorChart.svelte
```diff
@@ -15,7 +15,7 @@
 
 <div
 	width="100%"
-	class="grid grid-rows-auto box-content grid-cols-1 justify-center bg-red-50 text-grey-700 font-ui font-normal rounded border border-red-200 min-h-[{minHeight}px] py-5 px-8 my-5 print:break-inside-avoid"
+	class="grid grid-rows-auto box-content grid-cols-1 justify-center bg-red-50 text-grey-700 font-ui font-normal rounded border border-red-200 min-h-[{minHeight}px] py-5 px-8 my-5 print:break-inside-avoid relative z-[500]"
```
kwongz, line 18: Fixes error from being covered by leaflet map overlay

packages/ui/core-components/src/lib/unsorted/viz/map/components/Points.svelte
```diff
@@ -210,6 +213,64 @@
 
 	let values, minValue, maxValue, colorScale, sizeExtents, maxData, maxSizeSq;
 
+	/** @type {[string]} */
```
csjh, line 216:
```suggestion
	/** @type {string[]} */
```
this is the type of a single string tuple

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend.svelte
```diff
@@ -0,0 +1,92 @@
...
+							<span class="absolute text-[10px] left-0 top-2 block">{minValue}</span>
+							<span class="absolute text-[10px] right-0 top-2 block">{maxValue}</span>
+						</span>
+					</div>
+				{:else if legendType === 'category'}
+					<div class="overflow-y-auto max-h-52">
+						{#each colorPalette as color, i}
+							<div class="w-full pr-2 pl-2">
+								<span class="inline-block h-2 w-2 rounded-full" style="background-color: {color}" />
+								- {values[i] ? values[i] : `No value`}
```
csjh, line 84:
```suggestion
								- {values[i] || "No value"}
```

packages/ui/core-components/src/lib/unsorted/viz/map/components/Points.svelte
```diff
@@ -219,11 +280,46 @@
...
 			checkInputs(data, [lat, long]);
 			values = $data.map((d) => d[value]);
+			colorPalette = colorPalette.map((item) => chroma(item).hex());
+
+			if (legendType === 'category') {
+				let uniqueValues = new Set(values);
+				values = [...uniqueValues];
+				let i = 0;
+
+				while (colorPalette.length < values.length) {
```
csjh, line 290: I think I would prefer this to become a for loop but not that big of a deal

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -33,6 +33,9 @@ export class EvidenceMap {
 	/** @type {HTMLDivElement | undefined} */
 	#mapEl;
 
+	/** @type {import('svelte/store').Writable<object>} */
```
csjh, line 36: a better type for this would be nice

packages/ui/core-components/src/lib/unsorted/viz/map/PointMap.svelte
```diff
@@ -52,6 +53,16 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {[string]|undefined} */
+	export let colorPalette = undefined;
+
+	/** @type {string} */
```
csjh, line 59:
```suggestion
	/** @type {'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight'} */
```

packages/ui/core-components/src/lib/unsorted/viz/map/PointMap.svelte
```diff
@@ -52,6 +53,16 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {[string]|undefined} */
```
csjh, line 56:
```suggestion
	/** @type {string[]|undefined} */
```

packages/ui/core-components/src/lib/unsorted/viz/map/components/Areas.svelte
```diff
@@ -202,9 +264,46 @@
 
 		values = $data.map((d) => d[value]);
 
-		colorScale = chroma
-			.scale(colorPalette)
-			.domain([min ?? Math.min(...values), max ?? Math.max(...values)]);
+		//new
+		colorPalette = colorPalette.map((item) => chroma(item).hex());
+
+		if (legendType === 'category') {
```
csjh, line 270: this logic shouldn't be repeated verbatim here and Points.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Areas.svelte
```diff
@@ -265,6 +364,20 @@
 			await data.fetch();
 			await processAreas();
 		})();
+
+	function handleColor(item, value) {
```
csjh, line 368: also shouldn't be duplicated

csjh (commented):
I think the duplicated stuff between `Areas.svelte` and `Points.svelte` is my only major-ish concern

packages/ui/core-components/src/lib/unsorted/viz/map/Legend.svelte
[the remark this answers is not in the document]
kwongz: I think this was made accidentally, i'll remove it. Thanks!

hughess:
This is looking great. A couple of stylistic things:

- I think having the separate hide/show button takes up too much space. Would it be possible to use an icon with an arrow on the legend itself to indicate that it can be closed?
- I don't think the legend needs a title
- I would remove the hyphens in the categorical legend between the color circles and the labels
- For the scalar legend, I think this should allow for number formatting
   - Use the column's existing value format if set
   - Allow for an override of format

kwongz:
@hughess 


- I think having the separate hide/show button takes up too much space. Would it be possible to use an icon with an arrow on the legend itself to indicate that it can be closed?

Currently, the legend slides off and fades and gets removed from the DOM, Without the button there is no way to click on the legend in hidden mode. 

Would you prefer the legend to collapse into a small bar? or perhaps a smaller hide/show legend icon button instead of the text button?

- [x] I don't think the legend needs a title
- [x] I would remove the hyphens in the categorical legend between the color circles and the labels
- [x] For the scalar legend, I think this should allow for number formatting

hughess:
I think the button and the legend should be part of the same object visually - it's either minimized or expanded. Which I think means the legend gets replaced in the DOM by the "show legend" button or icon

archiewood:
Two stylistic remarks:
- Font: The font for the legend does not appear to match the font for the tooltip exactly - do we have control of this?
- Hyphen: I am not a fan of the dash between the legend spot and the Label - can we remove this and/or replace with whitespace?
CleanShot 2024-09-30 at 10 58 04@2x

kwongz:
@csjh Refactored the components, feels much cleaner. Thanks!

@hughess  
- Reworked the styling from your comments. 
- Note: Existing "categorical" legends method will note break ** see stories

@archiewood 
- Thanks for the font style, added legends font to be linked to tooltip font.

hughess:
@kwongz one more thing I just remembered - does this PR include a default color palette?

kwongz:
@hughess  We have a backup colors array we refer to if no colorPallete was specified. The backup colors also get used if the user has more categorizes then colors.

e.g 

1. categories.length = 5
2. colorPallete.length = 4
3. This will result in pulling from backup colors and assigning the 5th category with a unique color

Now that you mention it i will need to check behaviour with scalar legend

kwongz:
@hughess 

Here's a story for defaultColors with legends (no color palette prop). The backup color array consists of randomly selected, distinct colors I've added, with about 30 colors in total. I think it would be helpful to have a second pair of eyes review evidence-themed backup colors.

Note: The behavior for categorical legends with no color palette is to use the initial default colors: export let colorPalette = ['lightblue', 'darkblue'];, and then fall back on our backup colors.

image

archiewood:
Re the show / hide feature:
- Is this neccessary?
- I would suggest no one would ever want to hide the linear type legend, as it occupies so little space

kwongz:
Thoughts?

https://github.com/user-attachments/assets/fc5d4b8f-70ec-4a94-b916-0d29df3c52aa

archiewood:
> I would like to keep some functionality for the category legend, as it can span the height of the map.

Makes sense for category

> The min and max values appear quite close together, make it wider to add more space.

Yes I think that is a good idea. These could almost be arbitrarily long values

> Which in turn maybe justifies a minimization functionality.

Personally I don't think it's necessary - the user can easily move the map, and it takes up a very small % of the overall map. I'd be inclined to remove it

archiewood:
Final nits:
- I'd suggest that the chevron for the top example should be at the bottom of the legend - that way the user doesnt have to move the mouse between show and hide events - the button should remain stationary. For clarity, this would be the button would be at the top if the legend was top-left, and at the bottom if the legend was bottom-left
- The directions of the chevrons don't make sense to me visually. 
   - The Top example should be down, then up. The Bottom example should be left, then right. They should point in the direction the legend will move when pressed.

packages/ui/core-components/src/lib/unsorted/viz/map/AreaMap.svelte
```diff
@@ -59,6 +60,13 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {string} */
+	export let legendPosition = 'topLeft';
+	/** @type {string | undefined} */
```
zachstence, line 65:
```suggestion
	/** @type {'category' | 'scalar' | undefined} */
```

packages/ui/core-components/src/lib/unsorted/viz/map/AreaMap.svelte
```diff
@@ -59,6 +60,13 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {string} */
```
zachstence, line 63:
```suggestion
	/** @type {'topLeft' | 'topRight' | 'bottomRight' | 'bottomLeft'} */
```

packages/ui/core-components/src/lib/unsorted/viz/map/BubbleMap.svelte
```diff
@@ -58,6 +59,13 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {string} */
+	export let legendPosition = 'topLeft';
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 65: See type changes comment on `AreaMap.svelte`

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -431,4 +499,84 @@ export class EvidenceMap {
...
+		this.#legendData.update((legendData) => ({
+			...legendData, // Keep existing data
+			colorPalette,
+			values: arrayOfStringValues, // Make sure to update this property
+			minValue,
+			maxValue
+		}));
+	}
+
+	get getLegendData() {
```
zachstence, line 560: `get get` is redundant

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -431,4 +499,84 @@ export class EvidenceMap {
...
+		//determine legend style
+		if (legendType === 'category') {
+			let uniqueValues = new Set(values);
+			values = [...uniqueValues];
+			let i = 0;
+
+			while (colorPalette.length < values.length) {
+				if (!backupColors[i]) {
+					throw new Error('No more backup colors available.');
+				}
```
zachstence, line 513: I think its better to wrap around (i.e. `backupColors[i % backupColors.length]`) and reuse colors rather than throw an error

packages/ui/core-components/src/lib/unsorted/viz/map/PointMap.svelte
```diff
@@ -52,6 +53,16 @@
 	/** @type {string|undefined} */
 	export let title = undefined;
 
+	/** @type {string[]|undefined} */
+	export let colorPalette = undefined;
+
+	/** @type {'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight'} */
+	export let legendPosition = 'topLeft';
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 62: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendIcons.svelte
```diff
@@ -0,0 +1,40 @@
+<script>
+	/** @type {boolean} */
+	export let hideLegend = false;
+	export let handleLegendClick;
+
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 7: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendIcons.svelte
```diff
@@ -0,0 +1,40 @@
...
+	/** @type {'left' | 'right' | 'bottom' | 'top'} */
+	export let direction = 'bottom';
```
zachstence, line 9: `bottom` and `top` aren't directions, they should be `down` and `up`, respectively

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendIcons.svelte
```diff
@@ -0,0 +1,40 @@
+<script>
+	/** @type {boolean} */
+	export let hideLegend = false;
+	export let handleLegendClick;
```
zachstence, line 4: Need a type here
```suggestion
	/** @type {() => void} */
	export let handleLegendClick;
```

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendIcons.svelte
zachstence, line 1: Lets call this `LegendToggle.svelte` since it is a button that opens/closes the legend

packages/ui/core-components/src/lib/unsorted/viz/map/components/Areas.svelte
```diff
@@ -39,6 +38,8 @@
 
 	/** @type {string | undefined} */
 	export let name = undefined;
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 42: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Bubbles.svelte
```diff
@@ -9,6 +9,8 @@
 	export let size = undefined; // column containing size data
 	/** @type {number} */
 	export let opacity = 0.8;
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 13: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendTypes.svelte
zachstence, line 1: Let's call this `LegendContent.svelte` because it renders the contents of the legend, not a legend type.

If you wanted, you could also break out `CategoricalLegend.svelte` and `ScalarLegend.svelte` into separate components, but since they're pretty simple I don't mind them combined in one.

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendTypes.svelte
```diff
@@ -0,0 +1,52 @@
+<script>
+	import { fmt } from '@evidence-dev/component-utilities/formatting';
+	/** @type {string | undefined} */
+	export let legendType = undefined;
```
zachstence, line 4: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend.svelte
```diff
@@ -0,0 +1,93 @@
...
+		hideLegend = !hideLegend;
+	};
+
+	if (legendType === 'scalar') {
+		direction = legendPosition.includes('Left') ? 'left' : 'right';
+	} else if (legendType === 'category') {
+		direction = legendPosition.includes('bottom') ? 'bottom' : 'top';
+	}
+
+	console.log(direction, legendPosition);
```
zachstence, line 57:
```suggestion
```

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend.svelte
```diff
@@ -0,0 +1,93 @@
...
+	export const evidenceInclude = true;
+</script>
+
+<script>
+	import { mapContextKey } from '../constants.js';
+	import { getContext } from 'svelte';
+	import LegendIcons from './Legend Components/LegendIcons.svelte';
+	import LegendTypes from './Legend Components/LegendTypes.svelte';
+	export let legendType = undefined;
+	export let legendPosition = 'bottomLeft';
```
zachstence, line 11: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend.svelte
```diff
@@ -0,0 +1,93 @@
...
+
+	const handleLegendClick = () => {
+		hideLegend = !hideLegend;
+	};
+
+	if (legendType === 'scalar') {
+		direction = legendPosition.includes('Left') ? 'left' : 'right';
+	} else if (legendType === 'category') {
+		direction = legendPosition.includes('bottom') ? 'bottom' : 'top';
+	}
```
zachstence, line 55: Why do we expose `direction` as a prop above if we are overriding it here?

packages/ui/core-components/src/lib/unsorted/viz/map/components/Points.svelte
```diff
@@ -35,6 +33,9 @@
 	export let sizeFmt = undefined;
 	/** @type {number|undefined} */
 	export let size = undefined; // point size
+	/** @type {string|undefined} */
+	export let legendType = undefined;
```
zachstence, line 37: See type changes comment on AreaMap.svelte

packages/ui/core-components/src/lib/unsorted/viz/map/components/Points.svelte
```diff
@@ -287,7 +291,9 @@
 		<Point
 			{map}
 			options={{
-				fillColor: color ?? (value ? colorScale(item[value]).hex() : uiColours.blue700), // Fill color of the circle
+				// kw note:
+				//need to clean this logic
+				fillColor: color ?? map.handleFillColor(item, value, values, colorPalette, colorScale),
```
zachstence, line 296: Any changes needed here?

zachstence (changes requested):
Nice work, the legends look great!

Notes:
- Can authors use the `<Legend />` component directly? Should we add docs to the `base-map.md` page explaining how?
- I think calling it a `categorical` legend makes more sense than a `category` legend (`legendType`)
- Types can be improved in several places

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendIcons.svelte
```diff
@@ -0,0 +1,40 @@
...
+		left: { show: 'rotate-0', hide: 'rotate-180' },
+		right: { show: 'rotate-180', hide: 'rotate-0' }
+	};
+</script>
+
+<button
+	class="items-center {legendType === 'scalar' ? 'h-10 px-[2.9px]' : `h-5 pl-2 pt-0.5 w-full`}"
+	on:click={handleLegendClick}
+	on:dblclick={(e) => e.stopPropagation()}
+>
```
zachstence, line 23: Add an `aria-label` to this button that says either `Expand Legend` or `Collapse Legend` so that screenreader users know what it does

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
kwongz: True

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend Components/LegendTypes.svelte
kwongz: Sounds good, i'll leave as is for now, but I'll keep in mind of refactoring this and things like this in the future.

packages/ui/core-components/src/lib/unsorted/viz/map/components/Legend.svelte
kwongz: Sorry about that, was moving remotely to my laptop and was mid-code. I'll remove it

packages/ui/core-components/src/lib/unsorted/viz/map/components/Points.svelte
kwongz: nope, it was cleaned, thanks! will remove comment

kwongz:
@zachstence I didnt design the user to be able to interact with the legend component directly
image

currently the legend get rendered based on the type of legend they input "categorical" | "scalar" else if undefined does not render.

Legends is currently available for Points, bubbles and area. @hughess we can add it to basemap if needed, seems appropriate to me.

zachstence:
> @zachstence I didnt design the user to be able to interact with the legend component directly 

If it shouldn't be used by authors, we shouldn't have `export const evidenceInclude = true;`

https://github.com/evidence-dev/evidence/pull/2578/files#diff-7553509e2432a61e55a5b52d560f0f7527d86ecfa25f9546a235aee1acbccee5R2

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
kwongz: **Handling End of Color Options:**

@hughess, how do you feel about managing categorical values that exceed available color options, including both the color palette and backup colors?

@zachstence suggested avoiding throwing an error, instead opting to indicate a warning to the user.

We are considering using chroma.random() to generate new distinct colors. This approach would ensure that colors are always assigned to any number of categories, effectively handling dynamic categorical lists when we reach the end of the color palette and backup colors.

**_Caveat:_** The randomly generated colors would be independent on each refresh and across different browsers. For instance, the colors generated on your browser may differ from those generated on mine.

hughess: I would have it start at the beginning of whatever the first color palette is. So if they pass a custom palette, it should go back to the first color in that palette and repeat. 

If they don't pass a custom palette, it should go back to the first built-in color and go from there.

This is the same way the charts operate at the moment. In the future, if it becomes an issue we can just add a bunch of colors to the built-in palette to accommodate longer lists

kwongz:
@hughess  @zachstence 


https://github.com/user-attachments/assets/a5043c67-c609-47b8-8887-79fd1acbfe4b


Video Examples
1. 4 colors provided in color palette + categories.length > colorPalette.length (introduces looping)
2. no colors provided in color palette + categories.length > colorPalette.length (introduces looping)
3. no colors provided in color palette

**Changes**
1. Use mapColors array as default, if color palette = undefined
4. loop through color pallete if categories.length > colorPalette.length

Note: mapColors is a copy of chartColors - grey scale colors. I think the greyscale colors are too hard to see on the maps. 
We can also add more colors to here if we don't want to mutate chart colors. 

image

archiewood:
Looks great!

zachstence (approved):

hughess:
Great job @kwongz! This is a big improvement for maps!

PR #2590, opened 2024-09-27
DataTable sort feature

hughess:
### Description
Adds ability to define sort in DataTable code using `sort`, where you can pass one column and a sort direction ("asc" or "desc")

For groups (using `groupBy`), groups will be sorted according to subtotals where applicable.

#### Basic example
```html
<DataTable data={flights} sort="fare desc"/>
```

CleanShot 2024-09-27 at 13 48 45@2x

#### Simple group example

```html
	<DataTable {data} groupBy="category" sort="sales" subtotals=true>
		<Column id="category" />
		<Column id="item" />
		<Column id="sales" fmt="usd" />
	</DataTable>
```
CleanShot 2024-09-27 at 13 42 42@2x

### Checklist


packages/ui/core-components/src/lib/unsorted/viz/table/DataTable.stories.svelte
```diff
@@ -50,14 +55,46 @@
...
 		query
 	)}
-	<DataTable {data} title="Flights" search groupBy="regulator">
+	<DataTable
+		{data}
+		title="Flights"
+		search
+		groupBy="regulator"
+		sortBy="regulator"
+		sortAsc="true"
```
archiewood, line 64: this is inconsistent with the API on some of our other components, where you would do

`sort="column_name desc"`

hughess: good point. I was thinking about that one, but thought it would be harder to give clear error messages. But I think consistency is more important, especially once we move the components to accepting more SQL fragments

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -44,9 +44,18 @@
...
-	export let sortBy = undefined;
-	export let sortAsc = false;
-	$: sortAsc = sortAsc === 'true' || sortAsc === true;
+	export let sort = undefined;
+	let sortBy = undefined;
+	let sortAsc = undefined;
+	$: if (sort) {
+		const [column, direction] = sort.split(' ');
+		sortBy = column;
+		sortAsc = direction === 'asc' ? true : false; // Default to descending if no direction is provided
```
archiewood, line 53: this is the opposite to SQL standard

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -44,9 +44,18 @@
...
 
 	// Sort props
-	export let sortBy = undefined;
-	export let sortAsc = false;
-	$: sortAsc = sortAsc === 'true' || sortAsc === true;
+	export let sort = undefined;
+	let sortBy = undefined;
+	let sortAsc = undefined;
+	$: if (sort) {
+		const [column, direction] = sort.split(' ');
```
archiewood, line 51: this feels fragile but fully be sure why.

is this allowed?
```svelte
<DataTable 
  data={query}
  sort="column_one, column_two desc"
/>
```

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
hughess: Hmm yea I don't like that asc is default, but I guess if we move to sql snippets this will get confusing. Will update this

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
hughess: Wasn't contemplated - I'll take a look

hughess: This isn't supported - only allows for one column in this version.

I had the same thought about this seeming fragile. If there's a mistake, at the moment it fails silently. 

Might be worth including error messages for common scenarios people might run into. Some that come to mind:
- Attempting to use multiple columns in sort
- Misspelling column name
- Misspelling direction asc or desc
- Not wrapping column + direction in quotes in the prop

Any others?

hughess:
Last remaining issue on this:

- Initial sort affects the data shown in the query viewer. Subsequent sorts via clicking columns do not affect it. This happens even with an explicit `order by` in the sql query

sites/docs/pages/components/data-table.md
```diff
@@ -97,6 +98,28 @@ This example includes a `custom_format` column, which contains a different curre
...
+</DataTable>
+```
+
+<DataTable data={orders_summary} sort="sales desc">
+    <Column id=category/> 
+    <Column id=item/> 
+    <Column id=sales fmt=usd/> 
+</DataTable>
+
+<DataTable data={orders_summary} sort="sales asc">
```
kwongz, line 117: @hughess Just wanted to double check, i think you wanted to remove this second sort example?

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -250,68 +278,78 @@
...
 				? -1 * sortModifier
 				: (forceTopOfAscending(b[column]) && !forceTopOfAscending(a[column])) ||
 					  a[column] > b[column]
 					? 1 * sortModifier
 					: 0;
-		data.sort(sort);
-		filteredData = filteredData.sort(sort);
+
+		// data.sort(comparator);
+		const sortedFilteredData = [...filteredData].sort(comparator);
```
kwongz, line 312: replaced filteredData = filteredData.sort(sort); to not mutate the original data

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -250,68 +278,78 @@
...
 
-		const sort = (a, b) =>
+		const comparator = (a, b) =>
 			(forceTopOfAscending(a[column]) && !forceTopOfAscending(b[column])) || a[column] < b[column]
 				? -1 * sortModifier
 				: (forceTopOfAscending(b[column]) && !forceTopOfAscending(a[column])) ||
 					  a[column] > b[column]
 					? 1 * sortModifier
 					: 0;
-		data.sort(sort);
```
kwongz, line 276: removed data.sort(), don't think its being used

.changeset/lemon-cherries-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Add sortBy to DataTable
```
zachstence, line 5:
```suggestion
Add sort prop to DataTable
```

zachstence, line 5:
```suggestion
Add sort prop to DataTable
```

packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
```diff
@@ -10,8 +10,7 @@
 
 	let error;
 
-	const identifier = Symbol();
-
+	export let identifier = Symbol();
```
zachstence, line 13: Why does this need to be a prop? Doesn't appear to be used anywhere.

packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
```diff
@@ -81,7 +80,7 @@
 		blue: ['white', 'hsla(198, 56%, 56%,1)']
 	};
 
-	let colorPalette;
+	export let colorPalette;
```
zachstence, line 83: Why does this need to be a prop? If it does for whatever reason, we need docs for it.

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -163,6 +177,20 @@
...
 		columnSummary = getColumnSummary(data, 'array');
 
+		// Check if sort column is in table
+		if (sortBy) {
+			if (!columnSummary.map((d) => d.id).includes(sortBy)) {
+				throw Error(
+					sortBy +
+						' is not a column in the dataset. sort should contain one column name and optionally a direction (asc or desc). E.g., sort=my_column or sort="my_column desc"'
+				);
+			}
```
zachstence, line 187: Use a template literal

```suggestion
					`${sortBy} is not a column in the dataset. sort should contain one column name and optionally a direction (asc or desc). E.g., sort=my_column or sort="my_column desc"`
```

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -163,6 +177,20 @@
...
+
+			if (sortDirection && !['asc', 'desc'].includes(sortDirection)) {
+				throw Error(sortDirection + ' is not a valid sort direction. Please use asc or desc');
```
zachstence, line 190: Use a template literal

```suggestion
				throw Error(`${sortDirection} is not a valid sort direction. Please use asc or desc`);
```

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -250,68 +278,78 @@
 	// SORTING
 	// ---------------------------------------------------------------------------------------
 
-	let sortBy = { col: null, ascending: null };
-
-	$: sort = (column) => {
-		if (sortBy.col == column) {
-			sortBy.ascending = !sortBy.ascending;
+	$: sortClick = (column) => {
+		if (sortObj.col == column) {
```
zachstence, line 282: Use `===`

```suggestion
		if (sortObj.col === column) {
```

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -250,68 +278,78 @@
...
 			(forceTopOfAscending(a[column]) && !forceTopOfAscending(b[column])) || a[column] < b[column]
 				? -1 * sortModifier
 				: (forceTopOfAscending(b[column]) && !forceTopOfAscending(a[column])) ||
 					  a[column] > b[column]
 					? 1 * sortModifier
 					: 0;
-		data.sort(sort);
-		filteredData = filteredData.sort(sort);
+
+		// data.sort(comparator);
```
zachstence, line 311: Uncomment if needed or delete if not needed

zachstence (changes requested):

packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
hughess: this is used only internally - I added `export` to avoid some browser console warnings, but it's not needed for a user. Will remove

zachstence (approved):

PR #2607, opened 2024-10-02
Speed up `buildMultipartParquet`

csjh:
### Description

Switches to a custom written `jsToIPC` to elude overhead from going through multiple forms to use `tableToIPC`

Right now only accepts the data types we accept from data sources, should be relatively easy to expand in the future

### Checklist


ItsMeBrianD (changes requested):
This needs more (and more specific) tests for edge cases

csjh:
closing into stasis until we can determine this is a major bottleneck (as opposed to raw network)

PR #2608, opened 2024-10-03
Add sparkline and bar viz to DataTable

hughess:
### Description
CleanShot 2024-10-03 at 15 46 29@2x

#### Sparkline
- Fix spacing for sparkline - it sits on 'baseline', which looks worse than it being vertically centered
- Decide whether sparkline should allow `interactive=true`, since it affects spacing and labels can get cut off. Maybe enabling this should auto-add margin to the cell
- Decide on naming for sparkline props - simplest is to prepend 'sparkline' to everything, but then it's a bit verbose
- Include clear example of `array_agg` in docs for use with sparkline
- Add `array_agg` to duckdb autocomplete functions in VS Code extension
- Decide how to handle height adjustments to the sparkline - should this affect table row height?

#### Bar viz
- Add props to change:
   - bar color
   - background color
   - enable/disable background color
   - label alignment (possibly separate from column header alignment as set via `align` prop in `Column`)
- Need to figure out color contrast management
   - In some cases, the number label will overlap the edge of the bar, so may have a dark color on one side and light on the other - so luminosity calc won't help
   - One option is to apply a semi-transparent background to the label if it overlaps the bar edge, but that would require knowing if it will overlap (or adding the background to all labels, which would probably look bad)
   - Another option is to apply a text border to all labels. Could be white outline. Best is probably white font with outline color matching the bar color
- Need to figure out alignment of labels - left, center of bar, end of bar, right?
   - Where should column header go in these scenarios?
- Should we offer any control over column width? Imagining having 3 of these bar columns side by side - you want them to be identical width when doing a quick visual scan of the data   

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
@@ -103,6 +103,12 @@
 	export let chip = false;
 	$: chip = chip === 'true' || chip === true;
 
+	// Sparkline:
+	export let sparklineType = 'line'; // line, area, or bar
+	export let sparklineColor = undefined;
+	export let sparklineDateCol = undefined;
+	export let sparklineValueCol = undefined;
+
 	// Column Groups:
 	export let colGroup = undefined;
 
@@ -146,7 +152,11 @@
 		colorMid,
 		colorBreakpoints,
 		colorPalette,
-		redNegatives
+		redNegatives,
+		sparklineType,
+		sparklineColor,
+		sparklineDateCol,
+		sparklineValueCol
 	};
 
 	/**
--- a/packages/ui/core-components/src/lib/unsorted/viz/table/DataTable.stories.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/table/DataTable.stories.svelte
@@ -34,6 +34,56 @@
 	<DataTable {data} />
 </Story>
 
+<Story name="Bar Viz">
+	{@const data = Query.create(
+		`
+	SELECT 'a' as category, 100 as value
+	union all
+	SELECT 'a' as category, 80 as value
+	union all
+	SELECT 'a' as category, 70 as value
+	union all
+	SELECT 'b' as category, 30 as value
+	union all
+	SELECT 'b' as category, 24 as value
+	union all
+	SELECT 'b' as category, 12 as value
+	`,
+		query
+	)}
+	<DataTable {data}>
+		<Column id="category" />
+		<Column id="value" contentType="bar" fmt="usd" align="center" />
+	</DataTable>
+</Story>
+
+<Story name="Sparkline">
+	{@const data = Query.create(
+		`
+	select category, array_agg({'date': date, 'value': value}) as sparkline from (
+	SELECT 'Grocery' as category, '2024-01-01'::date as date, 100 as value
+	union all
+	SELECT 'Grocery' as category, '2024-01-02'::date as date, 80 as value
+	union all
+	SELECT 'Grocery' as category, '2024-01-03'::date as date, 70 as value
+	union all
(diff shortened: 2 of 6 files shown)


packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
```diff
@@ -103,6 +103,12 @@
 	export let chip = false;
 	$: chip = chip === 'true' || chip === true;
 
+	// Sparkline:
+	export let sparklineType = 'line'; // line, area, or bar
+	export let sparklineColor = undefined;
+	export let sparklineDateCol = undefined;
```
archiewood, line 109: `"sparklineType" = line` === `sparkLine` ?
`"sparklineType" = bar` === `sparkBar` ?

`sparkColor`
`sparkDate`
`sparkValue`

sites/example-project/src/pages/tables/data-table/+page.md
```diff
@@ -46,4 +46,41 @@ group by all
...
 	<Column id=orders/> 
 	<Column id=aov fmt=usd2 contentType=colorscale scaleColor={['#b52626','#FFFFFF','#2e9939']}/> 
- </DataTable>
\ No newline at end of file
+ </DataTable>
+
+## Sparkline
+
+ ```sql cats
+WITH daily_sales AS (
```
archiewood, line 54: monthly_sales

sites/example-project/src/pages/tables/data-table/+page.md
```diff
@@ -46,4 +46,41 @@ group by all
...
+<DataTable data={cats}>
+    <Column id=category/>
+    <Column id=sales contentType=sparkline sparklineDateCol=date sparklineValueCol=sales sparklineType=area/>
+</DataTable>
+
+## Bar Viz
+
+<DataTable data={summary}>
+  <Column id=category/>
+  <Column id=sales contentType=bar fmt=usd align=left/>
```
archiewood, line 82: this is very clean syntax

packages/ui/core-components/src/lib/unsorted/viz/table/DataTable.stories.svelte
```diff
@@ -34,6 +34,56 @@
...
+	SELECT 'Retail' as category, '2024-01-02'::date as date, 24 as value
+	union all
+	SELECT 'Retail' as category, '2024-01-03'::date as date, 12 as value
+	) group by all
+	`,
+		query
+	)}
+	<DataTable {data}>
+		<Column id="category" />
+		<Column id="sparkline" contentType="sparkline" sparklineType="area" sparklineColor="green" />
```
archiewood, line 81: `contentType=sparkArea`

archiewood:
Some initial UI thoughts.

### Sparkline
- Sparkline should stretch to width of column?
- No area under line by default
   - SparkArea feels not useful I think - could remove, but it's as easy to leave in so might as well
- Disabling interactivity by default
- perhaps should accept arrays also / structs with with no date column? What if I just have a sequence of values I am trying to show distribution of?


### Prior Art
- Some excellent [Tufte](https://www.edwardtufte.com/notebook/sparkline-theory-and-practice-edward-tufte/)
- What does excel do?
  - Defaults never have a dark background for bars - so black always works
    - other bad color combos are a userland problem?
CleanShot 2024-10-04 at 01 30 22@2x

packages/ui/core-components/src/lib/unsorted/viz/table/Column.svelte
hughess: that's a good idea, i like having multiple content types for these

hughess:
@archiewood I think your comments have all been addressed now - let me know what you think!

ItsMeBrianD (approved):

Issue #2560, opened 2024-09-19
[Bug]: Map markers don't always render in the same order


zachstence:
### Describe the bug

When using a `BaseMap` with multiple `Bubbles`/`Points`, the components don't always mount in the same order, and thus the `CircleMarkers` added to the map aren't consistently ordered.

### Steps to Reproduce

Load the BaseMap story a couple times and see that sometimes the red dots are underneath the blue bubbles, sometimes they are on top.

![image](https://github.com/user-attachments/assets/6af53d80-51d0-46f0-85aa-93c96ff87151)

![image](https://github.com/user-attachments/assets/e6e2f753-8410-4e9e-aca8-04845e8bfbf1)

### Severity

annoyance

### Additional Information, or Workarounds

Frustratingly, `CircleMarker` doesn't have `zIndexOffset` like `Marker` has

I think the best solution here is to `map.createPane` for each `Bubbles`/`Points` component that is used, and then [adjust the z-index of the panes](https://stackoverflow.com/questions/49501295/display-leaflet-circlemarker-on-top-of-other-markers) based on their appearance in the DOM (not sure how to do this, but a hacky solution could be rendering some invisible divs and checking their order).

PR #2612, opened 2024-10-04
fixed rendering layers Bubbles + Points

kwongz:
### Description
Points + Bubbles Layering inconsistency affecting chromatic checks.
This PR seperates Points and Bubbles into separate layers, ensuring Points layers on top of bubbles 

Video:
1. Multiple storybook refreshes showing points layered above bubbles
2. 2 browser refreshes showing points layered above bubbles

https://github.com/user-attachments/assets/39435e78-941e-408c-b064-09f38553e45a

--------------------------------------------------------------------------------------------------------------------------------------------

_**UPDATE**_

**-Example**
The order of layering from bottom - top = [bubbles (yellow), points (green), points(red)]

	`<BaseMap title="My Map" height="300">
		<Bubbles {...$$restProps} color="yellow"/>
		<Points {...$$restProps} color="green" />
		<Points {...$$restProps} color="red" />
	</BaseMap>`
	
**-How it Works**

1. We first register the components in the order they are  written in an `array`
2. We create seperate panes for each component and add their points/bubbles to their pane
3. We give those Panes a z-index based on their order in the `array`

**-Changes**

1. Removed paneType and z prop
2. No longer handle points over bubbles (up to user design)
3. Panes solve onClick layering issue

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
+++ b/packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
@@ -6,6 +6,7 @@ import { initSmoothZoom } from './LeafletSmoothZoom';
 import { writable, derived, readonly } from 'svelte/store';
 import chroma from 'chroma-js';
 import { uiColours } from '@evidence-dev/component-utilities/colours';
+import { toNumber } from '../../../utils.js';
 
 /** @template T @typedef {import('svelte/store').Writable<T>} Writable<T> */
 /** @template T @typedef {import('svelte/store').Readable<T>} Readable<T> */
@@ -239,10 +240,6 @@ export class EvidenceMap {
 		return geoJsonLayer;
 	}
 
-	#bubbleLayer;
-
-	#pointsLayer;
-
 	/**
 	 * Adds a circle marker to the map.
 	 * @param {object} item - The data item associated with the marker.
@@ -277,7 +274,11 @@ export class EvidenceMap {
 			} else if (circleOptions.pane === 'points' && circleOptions.z === undefined) {
 				this.#map.getPane('points').style.zIndex = 401; // Higher zIndex for points
 			} else if (circleOptions.z !== undefined) {
-				this.#map.getPane(circleOptions.pane).style.zIndex = 400 + circleOptions.z;
+				if (toNumber(circleOptions.z)) {
+					this.#map.getPane(circleOptions.pane).style.zIndex = 400 + toNumber(circleOptions.z);
+				} else {
+					this.#map.getPane(circleOptions.pane).style.zIndex = 400;
+				}
 			}
 		}
 
--- a/packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
+++ b/packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
@@ -270,17 +270,25 @@ export class EvidenceMap {
 		if (!this.#map.getPane(circleOptions.pane)) {
 			this.#map.createPane(circleOptions.pane);
 
-			if (circleOptions.pane.includes('bubbles') && circleOptions.z === undefined) {
-				this.#map.getPane(circleOptions.pane).style.zIndex = 400; // Lower zIndex for bubbles
-			} else if (circleOptions.pane.includes('points') && circleOptions.z === undefined) {
-				this.#map.getPane(circleOptions.pane).style.zIndex = 401; // Higher zIndex for points
-			} else if (circleOptions.z !== undefined) {
-				if (toNumber(circleOptions.z)) {
-					this.#map.getPane(circleOptions.pane).style.zIndex = 400 + toNumber(circleOptions.z);
-				} else {
-					this.#map.getPane(circleOptions.pane).style.zIndex = 400;
+			console.log(this.#paneArray, circleOptions.pane);
+
+			this.#paneArray.forEach((pane, index) => {
+				if (pane === circleOptions.pane) {
+					this.#map.getPane(pane).style.zIndex = 400 + index;
 				}
-			}
+			});
+
+			// if (circleOptions.pane.includes('bubbles') && circleOptions.z === undefined) {
+			// 	this.#map.getPane(circleOptions.pane).style.zIndex = 400; // Lower zIndex for bubbles
+			// } else if (circleOptions.pane.includes('points') && circleOptions.z === undefined) {
+			// 	this.#map.getPane(circleOptions.pane).style.zIndex = 401; // Higher zIndex for points
+			// } else if (circleOptions.z !== undefined) {
+			// 	if (toNumber(circleOptions.z)) {
+			// 		this.#map.getPane(circleOptions.pane).style.zIndex = 400 + toNumber(circleOptions.z);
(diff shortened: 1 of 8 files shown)


zachstence (commented):
What if we have multiple `<Points />` components and no `<Bubbles />` components, how do we guarantee correct rendering order?

kwongz:
@zachstence looking into this now

kwongz:
1. We are creating new panes depending on bubbles or points, to ensure the points are layered on top of bubbles.

image

2. Users can use the new paneType and z prop to control multiple layers of points or bubbles

image

Note:
If users have 2 points in a basemap without panteType and z, the map layers the points pseudo-randomly (I believe whichever point renders first). We can try to automate this for this for the user.

hughess:
@kwongz whats the syntax for pane type and z? And what would the user be specifying with those?

kwongz:
fixes #2560

kwongz:
@hughess 

Syntax: (layers blue points on top of red)
`<Points ... paneType="blue points" z='2'/>`
<Points ... paneType="red points" z='1'/>

Docs: panetype accepts strings (spaces included), z accepts numbers

image

**additional:**
Solves layering click issue, where clicking blocks other points/bubbles/areas

Current Issues:


https://github.com/user-attachments/assets/f2207fb3-3544-42ad-bc7f-88aa26ea4648


This PR Solves:


https://github.com/user-attachments/assets/1059c71d-17a5-4ed7-a901-fac3240a5b3f

kwongz:
@hughess I would also suggest removing the double-click zoom on areas points and bubbles.

Select and deselect can invoke double-click zoom, which is a minor annoyance. I can imagine  users': 

1. selecting + deselecting
2. map zooms in
3. The user needs to zoom out to access other areas of the map
4. OR user just keeps zooming deeper and deeper while interacting with the map

I think bubbles and points should not be too much of a problem, but since areas can cover a large portion of the map, this would mean user would have to double-click outside of any area layers (e.g. bodies of water, non-area layers on the map etc). Users can always use the scroll to zoom if needed.

hughess:
Is there any way for the component to automatically choose which layer should be on top? I'm thinking the easiest thing from a user perspective would be ordering the nested components by the order they want them to appear. E.g.:

```html
<!-- Points appear on top because they are first in the order below -->
<BaseMap>
   <Points/>
   <Bubbles/>
</BaseMap>
```

kwongz:
@hughess 

Ah, i understand, sorry to clarify, 

1. Currently without z or paneType, points will layer on top of bubbles. By default we check the chart type, and if it's points, it's assigned to the points pane (z-index 401), and if it's bubbles it on the bubbles pane (z-index 400).

2. the reasoning for this is regardless of how they are structured, leaflet would rendered which ever loaded the data first (I believe). resulting in storybook, and docs pages pseudo-randomly layer points or bubbles on top of each other.

3. the reasoning for paneType prop and z is for @zachstence situation where users may want to differentiate and layering of multiple points or multiple bubbles

4. The separation of panes also solves the onClick blocking issue, if we render on the same pane and depend on the ordering, we break the ordering and cause a layering issue

kwongz:
@hughess 

Spoke to Zach about a new handling for layers. 

1. Default Points layer on top of bubbles
2. automatically handle paneType for every point/bubble component used in basemaps
3. removed paneType prop
4. Use `z` to handle multiple points and bubbles


**Default points on top of bubbles**

`<Bubbles
data={la_locations}
lat="lat"
long="long"
pointName="point_name"
value="sales"
size="sales"
tooltipType="hover"
opacity="0.6"
/>`

`<Points data={la_locations} lat="lat" long="long" color="red" />`

image


**Red Points will always layer on top**

`<Points data={la_locations} lat="lat" long="long" color="red" z="3" />`
`<Points data={la_locations} lat="lat" long="long" color="purple" z="2" />`
`<Points data={la_locations} lat="lat" long="long" color="blue" z="1" />`

image

hughess:
Is there any way to detect the ordering of the child components, or do we need the user to tell us?

kwongz:
fixes #2613

kwongz:
`Is there any way to detect the ordering of the child components, or do we need the user to tell us?`

I'll look into this!

kwongz:
@zachstence  @hughess 

Bubble and Point layering now respect order user defines them in Basemap automatically.

**-Example**
The order of layering from bottom - top = [bubbles (yellow), points (green), points(red)]

	`<BaseMap title="My Map" height="300">
		<Bubbles {...$$restProps} color="yellow"/>
		<Points {...$$restProps} color="green" />
		<Points {...$$restProps} color="red" />
	</BaseMap>`
	
**-How it Works**

1. We first register the components in the order they are  written in an `array`
2. We create seperate panes for each component and add their points/bubbles to their pane
3. We give those Panes a z-index based on their order in the `array`

**-Changes**

1. Removed paneType and z prop
2. No longer handle points over bubbles (up to user design)
3. Panes solve onClick layering issue

hughess:
This is perfect!

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
```diff
@@ -510,7 +526,26 @@ export class EvidenceMap {
...
+		// Continue appending "-suffix" until a unique paneType is found
+		while (this.#paneArray.includes(newPaneType)) {
+			newPaneType = `${paneType}-${suffix}`;
+			suffix += 1;
+		}
+
+		// Add the unique paneType to the array
+		this.#paneArray.push(newPaneType);
+		return newPaneType;
 	}
```
zachstence, line 550: Is `paneType` used for anything? Can we get by with just the `suffix` - an integer that increments for each pane? Or maybe a `nanoid`?

sites/docs/pages/components/base-map.md
zachstence, line 1: Remove the docs for `paneType` and `z` since you removed those props

zachstence (changes requested):

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
kwongz: Yeah nanoid should work,

I did enjoy using the paneType to easily identify the DOM rendering pattern

image

If i were to come back to it, it would be easier for me to know which pane layer was which with the naming convention

sites/docs/pages/components/base-map.md
kwongz: will do

kwongz:
removed props from docs

packages/ui/core-components/src/lib/unsorted/viz/map/EvidenceMap.js
zachstence: Generating IDs in a while loop is generally not a great idea. If for some reason the logic broke (e.g. by a future change) it would cause an infinite loop and would lock everything.

We can keep pane type for debugging, but I'd prefer to use `nanoid` to ensure uniqueness instead of the loop.

kwongz: Sounds good, will change to nanoid!

zachstence (approved):

PR #2628, opened 2024-10-08
Use theme color tokens

zachstence:
### Checklist

~~- For UI or styling changes, I have added a screenshot or gif showing before & after~~
_Too many changes to screenshot, open the docs/example-project/test-env to check out the theming!_


packages/ui/tailwind/src/themes/defaultThemes.js
zachstence, line 1: This is where the default theme colors are defined, which are the colors shown in the screenshots in this PR's description

archiewood:
Overall coming on great

## Components

- `Alerts`: Note that if you use markdown inside an alert, it colours the text `p.markdown` colours. 
  ```
  <Alert status=info>
  
  ## Heading
  
  Para
  
  </Alert>
  ```

- `Delta`s seem a bit bright in Light mode, they were more muted before
  - Same for `BigValue` green text
  - Same for Datatable deltas
- `DataTable` redNegatives are too dark in dark mode

packages/lib/sdk/src/configuration/getEvidenceConfig.legacy.js
```diff
@@ -11,7 +11,8 @@ export const getEvidenceConfigLegacy = () => {
 		const filepath = path.join(projectRoot, 'evidence.plugins.yaml');
 		fileContents = fs.readFileSync(filepath, 'utf-8');
 	} catch (e) {
-		throw new EvidenceError('Could not find an evidence.plugins.yaml file.');
+		console.error('Could not find an evidence.plugins.yaml file.');
```
ItsMeBrianD, line 14: Any specific reason this doesn't throw anymore? We moved over from having a default to throwing when migrating to the sdk

zachstence: Reverted [b1eb047](https://github.com/evidence-dev/evidence/pull/2628/commits/b1eb0474c507215036347d8d03c9cf3e5bbd9f5f)

mcrascal:
I find the “up” green in the big value comparison and elsewhere too light in the light mode. 

I find the various alert statuses not quite right in dark mode — I think it’d be worth exploring some transparency on their backgrounds in dark.

archiewood:
@mcrascal FYI the screenshots are outdated, both of those have since been adjusted

mcrascal:
Ah sorry, trying to do it from my phone.

zachstence:
For reviewers: In addition to checking out the docs deployment to review theming, I have some questions...

1. In addition to our semantic colors, we have `primary`, `secondary`, and `accent` a concept I borrowed from [DaisyUI](https://daisyui.com/docs/colors/). Currently, `secondary` isn't used anywhere, and `accent` is only used for Map Area/Point/Bubble selected state. Anybody have ideas for more places where those colors could be used appropriately?
2. There may be places where the `info` color was used but instead should be `primary`. I had a tough time determining when to use what (especially since they're both a blueish shade)

hughess:
This is awesome! I've done a scan through the docs and have some notes on the styles in there which I'll post here, and will play around with the actual config separately!

## Style Notes

In the docs preview I see a flash of a black underline on the page header:
CleanShot 2024-10-18 at 14 26 41


### Code Block
- I find the background color too dark - makes it a bit hard to read the text, and makes the copy/paste button seem too bright

### Details component
- I prefer the lighter grey triangle icon rather than the black one - looks too much like stock HTML

### Link Button
- Border darkens a bit too much when hovered

### Alerts
- Background color and border too dark vs. text
- Links get washed out on blue background (https://docs-git-feat-2499-theming-evidence.vercel.app/components/mixed-type-charts/)
- The warning alert looks semi-brown. I think to get darker yellow you should shift into deeper orange color to avoid the color getting greyed out

### Modal
- Too dark, stands out too much on page. Think this should probably match the styling of the inputs like dropdown/daterange/etc, but that's an actual component change
   - I would put Tabs, LinkButton and BigLink in this category as well - should be revised to match styling of input components including font/size/shadows/etc

### Tabs
- Colors are too saturated
- Re: revising this component - idea for style: the "preview" and "code" tabs here: https://ui.shadcn.com/docs/components/accordion)

### Text Input
- Focus outline is too dark when clicked

### Blockquote
- Left edge looks too dark

### DataTable
- Pagination background color makes it look inaccessible (looks intentionally greyed out, so I assume I can't click it)

zachstence:
Note for myself: Need to theme the dev mode splash loading screen

mcrascal:
Very excellent. 

Dark mode: 
* I like the treatment of the alerts in dark mode, think they've regressed in light mode. 
* Fullscreen tables render incorrectly in dark mode 
* I'd consider taking the overall background one tick closer to black 
* I find the link blue slightly washed out 
* I think we're a little low contrast on the axis labels in charts. I think these might be a bit low contrast on light mode as well

General: 
* Are you going to offer the first version of dark mode/theming as an option users can opt-into?

zachstence:
> In the docs preview I see a flash of a black underline on the page header:

Looking into this first. Seems like its because the HTML from the server doesn't have `data-theme`, so we're getting a bit of FOUC (flash of unstyled content)

> Note for myself: Need to theme the dev mode splash loading screen

Whatever solution I come up with here will also help with this

archiewood (commented):
This is so good.

I've taken a spin through the docs. Here are some observations

- There appears to be a chart bug when you toggle the Appearance - Gridlines are too dark in light mode, and there is an issue with subtitle backgrounds, amongst other things
CleanShot 2024-10-21 at 12 49 01
- copy button on code block needs to match the background as the code block
CleanShot 2024-10-21 at 12 40 04@2x
CleanShot 2024-10-21 at 12 43 22@2x
- algolia search box border contrast is too low in dark mode 
CleanShot 2024-10-21 at 12 46 41@2x
- Modal and link button styling are not consistent
CleanShot 2024-10-21 at 12 52 25@2x
CleanShot 2024-10-21 at 12 52 46@2x
- DataTable colorscale prop in dark mode starts from white point 
CleanShot 2024-10-21 at 12 56 54@2x


Note for us rather than for you:
- There are places in the docs where we use screenshots or gifs that look weird in dark mode. We should avoid or eliminate these where possible
CleanShot 2024-10-21 at 12 55 06@2x

packages/ui/core-components/src/lib/organisms/layout/BreadCrumbs.svelte
```diff
@@ -58,7 +58,7 @@
 </script>
 
 <div class="flex items-start mt-0 whitespace-nowrap overflow-auto">
-	<div class="inline-flex items-center text-sm capitalize gap-1 text-gray-500 mb-2 sm:mb-4">
+	<div class="inline-flex items-center text-sm capitalize gap-1 text-base-content/50 mb-2 sm:mb-4">
```
mcrascal, line 61: `text-base-content/50` is a smell to me we've got this ~35 times, shouldn't we have something semantic, e.g. muted-content, muted-background?

zachstence: Agree 100%

sites/docs/pages/components/bar-chart.md
```diff
@@ -8,6 +8,8 @@ queries:
 - categories_by_channel.sql
 ---
 
+<BarChart data={[]} emptySet="pass" />
```
archiewood, line 11: This is presumeably just for testing - should be deleted

archiewood:
in dark mode the map outline colours appear too bright - should be a darker grey?

CleanShot 2024-10-23 at 14 49 54@2x

CleanShot 2024-10-23 at 14 51 05@2x

sites/docs/pages/components/bar-chart.md
zachstence: [852f0a9](https://github.com/evidence-dev/evidence/pull/2628/commits/852f0a93967e126ade2c08fa48e83b1538263e6e)

hughess:
This is great, so close!

Some tweak suggestions:
- US Map lines look too faint in both light and dark mode
   CleanShot 2024-11-27 at 12 50 48@2x
   CleanShot 2024-11-27 at 12 50 41@2x
- Custom color palette for heat looks unexpected in dark mode:
   CleanShot 2024-11-27 at 12 48 47@2x
- Dimension grid bar colors may be a bit too light in dark mode?
   CleanShot 2024-11-27 at 12 51 34@2x
- The warning color for reference area still looks slightly off to me - I think it needs to be more yellow. I'm not opposed to this color if needed, just looks a bit too organge/grey to my eyes
   CleanShot 2024-11-27 at 12 51 59@2x

zachstence:
@archiewood @hughess Thanks for the feedback! I've fixed all of your comments

hughess:
Last thing I just noticed - the gridlines in light mode are a bit too dark. In dark mode they're perfect

hughess:
Sorry one more thing - the colour palette on the area map for heat looks slightly off. Looks like same issue as the heatmap had before: 
https://docs-1ndodx904-evidence.vercel.app/components/area-map/#custom-color-palette

mcrascal:
This is looking really good. I tried out making a theme, and it's fantastic how far you can get with a new base + primary colour, and a couple tweaks to the chart palette. 

Some feedback from that process: 

1. If default appearance is set, and the appearance switcher is off, we should force dev mode onto that appearance on server start. Right now if I’m in dark mode, then I set the `defaultAppearance` to light, and remove the appearance switcher, my project will be stuck in dark mode. 
2. Needs Docs -- let us know if you want a hand w/ that.  
3. IMO The chart palette should be in the evidenceConfig for the template. It's going to be one of the first things folks want to change, and it's a little fiddly to do the array in yaml.  

Later improvements: 
1. We should hot reload this if it's possible, or even auto re-boot the dev server. Manually restarting the dev server is tough. 
2. UI for configuring this?

hughess:
I've been playing around with this and it is so nice!

I noticed there appears to be a bug with tooltips for scatterplot and bubble charts. Getting errors in the browser console when I hover over points, and tooltips not appearing.

hughess:
Sorry for last minute notes. 2 other areas I'm unsure of:

- Map legends. In dark mode it looks good when the legend color matches background, but I find with this example in light mode it doesn't look right:
CleanShot 2024-12-02 at 12 25 53@2x

As examples, looking at FT and Economist charts/maps, seems they typically use white for those elements (at least in light mode):
CleanShot 2024-12-02 at 12 25 19@2x
CleanShot 2024-12-02 at 12 22 01@2x

- I have the same question about chart tooltips, but that one I don't mind as much - they just seem to blend into the background a lot when hovered:
CleanShot 2024-12-02 at 12 26 08@2x

@mcrascal what do you think about these ones?

zachstence:
> I've been playing around with this and it is so nice!
> 
> I noticed there appears to be a bug with tooltips for scatterplot and bubble charts. Getting errors in the browser console when I hover over points, and tooltips not appearing.

@hughess I just pushed a fix for this

mcrascal:
@hughess agreed they don't look right.

Couple ideas: 

1. Explicitly code the ones in viz and maps white 
2. Expose a token for popover surfaces and apply it everywhere (input popovers are also filled with the background colour right now, though I think they look good). Chroma this one towards white by default

zachstence:
@hughess @mcrascal 

I think forcing map legends and tooltips to be white in light mode is the easiest solution for now. We can expose more tokens in a future release for further customization.

Should I force them to be black in dark mode? I prefer as much consistency as possible between the two appearances

archiewood (commented):
On load: 
- Rule under header very heavy
- Logo splash appeared

UI
- Rule under header too faint (light + dark)
- Body text too dark (light mode)

Keyboard shortcut
- Firefox works
- Safari works (but competes with bookmark)
- Chrome does not work
- Arc works

config:
- `themes:` -> `theme:` ? (only one theme allowed?)
- hard to discover names for things
- Zod errors hard to debug
   - if you misspell a key, eg `colorPalette` not `colorPalettes` you dont get any warning
- lack of HMR makes debugging very slow

hughess:
Last couple things - these are minor and I think can be tackled later if needed:

1. Line chart markers are white inside when `emptyCircle` is used:
CleanShot 2024-12-03 at 11 12 41@2x


2. Data labels in charts have a white outline in light mode / black in dark mode. Ideal would be same as background color, but not a big issue

CleanShot 2024-12-03 at 11 12 52@2x

zachstence:
@archiewood 

> On load:
> 
> * Rule under header very heavy
> * Logo splash appeared

Archie said this only happens once when he changed branches - this indicates its due to cached artifacts from the previous branch.

> UI
> 
> * Rule under header too faint (light + dark)
> * Body text too dark (light mode)

Fixed rule under header: [4e6874e](https://github.com/evidence-dev/evidence/pull/2628/commits/4e6874e953409dc004fdde0c12fe6ebe2aa8f896)
Fixed body text: [5fba3c4](https://github.com/evidence-dev/evidence/pull/2628/commits/5fba3c4360079d716a12e7be4fab5cf7e076cdad)

> Keyboard shortcut
> 
> * Firefox works
> * Safari works (but competes with bookmark)
> * Chrome does not work
> * Arc works

Debugged with Archie: the keyboard shortcut was broken in Chrome because it conflicted with the 1Password extension (this happened to me also). Removing the conflicting 1Password shortcut fixes the shortcut in Evidence.

> config:
> 
> * `themes:` -> `theme:` ? (only one theme allowed?)
> * hard to discover names for things
> * Zod errors hard to debug
>   * if you misspell a key, eg `colorPalette` not `colorPalettes` you dont get any warning
> * lack of HMR makes debugging slow

`themes:` -> `theme:`: [1050674](https://github.com/evidence-dev/evidence/pull/2628/commits/10506749e551883aaba10e20900b0900f8a94dc9)
Don't allow unknown keys under theme: [fad1333](https://github.com/evidence-dev/evidence/pull/2628/commits/fad13335373058ad13a7df7dc15de2a56406813e)

@archiewood will discuss with @mcrascal if HMR is necessary for initial release

mcrascal:
HMR would be nice, auto re-boot of the server on config change would also make sense to me. I don't think required for the initial release.

zachstence:
@hughess 

> Last couple things - these are minor and I think can be tackled later if needed:
> 
> 1. Line chart markers are white inside when `emptyCircle` is used:

Unfortunately its not possible to style the inside of `emptyCircle`: https://github.com/apache/echarts/issues/19386

> 2. Data labels in charts have a white outline in light mode / black in dark mode. Ideal would be same as background color, but not a big issue
> 

Fixed [916bff5](https://github.com/evidence-dev/evidence/pull/2628/commits/916bff5098c0b59efc4da5f99329f3e4ca5bd88f)

zachstence:
@archiewood 

> Exporting images in dark mode:
> 
> We should either:
> 
> * make the whole thing transparent (but I think there are some annoying label backgrounds that make this not work?
> * use the background from dark mode in the export

Fixed by setting image background color [2ddbe30](https://github.com/evidence-dev/evidence/pull/2628/commits/2ddbe30ea94a009a219fe42d05828ce35126db0c)

zachstence:
I think this library could help with config file HMR: https://www.npmjs.com/package/vite-plugin-virtual

I'll look into it if I have time

hughess:
@zachstence FYI in the most recent docs deploy preview I'm seeing errors on the Scatter Plot page

zachstence:
> @zachstence FYI in the most recent docs deploy preview I'm seeing errors on the Scatter Plot page

Fixed here [81e70ac](https://github.com/evidence-dev/evidence/pull/2628/commits/81e70ac368914cd711e8411822d2dfd4004ba355)

ItsMeBrianD (approved):

PR #2630, opened 2024-10-08
Add e2e prerender-related tests

### LINKED ISSUE #2581: e2e tests for validating prerendering/lack of double loading
Add e2e tests that make sure of a few build-time behaviours:
- data for all queries should be there (inputs, page queries, component-issued)
    - components/inputs should be prerendered with values (minus charts which can't be ssr'd)
- there hould be no double loading in charts
    - check if https://echarts.apache.org/en/api.html#events.rendered has proper behaviour? good option if so
- check that everything has rendered (charts, datatable, etc) by the time the duckdb-wasm package starts loading
    - figure out how to intercept/detect network request with jest

csjh:
### Description

Closes #2581

Adds tests for chart double loading, dropdown prerendering, `initialData` usage in page/input/manual queries, etc

a lot more can be added (i.e. one for each input, bigvalue test, etc) but I think this is a good start

### Checklist


e2e/prerender/playwright.config.js
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from '@playwright/test';
+import { config } from '../playwright-config';
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+export default {
+	...defineConfig(config),
+	webServer: {
+		command: 'pnpm sources --changed && pnpm build && pnpm preview',
```
zachstence, line 8: Is there a reason you needed to customize the behavior here?

`sources`, `build`, and `preview` are already included in the E2E action here:
https://github.com/evidence-dev/evidence/blob/812ea7506c4585366176beac39f5e2679de135a3/.github/workflows/e2e.yml#L68

The base playwright config also has `port`, `reuseExistingServer`, and `timeout` specified too
https://github.com/evidence-dev/evidence/blob/812ea7506c4585366176beac39f5e2679de135a3/e2e/playwright-config.js

---

If you made this change to only run `preview` mode tests (no `dev` mode tests), instead you should do something similar to what I did in `e2e/hmr`, which only runs tests in `dev` mode:
https://github.com/evidence-dev/evidence/blob/812ea7506c4585366176beac39f5e2679de135a3/e2e/hmr/package.json#L11

e2e/prerender/tests/tests.spec.js
```diff
@@ -0,0 +1,121 @@
...
+
+/** @param {import("@playwright/test").BrowserContext} context */
+async function waitForWasm(context) {
+	return new Promise((res) => {
+		context.route(/duckdb-eh\..*?\.wasm$/, async (route) => {
+			await route.abort();
+			res(undefined);
+		});
+	});
+}
```
zachstence, line 12: This should go in `e2e/test-utils.js` and accept `page` as an argument (not `context`) to keep consistency with the other existing util function

packages/lib/component-utilities/src/echarts.js
```diff
@@ -11,7 +11,7 @@ import * as chartWindowDebug from './chartWindowDebug';
  * } ActionParams
  */
 
-const ANIMATION_DURATION = 500;
+const ANIMATION_DURATION = navigator.webdriver ? 0 : 500;
```
zachstence, line 14: Is this disabling animations in tests? If so, I don't think we should make this change. Testing in general is most effective when the code we're testing matches the code users are running. Adding differences like this reduces the effectiveness of testing.

If you added this to avoid adding delays to the tests, instead you should look for another way to know that the chart is fully mounted/animated (like a callback/flag) if that is important for the tests

packages/ui/core-components/src/lib/atoms/inputs/dropdown/Dropdown.svelte
```diff
@@ -209,7 +209,7 @@
 {/each}
 
 <HiddenInPrint enabled={hideDuringPrint}>
-	<div class="mt-2 mb-4 ml-0 mr-2 inline-block">
+	<div class="mt-2 mb-4 ml-0 mr-2 inline-block" data-testid="Dropdown-{name}">
```
zachstence, line 212: Instead of adding a `data-testid` here, we should add an `aria-label` to the dropdown trigger (`<Button />`) so that it is accessible

```jsx
<Button
  role="combobox"
  aria-label={title ?? formatTitle(name)}
/>
```

Then we can query this dropdown like this:
```js
page.getByRole('combobox', { name: '<dropdown title or name>' })
```

packages/ui/core-components/src/lib/unsorted/viz/table/_DataTable.svelte
```diff
@@ -479,6 +479,7 @@
 	{/each}
 
 	<div
+		data-testid={isFullPage ? undefined : `DataTable-${data?.id ?? 'no-id'}`}
```
zachstence, line 482: Since there's no better way to query for a DataTable (no title or label), I'm okay with using `data-testid` here

zachstence (changes requested):

packages/lib/component-utilities/src/echarts.js
csjh: I did that so I could use the `finished` event to figure out chart rerenders - there's  a `rendered` event I'm trying but it seems like it might not be very consistent

ItsMeBrianD:
@zachstence to check on why e2e tests aren't running here

zachstence:
> @zachstence to check on why e2e tests aren't running here

Seems fixed by resolving merge conflicts

e2e/prerender/tests/tests.spec.js
```diff
@@ -0,0 +1,117 @@
...
+
+	const received = await page.evaluate(async () => {
+		await new Promise((resolve) => setTimeout(resolve, 3000));
+
+		return window[Symbol.for('chart renders')];
+	});
+
+	// 3 renders is what is expected (creation -> update -> (deletion) -> creation)
+	// double loading makes 4 renders, adding another delete/create in
+	// TODO: why is 3 renders expected
```
zachstence, line 113: Do we need this TODO?

zachstence (approved):

e2e/prerender/tests/tests.spec.js
csjh: I think so, we should ideally (all) have a very firm grasp of what queries go through on initial page load, since it's both the most important part of the rendering experience + easy to spawn bugs within

csjh: In this case, I don't really know why the charts are created twice - might be intended, but should probably be documented somewhere

PR #2665, opened 2024-10-21
Remove columns on destroy

### LINKED ISSUE #2664: [Bug]: DataTable does not respect if-else logic
### Describe the bug

When implementing a logic where user should be able to filter `DataTable` columns based on 2 separate `Dropdown`  values for `firstPeriod` and `secondPeriod` as comparison for arbitrary periods, each with min, max, mean values in separate columns, by `{#if inputs.period1.value == 1} <Column id=1_min><Column id=1_max>.... {:else if inputs.period1.value == 2} <Column id-2_min><Column id=2_max>... {/if} ` etc to catch first period, and then another if-else for the period to compare with, periods' changes in the `Dropdown` adds the columns instead of replacing them in the `DataTable`, disrespecting the if-else if logic.

### Steps to Reproduce

As described above.

### Logs

_No response_

### System Info

_No response_

### Severity

annoyance

_No response_

csjh:
### Description

Fixes #2664

Adds an `onDestroy` callback in `Column.svelte` which removes columns when their corresponding component is destroyed

### Checklist


packages/ui/core-components/src/lib/unsorted/viz/table/DataTable.stories.svelte
```diff
@@ -179,3 +181,27 @@
 	)}
 	<DataTable {data} link="link" showLinkCol />
 </Story>
+
+<Story name="Conditional Columns">
```
zachstence, line 185: @csjh Please add a `play` function with some assertions to test this functionality

zachstence (changes requested):

zachstence (approved):

PR #2667, opened 2024-10-22
Add seriesOrder parameter (fixes #2663)

### LINKED ISSUE #2663: [Bug]: Series order unpredictable in AreaChart and others
### Describe the bug

When there's more than one series in an `AreaChart` (and other types, e.g. LineCharts), the order of series depends on the data in (depending on the use-case) non-intuitive and hard to predict ways where playing with `order by` or `sort=true/false` doesn't result in a robust solution. 

For some kinds of visualization we care about the order the series are represented, either because they imply a certain meaning or because we use more sophisticated echarts features that rely on the order of series for their logic. 

### Steps to Reproduce

As an example, I was trying to build a chart where one series is subtracted from the other: 

````
```sql data
select 10000 as amount,  '2022-01-01'::date as month, 'a' as type   union
select 20000 as amount,  '2022-02-01'::date as month, 'a' as type   union
select 15000 as amount,  '2022-03-01'::date as month, 'a' as type   union
select  8000 as amount,  '2022-04-01'::date as month, 'a' as type   union
select  7000 as amount,  '2022-05-01'::date as month, 'a' as type   union
select 12000 as amount,  '2022-06-01'::date as month, 'a' as type   union
select 10000 as amount,  '2022-07-01'::date as month, 'a' as type   union
                                                                         
select -8000 as amount,  '2022-01-01'::date as month, 'b' as type union
select -10000 as amount, '2022-02-01'::date as month, 'b' as type union
select -12000 as amount, '2022-03-01'::date as month, 'b' as type union
select -11000 as amount, '2022-04-01'::date as month, 'b' as type union
select -10000 as amount, '2022-05-01'::date as month, 'b' as type union
select -15000 as amount, '2022-06-01'::date as month, 'b' as type union
select -20000 as amount, '2022-07-01'::date as month, 'b' as type 
```

```sql ordered_data
select * from ${data}
         order by type
```

<AreaChart
data={ordered_data}
x=month
y=amount
series=type
seriesOptions={{
'stackStrategy': 'positive'
}}
seriesColors={{
'a': 'green',
'b': 'red'
}}
sort=false
/>
````

ordered_data_2024-10-20T17-53-56

Now drop the first data point and the order of series has flipped, breaking the whole visualisation
ordered_data_2024-10-20T17-55-10

### Logs

_No response_

### System Info

_No response_

### Severity

annoyance

Digging around on Slack a bit, the introduction of a seriesOrder configuration has come up a few times but as far as I can see hasn't been implemented so far. 
I've prototyped this for myself (duplicating the AreaChart component and sorting the `seriesConfig` according to a specific `seriesOrder`) and it seems to be working fine. 

I'm happy to work on a PR (maybe building this into the `getSeriesConfig` function so the behavior is consistent for other types of chart) if that's the way we want to go with this.

flosell:
### Description

Attempt to fix #2663 - let me know what you think 

### Checklist

- [ ] ~For UI or styling changes, I have added a screenshot or gif showing before & after~
- [ ] ~I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable~

### DIFF
--- a/.changeset/tasty-parents-punch.md
+++ b/.changeset/tasty-parents-punch.md
@@ -1,0 +1,6 @@
+---
+'@evidence-dev/component-utilities': patch
+'@evidence-dev/core-components': patch
+---
+
+Add seriesOrder parameter to allow users to specify a specific order the series are displayed in (fixes #2663)
--- a/packages/lib/component-utilities/src/getSeriesConfig.js
+++ b/packages/lib/component-utilities/src/getSeriesConfig.js
@@ -10,6 +10,7 @@ export default function getSeriesConfig(
 	name,
 	xMismatch, // this checks for scenarios where xType is string and xDataType is number. When this is the case, we need to inject strings into the x axis, or else it will cause echarts to think there are duplicate x-axis values (e.g., "4" and 4)
 	columnSummary,
+	seriesOrder,
 	size = undefined,
 	tooltipTitle = undefined,
 	y2 = undefined
@@ -216,6 +217,10 @@ export default function getSeriesConfig(
 
 		tempConfig = generateTempConfig(seriesData, seriesName, yAxisIndex, baseConfig);
 		seriesConfig.push(tempConfig);
+	}
+
+	if (seriesOrder) {
+		seriesConfig.sort((a, b) => seriesOrder.indexOf(a.name) - seriesOrder.indexOf(b.name));
 	}
 
 	return seriesConfig;
--- a/packages/ui/core-components/src/lib/unsorted/viz/area/Area.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/area/Area.svelte
@@ -64,7 +64,7 @@
 		labelFormat = getFormatObjectFromString(labelFmt);
 	}
 	export let showAllLabels = false;
-
+	export let seriesOrder = undefined;
 	// Prop check. If local props supplied, use those. Otherwise fall back to global props.
 	$: data = $props.data;
 	$: x = $props.x;
@@ -155,7 +155,8 @@
 		baseConfig,
 		name,
 		xMismatch,
-		columnSummary
+		columnSummary,
+		seriesOrder
 	);
 
 	$: config.update((d) => {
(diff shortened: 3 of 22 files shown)


flosell:
Saw the netlify errors above but couldn't find a reason for those to fail - the example project and docs build just fine locally (and at least the latter seem to also be fine deploying to vercel...). Any pointers? Or is this error unrelated to this PR?

archiewood:
This appears to be failing CI in node 18

hughess:
@flosell thanks for the PR! This looks great overall.

The test issues with node v18 look to be coming from `toSorted` (see here: https://stackoverflow.com/questions/76006398/tosorted-works-only-in-a-browser)

Could you swap this out for another method?

Otherwise this will be a great addition!

flosell:
Hi @hughess @archiewood, thanks for checking and sorry about that! 
Changed things to in-place `sort()`, this should fix things

hughess:
Awesome thanks @flosell !

hughess:
@flosell it looks like the series order doesn't take effect in BarChart for some reason. All the other charts I've tested have worked though

archiewood:
kicked off tests

flosell:
@hughess good catch, thanks! Accidentally passed the argument to `<Chart/>` instead of `<Bar/>`. `BubbleChart` had the same problem by the way. 
Both should be fixed now

zachstence (changes requested):
The functionality here looks good, but I'd like to see some stories written to test this. @flosell are you able to write some stories?

flosell:
Hi @zachstence! Added some stories - this is my first time working with storybook so feel free to give feedback. 

Also, if you expect folks to add stories when contributing, maybe add a few notes in CONTRIBUTING.md to save some review cycles? 

Some notes on the change: 
* On some chart types (e.g. BubbleChart, ScatterPlot), I combined seriesColor and seriesOrder to make the effect of the seriesOrder a bit more visible. Let me know if you'd rather not mix multiple attributes in the same story
* `ScatterPlot` didn't have a story file yet so I created one, inspired by BubbleChart (those two chart types seem to be very similar...) 
* I noticed chromatic doesn't seem to pick up the new stories - is that intentional or is there something I did wrong?

zachstence:
@flosell Everything looks good!

I'm trying to figure out why Chromatic isn't snapshotting the new stories...

zachstence:
I think Chromatic didn't snapshot your new stories because our GitHub action always clones the upstream `evidence-dev/evidence` repo, and not your fork. I'll look into the appropriate way to handle that.

Going to go ahead and merge this since it looks good! Thanks @flosell !

zachstence (approved):

PR #2670, opened 2024-10-23
fixed string handling for prop values

### LINKED ISSUE #2668: [Bug]: Slider should support numbers as strings
### Describe the bug

`defaultValue` and `step` props currently do not convert string inputs into numbers, causing the component to break in some scenarios. Same logic for `min` and `max` should be applied to those props

### Steps to Reproduce

Add slider with `defaultValue` and `step` specified

### Logs

_No response_

### System Info

_No response_

### Severity

annoyance

Wrap numbers with `{}`. E.g.,:

```
<Slider
   defaultValue={0.11}
   step={0.02}
/>
```

kwongz:
### Description

fix #2668 

1. Fixed: ToNumber String Values to avoid errors
2. Improved: console.error for NaN values

`<Story
	name="handle string values"
	args={{
		defaultValue: '18',
		fmt: 'usd0',
		steps: '2',
		max: '20',
		min: '0'
	}}
/>`

https://github.com/user-attachments/assets/7c02e1f7-cc94-4ff3-85a3-30facc1860ed

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
@@ -23,14 +23,14 @@
 	/** @type {string} */
 	export let name;
 
-	/** @type {number|undefined} */
+	/** @type {number} */
 	export let min = 0;
 
 	/** @type {number} */
 	export let max = 100;
 
 	/** @type {number} */
-	export let step = undefined;
+	export let step = 1;
 
 	/** @type {boolean} */
 	export let showMaxMin = true;
@@ -72,23 +72,12 @@
 
 	if (min !== undefined) {
 		min = validateNumber(min, 'min');
-		if (min !== undefined && max !== undefined) {
-			checkMinMax(min, max);
-		}
 	}
-
 	if (max !== undefined) {
 		max = validateNumber(max, 'max');
-		if (max !== undefined && min !== undefined) {
-			checkMinMax(min, max);
-		}
 	}
-
-	if (step !== undefined) {
-		step = validateNumber(step, 'step');
-		if (step !== undefined && step <= 0) {
-			console.error('step must be greater than 0');
-		}
+	if (max !== undefined && min !== undefined) {
+		checkMinMax(min, max);
 	}
 
 	$: if (defaultValue !== undefined) {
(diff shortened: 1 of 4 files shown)


packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -25,35 +26,11 @@
 	/** @type {number|undefined} */
```
zachstence, line 26: Since `min` has a default value (`0`), it will never be `undefined`, so we can remove `undefined` from the type

```suggestion
	/** @type {number} */
```

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -25,35 +26,11 @@
...
-		} else if (max < min) {
-			throw Error('max cannot be less than min');
-		}
-	} else {
-		max = 100;
-	}
-
 	/** @type {number} */
-	export let step;
+	export let step = undefined;
```
zachstence, line 33: We should default `step` to `1`

```suggestion
	export let step = 1;
```

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -72,11 +49,61 @@
...
+	}
+
+	if (max !== undefined) {
+		max = validateNumber(max, 'max');
+		if (max !== undefined && min !== undefined) {
+			checkMinMax(min, max);
+		}
+	}
+
+	if (step !== undefined) {
```
zachstence, line 87: Since we're defaulting `step` to `1`, we don't need this check either

packages/ui/core-components/src/lib/atoms/inputs/slider/Slider.svelte
```diff
@@ -72,11 +49,61 @@
...
+			checkMinMax(min, max);
+		}
+	}
+
+	if (max !== undefined) {
+		max = validateNumber(max, 'max');
+		if (max !== undefined && min !== undefined) {
+			checkMinMax(min, max);
+		}
+	}
```
zachstence, line 85: `checkMinMax` is called twice, it only needs to be called once

```suggestion
	if (min !== undefined) {
		min = validateNumber(min, 'min');
	}
	if (max !== undefined) {
		max = validateNumber(max, 'max');
	}
	if (max !== undefined && min !== undefined) {
		checkMinMax(min, max);
	}
```

zachstence (changes requested):

zachstence (approved):

Issue #2539, opened 2024-09-13
[Bug]: Slider can't accept dynamic defaultValue


Ash-Woods:
### Describe the bug

I am attempting to pass a dynamic defaultValue on a Slider so that users can adjust the value as needed from a more configurable position, rather than solely starting at 0. I am using the below code:

```svelte
<Slider
    title="Median Days" 
    name=days_slider
    min=0
    max={max_flow_time_median_days[0].max_value}
    defaultValue={max_flow_time_median_days[0].max_value}
/>
```

and it continues to default at 0.

![image (13)](https://github.com/user-attachments/assets/8d49155b-4b59-4c87-8769-deea354c2aad)

### Steps to Reproduce

Create the query, then create the Slider. Once complete, load the page in dev mode and it navigate to the rendered component on the screen.

### System Info

```Shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 185.11 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.13.1 - /usr/local/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.5.2 - /usr/local/bin/npm
    Watchman: 2024.08.26.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 128.0.6613.120
    Safari: 17.5
  npmPackages:
    @evidence-dev/bigquery: ^2.0.7 => 2.0.7 
    @evidence-dev/core-components: ^4.7.5 => 4.7.5 
    @evidence-dev/csv: ^1.0.12 => 1.0.12 
    @evidence-dev/databricks: ^1.0.7 => 1.0.7 
    @evidence-dev/duckdb: ^1.0.11 => 1.0.11 
    @evidence-dev/evidence: ^39.1.4 => 39.1.4 
    @evidence-dev/motherduck: ^1.0.2 => 1.0.2 
    @evidence-dev/mssql: ^1.0.9 => 1.0.9 
    @evidence-dev/mysql: ^1.1.3 => 1.1.3 
    @evidence-dev/postgres: ^1.0.6 => 1.0.6 
    @evidence-dev/snowflake: ^1.1.1 => 1.1.1 
    @evidence-dev/sqlite: ^2.0.6 => 2.0.6 
    @evidence-dev/trino: ^1.0.8 => 1.0.8 
[...]

PR #2672, opened 2024-10-24
Fix/2539 slider dynamic default value

kwongz:
### Description

Slider accepts Data prop, can use 

1. maxColumn to get the first row value of a column and sets slider's max value
2. minColumn to get the first row value of a column and sets slider's min value
3. defaultValue now accepts strings, gets first row of a column and set slider's default value

### Checklist


kwongz:
Had some questions on handling the dynamic values

- Would it be for the user to avoid creating new columns (`max_value`, `min_value`)? Instead accept a single `column` (sales) and if `maxColumn`/`minColumn`  prop is provided we provide the max and min values from that column?
- Currently it feels bad to be arbitrarily taking the first value of the max and min columns

ItsMeBrianD:
> Currently it feels bad to be arbitrarily taking the first value of the max and min columns

This is the way `Value` and `BigValue`, so I don't think this is an issue

> Would it be for the user to avoid creating new columns (max_value, min_value)? Instead accept a single column (sales) and if maxColumn/minColumn prop is provided we provide the max and min values from that column?

I'm not sure what the question is asking here

kwongz:
all good, not important

packages/ui/core-components/src/lib/atoms/inputs/slider/_Slider.svelte
```diff
@@ -0,0 +1,175 @@
...
+			error = e.message;
+			const setTextRed = '\x1b[31m%s\x1b[0m';
+			console.error(setTextRed, `Error in Value: ${error}`);
+			if (strictBuild) {
+				throw error;
+			}
+		}
+	}
+
+	// Query data integration
```
zachstence, line 156: What's this comment for?

zachstence, line 156: What's this comment for?

zachstence (approved):

packages/ui/core-components/src/lib/atoms/inputs/slider/_Slider.svelte
kwongz: that should not be there, will remove

zachstence: [af2d4db](https://github.com/evidence-dev/evidence/pull/2672/commits/af2d4dbf110bf709270453da950befd25bc2f5c2)

kwongz:
fixes #2539

Issue #2675, opened 2024-10-24
Better docs previews


zachstence:
Our docs previews (rendered example and code) aren't standardized across the app (different order, spacing, etc) and it makes it tricky to follow examples. We should improve this experience by implementing some simple tabs to wrap the preview and code.

ShadCN's docs do a great job of this
https://ui.shadcn.com/docs/components/button

https://github.com/user-attachments/assets/d781d4d4-762b-49d5-a5c4-151d7ed1270e

Our implementation should basically mirror ShadCN's, with a few tweaks:
- We don't need the "Style" dropdown, copy button, or "Open in V0" button on the Preview tab
- @mcrascal has requested that the tab underline slide when changing tabs
  - I implemented something similar in an early version of the [`ThemeChanger` component](https://github.com/evidence-dev/evidence/pull/2520/commits/908629b87f3459c01dce1cdd8bb0629d2c838558#diff-6649ab1eb84f0e09685218aff2ed0373b3bb8c3f6be1b239ef6b6d3094c36d32), the logic for the sliding background should help you out

This should be a custom component in our `docs`, not in `core-components`.

hughess:
I think we should consider replacing our Tabs component with whatever we land on for this

PR #2681, opened 2024-10-25
Fix/2675 docs tabs

kwongz:
### Description

https://github.com/user-attachments/assets/08f3356a-5213-461b-93fd-133ac4faf74f

**Notion Docs insturctions**
https://www.notion.so/Docs-Doc-Tabs-12a5287a4ba08046b7f1cef7f7b03eae

### Checklist

- ~~[ ] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)~~


kwongz:
fixes #2675

archiewood:
This will be great!

Couple of suggestions in loom

mcrascal:
This is cool. Just looking at it in the preview docs: 

- the vertical slide transition currently fires on navigation to the page
- text-xs is way too small for this imo 
- needs more margin top, probably less margin bottom. Right now in a page with lots of these, it's not clear what sections it's denoting. 
- Need a horizontal bottom border for the width of the page
- Possibly need something wrapping the content which is inside the tabs

sites/docs/components/DocTab.svelte
```diff
@@ -0,0 +1,93 @@
...
+		{#each tabs as tab, index}
+			<button
+				class="p-1 cursor-pointer transition-colors duration-300 text-xs font-semibold ease-in-out {activeTab ===
+				tab
+					? 'text-black'
+					: 'text-gray-600'}"
+				on:click={() => setTab(tab, index)}
+				bind:this={tabButtons[index]}
+			>
+				{toCapitilize(tab)}
```
mcrascal, line 71: Could we consider using the `capitalize` class instead of introducing a new utils function for this?

sites/docs/components/DocTab.svelte
```diff
@@ -0,0 +1,93 @@
...
+			style="width: {activeBorderStyles.width}; transform: {activeBorderStyles.transform};"
+			class="absolute bottom-0 left-0 h-[2px] bg-black transition-[width, transform] duration-300 ease-in-out"
+		></div>
+	</div>
+
+	<div
+		bind:this={tab}
+		style="height: {tabContentHeight};"
+		class="transition-height overflow-hidden ease-in-out duration-300"
+	>
```
mcrascal, line 84: Why isn't svelte `transition:slide|local` on the child divs suitable appropriate here?

mcrascal (changes requested):

sites/docs/components/DocTab.svelte
kwongz: Currently playing around with svelte transitions to see possible solutions.

Originally i chose to transition height because the default slot with the code grows in height after initially rendering on the page, causing the rest of the page to grow downwards (Giving a bad delayed jump effect on the content). This was handled smoothly by the height observers to detect and transition any changes in height of the div on tab change.

I think with the non svelte transitions we can achieve a lot more custom and fine tuned animations/transitions. But happy to explore other avenues.

kwongz:
@mcrascal 

1. Updated Accordion and Slider to DocTabs
2. Switched to svelte animations (still working on the smooth transition w/ svelte transitions

archiewood:
added a couple of suggestions on styling, feel free to revert if they feel less good.
- back to font-sans
- more padding around the preview
- adjust preview border colour and shadow to match code

archiewood:
The svelte slide animation is still too janky IMO.

CleanShot 2024-10-31 at 11 27 37

code -> preview is fine
but 
preview -> code has a weird jump at the end

archiewood:
on testing further, this seems related to the CodeBlock, not the transition. 

We can fix that separately

kwongz:
@archiewood I found a solution to the code --> preview, the prism loads in async, adding a margin after they lifecycle methods giving it the height 'jank'

By adding the exact margin as a tailwind style to the div container we avoid this height jank.

kwongz:
On Another note,

Current svelte transition unmount and mounts the preview content or code content. For pages like DataTable, where the DataTable get remounted, it can really slow down the page, as well as the transitions between the tabs.

A solution I'm looking into is doing CSS transitions and instead of unmounting the component, we can render everything and only make visible the content based on the tab chosen.

archiewood:
@kwongz the margin jank should have been fixed in #2706

kwongz:
@archiewood @mcrascal 

Added Custom CSS transitions to handle rendering issues with svelte transitions (unmounting/remounting components). Causing particular issues with datatables page.

The margin and padding layout is consistent across the preview and code block, but individual components such as Area Chart can appear to have more spacing due to margin/padding within the component itself.

I think its a pretty suddle difference that users won't notice immediately when they are consistent across components on the same doc page. But I thought we could introduce a standard margin for all components to create a more uniform look in the future.

_preview container margin_

image

_Component margin_

image

__slider spacing not good__

image

sites/docs/components/Pedistal.svelte
archiewood, line 1: is this component used

kwongz: Pedistal was our original solution that had a cleaner transition, I left it out commented to come back to at another time

archiewood (approved):

mcrascal (changes requested):
This seems close, but not quite there for something people will be using constantly in the docs. Prob worth a quick discussion. 

1. Transitioning this horizontally seems more annoying than useful. It doesn't remove the vertical jank, which I believe was the original goal of including a transition, but it adds quite a lot of time between seeing the output and the code. I think no transition between output and code would be preferable to the horizontal movement. 
2.  If we stay with the horizontal transition, you'd expect a carrousel type of movement (the preview is getting pushed off the screen by the code, and vice versa). This has the original object disappearing and then a new one sliding in. 
3. Focus rings on the buttons need styling 
4. I still find xs very small for something you're clicking all the time.

kwongz:
removed hover, added pointer, matched dropdown focus state

image

image

mcrascal (approved):

PR #2690, opened 2024-10-28
Fix transitions between chart states w/ updated data

csjh:
### Description

Broken:
Broken

Fixed:
Fixed

Note: removes safeguard Brian put for [showing the empty chart error state when all rows in a series are null](https://github.com/evidence-dev/evidence/pull/2026) - I would think this PR takes higher priority, but would be nice to recover that error indication

### Checklist

### DIFF
--- a/e2e/prerender/tests/tests.spec.js
+++ b/e2e/prerender/tests/tests.spec.js
@@ -105,13 +105,16 @@ test('charts should render once', async ({ page }) => {
 	const received = await page.evaluate(async () => {
 		await new Promise((resolve) => setTimeout(resolve, 3000));
 
-		return window[Symbol.for('chart renders')];
+		return window[Symbol.for('chart renders')];	
 	});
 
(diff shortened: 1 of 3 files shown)


e2e/prerender/tests/tests.spec.js
```diff
@@ -108,10 +108,13 @@ test('charts should render once', async ({ page }) => {
...
-	// double loading makes 4 renders, adding another delete/create in
-	// TODO: why is 3 renders expected
-	const expected = 7;
+	// 4 renders is what is expected (creation -> update x3)
+	// double loading makes 6 renders, adding another 2 updates in
+	// TODO: why are 3 updates expected
+	// - first is for initial render w/o chart props
+	// - second is for render w/ chart props
+	// - third is for ???, fourth is for ???
+	const expected = 10;
```
zachstence, line 117: This is getting smellier and smellier. What do we need to nail down what is expected here and remove these TODOs?

zachstence (approved):

PR #2697, opened 2024-10-30
Feat/dimensiongrid fmt support

mcrascal:
Adds `fmt` support to dimension grid


ItsMeBrianD (approved):

packages/ui/core-components/src/lib/molecules/dimension-grid/DimensionRow.svelte
```diff
@@ -23,14 +23,24 @@
...
+				'bg-gray-200': row.metric === undefined
+			},
+			{
+				// null can occur naturally, allow selection state to show through
+				'bg-transparent': row.metric === null
+			},
+			{
+				'bg-blue-200': row.metric && isSelected
+			}
 		)}
```
ItsMeBrianD, line 37: Could this be written with a series of `class:bg-someclass={condition}` rather than this object syntax? I'm not sure we use it elsewhere and introducing a new way to do conditional classes seems like a big step

PR #2693, opened 2024-10-28
Add connectionTimeout option to MSSQL connector

ebardelli:
### Description

This pull request adds the connectionTimeout option to the MSSQL connector.

[connectionTimeout](https://tediousjs.github.io/node-mssql/#general-same-for-all-drivers) allows for the extension of the time that a query can run. The current value is left undefined, which leads to the connector defaulting to 15000 ms. This could cause some connections to be terminated if they take longer than 15 seconds to run.

### Checklist

- [x] For UI or styling changes, I have added a screenshot or gif showing before & after
- [x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)
- [x] I have added to the docs where applicable
- [x] I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable

PR #2699, opened 2024-10-30
Add requestTimeout option to mssql connector

ebardelli:
### Description

This is a follow-up to #2693. I believe I actually need `requestTimeout` instead of `connectionTimeout`.

Let me know if you would like to only keep `requestTimeout` instead of `connectionTimeout`. I'm not sure what's the difference between the two. The upstream documentation isn't super clear.

### Checklist

### DIFF
--- a/.changeset/blue-eagles-visit.md
+++ b/.changeset/blue-eagles-visit.md
@@ -1,5 +1,0 @@
----
-'@evidence-dev/mssql': minor
----
-
-Add alternatives authentication methods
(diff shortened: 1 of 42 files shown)


archiewood:
@ebardelli - you should be able to test this locally to see if it has the intended effect.

follow the steps in https://github.com/evidence-dev/evidence/pull/2706#event-15045563654 to set up the example project.

Then:
- add a mssql connection with your settings
- see if they have the intended effect

archiewood:
okay great, if you can get it to fail by toggling the timeout lower, it seems clear it is working as intended

archiewood (approved):

PR #2707, opened 2024-10-31
Fix HMR, HMR e2e tests

csjh:
### Description

Fixes HMR & associated e2e test

Seems like a full page reload is unavoidable when a file is added/deleted, if someone could confirm whether we've had it working with non-full reload I'll look into it more

### Checklist


csjh:
There's also a weird bug where if you delete a file above your current page, after reloading, it displays the page below your true page despite no URL change. Seems independent of the `import.meta.glob` thing, was happening before

e2e/hmr/tests/tests.spec.js
```diff
@@ -33,7 +34,8 @@ test.describe('pages', () => {
 		await expect(page.getByText('Index')).toBeVisible();
 
 		createFile('pages/new-page.md', 'This is a new page');
-		await waitForHMR(page);
+		// file deletions trigger full reload
+		// await waitForHMR(page);
```
zachstence, line 38: Remove the commented out code and improve the comment to explain

```suggestion
		// file deletions trigger full reload, so we don't waitForHMR() here
```

PR #2729, opened 2024-11-05
[Themes] Implement `colorPalettes` and `colorScales` config

zachstence:
This PR implements the `colorPalettes` and `colorScales` themes config.

It also replaces `chartColours` and `mapColours` with the `default` color palette.

There are plenty of other places where we should use palettes and scales from config (e.g. `DataTable`), but I'm going to tackle those in a separate PR.


packages/ui/tailwind/src/themes/defaultThemesConfig.js
```diff
@@ -92,6 +92,64 @@ export const defaultThemesConfig = {
...
+					'hsla(195, 49%, 51%, 1)', // Teal
+					'hsla(207, 69%, 79%, 1)', // Light Blue
+					'hsla(202, 28%, 65%, 1)', // Grey
+					'hsla(179, 37%, 65%, 1)', // Light Green
+					'hsla(40, 30%, 75%, 1)', // Tan
+					'hsla(38, 89%, 62%, 1)', // Yellow
+					'hsla(342, 40%, 40%, 1)', // Maroon
+					'hsla(207, 86%, 70%, 1)', // Blue
+					'hsla(160, 40%, 46%, 1)', // Green
+					// TODO should we get rid of these gray scale colors?
```
ItsMeBrianD, line 109: todo

ItsMeBrianD (approved):
changeset?

zachstence:
> changeset?

Going into a feature branch so no need for a changeset here. Will have appropriate changesets on the final merge into next.

packages/ui/tailwind/src/themes/defaultThemesConfig.js
zachstence: @hughess Thoughts on this?

TLDR: Our `chartColours` currently includes gray scale values

https://github.com/evidence-dev/evidence/blob/1182936112a2f5b78aab4e71eccf7a6f98f2d044/packages/lib/component-utilities/src/colours.js#L56-L66

These gray scale values only get used on a chart with a lot of series. I think it would be better to just loop the colors rather than using gray scale. What do you think?

zachstence: Removed here: [9bd8a8a](https://github.com/evidence-dev/evidence/pull/2729/commits/9bd8a8a29128af2f495f0ea71fa3f2733169dbd0)

We can add more colors when we fine-tune the theme colors

PR #2730, opened 2024-11-05
Fix map loading screen layout shift

csjh:
#### Before
before-map

#### After
after-map

### Checklist

### DIFF
--- a/.changeset/cyan-sloths-train.md
+++ b/.changeset/cyan-sloths-train.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+fix map loading state layout shift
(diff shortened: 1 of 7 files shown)


hughess:
Nice! is there no margin on the maps when there is no title? Looking at the last map in the "after" gif - looks like it's right up against the previous map. Can't remember if our charts do top / bottom / both for margin

csjh:
@zachstence #2302 was the one I was wondering about, seems like the logic is the same between all the maps except `AreaMap`, which has this extra bit:
image

I don't *think* this is needed anymore?

csjh:
> Nice! is there no margin on the maps when there is no title? Looking at the last map in the "after" gif - looks like it's right up against the previous map. Can't remember if our charts do top / bottom / both for margin

Yeah seems like there's no margin by default

hughess:
While you're in the maps files, can you set up margin so it does the same thing as `<Chart/>`?

hughess:
Here’s an example where the top of the map overlaps content above: https://next-docs-6up4eq5mh-evidence.vercel.app/components/area-map/#custom-tooltip

PR #2738, opened 2024-11-06
Dynamic Base Path

ItsMeBrianD:
### Checklist


archiewood (commented):
Can we add some docs to this?

Where does a user define basePath? `evidence.plugins.yaml`?

If so I'd suggest a section in https://docs.evidence.dev/deployment/overview/

.changeset/stale-kids-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@evidence-dev/evidence': patch
+---
+
+Removed EVIDENCE_DISABLE_INCLUDE, behavior is now forced by `--debug`
```
zachstence, line 5: Would be better to leave these separate in my opinion. There are situations where we want `EVIDENCE_DISABLE_INCLUDE` without `--debug`, and vice versa.

packages/ui/core-components/src/lib/atoms/shadcn/badge/badge.svelte
```diff
@@ -16,7 +16,6 @@
 
 <svelte:element
 	this={href ? 'a' : 'span'}
-	{href}
```
zachstence, line 19: Why is this change necessary?

packages/lib/sdk/src/utils/svelte/buildUrl.js
```diff
@@ -0,0 +1,28 @@
...
+    
+    let basePath = config.deployment.basePath;
+    if (basePath) {
+        if (!basePath?.startsWith('/')) {
+            basePath = `/${basePath}`;
+        }
+        if (basePath.endsWith('/')) {
+            basePath = basePath.slice(0, -1);
+        }
+        return `${basePath}${path}`;
```
zachstence, line 24: This assumes that `path` always starts with `/`. Is this guaranteed to be true?

I think we should either
- Ensure append a `/` if `path` doesn't start with `/`, or
- Throw a descriptive error if `path` doesn't start with `/` rather than returning an incorrect URL (e.g. `/basetest` instead of `/base/test`)?

packages/lib/sdk/src/utils/svelte/buildUrl.js
```diff
@@ -0,0 +1,28 @@
...
+ * Adjusts a path to include the configured base path
+ * Ignores undefined, and absolute URLs
+ * @param {string} path 
+ * @example buildUrl('http://localhost:3000/test') // 'http://localhost:3000/test'
+ * @example buildUrl('/test') // '/base/test'
+ * @example buildUrl(undefined) // undefined
+ * 
+ * @returns 
+ */
+export const buildUrl = (path) => {
```
zachstence, line 12: I think this could be named better, maybe `addBasePath`? Something that describes what it does specifically

zachstence (changes requested):

packages/ui/core-components/src/lib/atoms/shadcn/badge/badge.svelte
ItsMeBrianD: This was in error

zachstence:
Note for myself: Need to exclude base path from breadcrumbs
image

.changeset/stale-kids-tan.md
zachstence: @ItsMeBrianD What do you think about this?

hughess:
This looks great! All working on my end

hughess (approved):

.changeset/stale-kids-tan.md
zachstence: Reverted this change: [0853ee0](https://github.com/evidence-dev/evidence/pull/2738/commits/0853ee02dbac3f3deac0d69a974be9afca59a953)

We can discuss more later if necessary

zachstence (approved):

PR #2739, opened 2024-11-06
[Themes] support custom color tokens, automatically create css vars

zachstence:


ItsMeBrianD (approved):
Changeset?

zachstence:
> Changeset?

Going into a feature branch so no need for a changeset here. Will have appropriate changesets on the final merge into next.

PR #2754, opened 2024-11-09
add explanation how to selfhost SPA

fboerman:
### Description

This PR updates the docs to provide an example nginx config for self hosting evidence in SPA mode

PR #2756, opened 2024-11-11
Auto-pass sources tests on docs only changes

archiewood:
@zachstence for your consideration


archiewood:
This would prevent approval being required to run CI on sources in eg https://github.com/evidence-dev/evidence/pull/2754

archiewood:
actually, this would not help yet as the whole test needs the external-pr approval

Issue #2711, opened 2024-11-01
[Bug]: DataTable sort icons cause ugly wrapping


hughess:
### Describe the bug

Sometimes when clicking to sort columns in DataTable, sort icon will jump to next row. This shouldn't happen

### Steps to Reproduce

See example: https://x.com/jmduke/status/1852148356161659062

### Severity

annoyance

### Additional Information, or Workarounds

A few ideas:
- Add small right gutter to all columns with enough space to put a smaller sort icon when sorted
- Alex Prejean's suggestion: What if there was a transparent graded color fill on the header cell with darker transitioning to lighter top to bottom to show descending order and inverse for ascending. Eye is quickly drawn to the column being sorted and you keep all the real estate in the column name cell
- Add the sort icon in a layer above the column header cell - when applied it appears with a semi-transparent background. If it overlaps the text, it will partially block the text but keep the cell size the same

Requirements:
- No default icon - only show when sort is applied (too cluttered otherwise)
- Eliminate cell shift/resizing
- Sort icon should never be on its own on a row

archiewood:
I would vote for gutter here

kwongz:
![Image](https://github.com/user-attachments/assets/261d5c04-96ed-4345-b70f-4b0a479d1463)

kwongz:
just for reference

![Image](https://github.com/user-attachments/assets/24c0200e-3b7c-4cb8-93a2-f3fc95b68340)

PR #2766, opened 2024-11-12
fixed table header chevron layout shift

kwongz:
### Description
First page is new datatable, second page is old datatable
https://www.loom.com/share/5608209a1dfe479f89c7dbb3475de4f1?sid=37b1a37f-9c53-4c2b-9c68-18f4da575669

### Checklist


hughess:
The layout shift looks a lot better, but unfortunately we'll need to make sure the column title and the values in the column are aligned - for example in a numeric column they should both be right aligned, so any gap you're applying to the column title should also be applied to the individual cells in the column.

That might work well if we also reduce the left/right padding in the cells, and let the new gutter use that space

kwongz:
@hughess

improved datatable header

kwongz:
improved datatable header

kwongz:
image

image

hughess:
This looks good in general, but there's a few cases where I think we need to adjust. 

Before sending over those, just a quick clarifying question - is `wrapTitles` applied in this table? https://docs-i8kx0rshh-evidence.vercel.app/components/data-table/#column-groups

kwongz:
Column Groups example i removed wraptitles from the code example. I see I forgot to remove it from the code block. 

It seems not necessary since we have wraptitle example just below

hughess:
Here's a few scenarios that look a bit off:

1. Not enough padding on left side of cells
CleanShot 2024-11-13 at 16 52 52@2x

2. Unclear which column the sort icon applies to
CleanShot 2024-11-13 at 16 51 41@2x
CleanShot 2024-11-13 at 16 49 42@2x

3. Too much padding on right side of cells - is more apparent when conditional formatting is used
CleanShot 2024-11-13 at 16 50 36@2x

I think the solution is some combo of:
- Increase padding on left side of cells
- Reduce space between title and left side of chevron icon
- (possibly) reduce size of chevron icon
- (possibly) reduce space on right of chevron

kwongz:
Sounds good ill get these out right now

kwongz:
Hey Sean made some of your changes

- cleaned up spacing between cells
- reduced size of chevron
- added this idea of a gradient background to highlight the sorted column + indicate asc or desc sort
- added margin left to chev, I thought it gets too close to header when datatable have many columns

_without the current margin left, chev will stick next to header text with many columns_
image


Table cell padding-right was 19px, reduced to 12px
_table cell spacing_
image

_other datatables_

image

_table cell centered text_
image

image

kwongz:
ah i c, to fix the wrapped titles I wrapped them in a flex box structure to get the chevron to behave properly. 

ill fix
- Align center header text
- remove linear gradient
- Do we want the chev to always be on the right in the "gutter" or do we want it to be right next to header text?

kwongz:
- Fixed the alignment issue
image

- more spacing left of chevron then right (compared to right adjacent table header text)
image


__________________________________________________________________________________________________________________________________________
-Something I dislike is alignment center, the chev is staying on the right, when the datatable has a lot of space the chev feels unconnected to that table. We Could allow alignment center chevs to be centered as well

_alignment= center: columns country ID chev feels disconnected_
image

_Solutions: center chev when alignment is center_
image

I think this is the case even when text is default aligned: left, if the column has a large width, there s a lot of empty white space in between text and chev. Maybe the chev should follow the text alignment?

_Current: align=left_
image

_Proposal: chev follows alignment rule
image

hughess:
Nice! Looks a lot cleaner, easier to tell which column it applies to.

Agree with your proposal on following the alignment - those bottom screenshots look good

kwongz:
updated chev to follow the alignment of text, similar to the original behavior


https://github.com/user-attachments/assets/2dc00b34-b70b-4f79-8387-993af0199c81


https://github.com/user-attachments/assets/f06079a2-3002-400c-898c-22d14ec9c211

spacing fixes

image

kwongz:
fixes #2711

hughess:
Getting close!

- When the column is right-aligned now, the title has extra padding that's not present in the cells:
   CleanShot 2024-11-18 at 18 48 45@2x

- The chevron looks a bit far from the text in the title:
   CleanShot 2024-11-18 at 18 50 05@2x

- When columns are close together, it's too hard to tell which column the chevron belongs too - need reduced padding/margin on left side of chevron
   CleanShot 2024-11-18 at 18 50 55@2x

Removing the extra space from my first bullet might solve the other 2

kwongz:
conflicts resolved in #2830

PR #2789, opened 2024-11-14
fix: always log errors from handleError

zachstence:


archiewood:
changeset & linting needed

PR #2826, opened 2024-11-20
Upgrade DuckDB WASM to 1.29.0

csjh:
### Description

Allows for dynamic extension usage in the future. Things checked:
- [x] do components that issues queries still work (Dropdown, DateRange, DimensionGrid, ButtonGroup, DataTable)
- [x] Returned data types are the same (or fixed to match before)
    - note: `test_all_types()` doesn't work now because one of the conversions tries to go from bigint to an out of range number
- [x] checked extensions work

### Checklist


e2e/basic/pages/index.md
```diff
@@ -34,7 +34,7 @@ This page can be found in your project at `/pages/index.md`. Make a change to th
       category
   from needful_things.orders
   where category like '${inputs.category.value}'
-  and date_part('year', order_datetime) like '${inputs.year.value}'
+  and cast(date_part('year', order_datetime) as varchar) like '${inputs.year.value}'
```
archiewood, line 37: what's going on here?
are users going to have to update their code if they use date_part or date_trunc?

csjh: Yeah `date_part` returns a number, and doesn't seem like duckdb allows for the implicit conversion from number to string for the comparison with `'${inputs.year.value}'` anymore

csjh: Actually just realized this exists: https://duckdb.org/docs/configuration/pragmas.html#implicit-casting-to-varchar

csjh: don't think there's any user-facing breaking changes after that fix

archiewood:
This autoloads some extensions which is pretty awesome:

tested:
- json
- httpfs

archiewood (approved):

packages/lib/universal-sql/src/client-duckdb/browser.js
```diff
@@ -54,7 +54,7 @@ export async function initDB() {
 		const worker = new DUCKDB_CONFIG.mainWorker();
 
 		// use an intermediate variable to prevent db from being a not-ready database
-		const _db = new AsyncDuckDB(logger, worker);
+		const _db = (window[Symbol.for('duckdb instance')] = new AsyncDuckDB(logger, worker));
```
ItsMeBrianD, line 57: Break into multiple statements, multiple assignment is hard to read

packages/lib/universal-sql/src/client-duckdb/browser.js
```diff
@@ -67,6 +67,11 @@ export async function initDB() {
 			}
 		});
 		connection = await db.connect();
+
+		// revert breaking changes
```
zachstence, line 71: Please elaborate here about what the breaking changes were, between what versions, and how this fixes it

packages/lib/universal-sql/src/client-duckdb/node.js
```diff
@@ -67,6 +67,11 @@ export async function initDB() {
 			}
 		});
 		connection = db.connect();
+
+		// revert breaking changes
```
zachstence, line 71: Please elaborate here about what the breaking changes were, between what versions, and how this fixes it

zachstence (changes requested):

packages/lib/universal-sql/src/client-duckdb/browser.js
```diff
@@ -55,6 +55,8 @@ export async function initDB() {
 
 		// use an intermediate variable to prevent db from being a not-ready database
 		const _db = new AsyncDuckDB(logger, worker);
+		window[Symbol.for('duckdb instance')] = _db;
```
ItsMeBrianD, line 58:
```suggestion
		window[Symbol.for('EVIDENCE_QUERY_ENGINE')] = _db;
```

ItsMeBrianD (approved):
Not to be merged until next week

packages/lib/universal-sql/src/client-duckdb/browser.js
```diff
@@ -55,7 +55,7 @@ export async function initDB() {
 
 		// use an intermediate variable to prevent db from being a not-ready database
 		const _db = new AsyncDuckDB(logger, worker);
-		window[Symbol.for('duckdb instance')] = _db;
+		window[Symbol.for('EVIDENCE_QUERY_ENGINE')] = _db;
```
archiewood, line 58: did this change cause CI to fail

csjh: yeah I have no idea what's up with that

zachstence (approved):

PR #2847, opened 2024-11-23
feat: tailwind configuration works in mono repos

mnahkies:
### Description
Previously the tailwind configuration only worked if the directory structure looked like:
```
/root
├── package.json
├── pages
├── evidence.plugins.yaml
├── ...
├── .evidence
│   └── template
│       ├── tailwind.config.cjs
│       └── node_modules # ./node_modules
│           └── some-module
│               └── dist/**/*.{html,js,svelte,ts,md}  
└── node_modules # ../../node_modules
    └── some-module
        └── dist/**/*.{html,js,svelte,ts,md}  
```

This PR aims to replace the assumptions on possible `node_modules` locations with following the same package resolution algorithm used elsewhere in the sdk, meaning that a hoisted monorepo layout like this will now work:
```
/root
├── package.json
├── node_modules
│   └── some-module # will now work
│       └── dist/**/*.{html,js,svelte,ts,md}  
└── packages
    └── my-evidence-proj
        ├── package.json
        ├── node_modules
        │   └── some-other-module # will still work
        │       └── dist/**/*.{html,js,svelte,ts,md}
        ├── pages
        ├── evidence.plugins.yaml
        ├── ...
        └── .evidence
            ├── template
            └── tailwind.config.cjs
```

There remains an assumption/convention that packages must structure their artifacts into a `/dist/**/*.{html,js,svelte,ts,md}` path, but that seems fine to leave alone.

### Checklist

### DIFF
--- a/packages/lib/sdk/package.json
+++ b/packages/lib/sdk/package.json
@@ -78,6 +78,12 @@
 			"import": "./src/plugins/index.js",
 			"require": null,
 			"types": "./src/plugins/index.js"
+		},
+		"./plugins/discoverPluginPackageRootPathSync": {
+			"default": "./src/plugins/discoverPluginPackageRootPathSync.js",
+			"import": "./src/plugins/discoverPluginPackageRootPathSync.js",
+			"require": "./src/plugins/discoverPluginPackageRootPathSync.js",
+			"types": "./src/plugins/discoverPluginPackageRootPathSync.js"
 		}
 	},
 	"scripts": {
--- a/packages/lib/sdk/src/plugins/discoverPluginPackageRootPathSync.js
+++ b/packages/lib/sdk/src/plugins/discoverPluginPackageRootPathSync.js
@@ -1,0 +1,35 @@
+import path from 'path';
+import { createRequire } from 'module';
+import fs from 'fs';
+import { log } from '../logger/index.js';
+import chalk from 'chalk';
+
+/**
+ *
+ * @param {string} packageName
+ * @returns {string|null}
+ */
+export const discoverPluginPackageRootPathSync = (packageName) => {
+	// note: this will return the directory of the package entrypoint, eg: `some-package/dist/index.js`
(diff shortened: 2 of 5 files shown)


packages/lib/sdk/src/plugins/discoverPluginPackageRootPathSync.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ *
+ * @param {string} packageName
+ * @returns {string|null}
+ */
+export const discoverPluginPackageRootPathSync = (packageName) => {
+	// note: this will return the directory of the package entrypoint, eg: `some-package/dist/index.js`
+	//       so we need to walk up the tree until we find a `package.json` file to discover the root.
+	let pluginPackageDirectory = path.dirname(createRequire(import.meta.url).resolve(packageName));
+	let pluginPackageDirContents = fs.readdirSync(pluginPackageDirectory);
```
mnahkies, line 16: Using `sync` filesystem access, as calling from a synchronous context in the tailwind config

packages/lib/sdk/package.json
```diff
@@ -78,6 +78,12 @@
 			"import": "./src/plugins/index.js",
 			"require": null,
 			"types": "./src/plugins/index.js"
+		},
+		"./plugins/discoverPluginPackageRootPathSync": {
```
mnahkies, line 82: Using a dedicated export, as needs to be required from a `commonjs` context, and `esm` only modules are included in the main `./plugins` export

ItsMeBrianD:
A few notes on this one;

1. I think we can rather easily convert tailwind configuration from `.cjs` to `.mjs`, should help us avoid needing to add another export path to the `sdk` (we have our ideal end state for the package documented in the [monorepo roadmap doc](https://github.com/evidence-dev/evidence/blob/next/MONOREPO_ROADMAP.md#evidence-devsdk)
2. This is not working for me when I run it locally in our monorepo, but the `next` branch does still work - there might be a regression here

PR #2866, opened 2024-11-29
Fix peer dep mismatch

csjh:
### Description

Just tried permutations until something worked, my best guess to why this one works is because peer dependencies are resolved before dev deps + regular deps so the vite-plugin-svelte version we specify takes priority?

### Checklist


csjh:
Seems like doing `npm i @evidence-dev/core-components@latest @evidence-dev/evidence@file:./evidence-dev-evidence-39.2.2.tgz` doesn't upgrade the sveltekit version if you're on the previous evidence version which might be a regression, might have to try more permutations later

ItsMeBrianD (approved):

PR #2870, opened 2024-11-29
Add tests for (real world) package installation

csjh:
### Description

Adds an e2e test that packs every package into tarballs and installs from those using npm


csjh:
Not sure where to put the `node ./build-deps-to-tgz.js && npm i` step so it's in sources for now

csjh:
And need to move away from `import("fs").glob` for the node 18 suite

csjh:
`e2e/packaging` tests are expected to fail until #2866 merges

.github/workflows/e2e.yml
```diff
@@ -65,9 +65,9 @@ jobs:
       - name: Install Playwright and browsers
         run: pnpm playwright install --with-deps
 
-      - name: Run sources
+      - name: Run setup
```
zachstence, line 68: I'd rather not modify all the `e2e` projects to have a `setup` script.

Instead, can you add another step to this workflow to run `setup` if it exists?

```sh
pnpm run --if-present setup
```

package.json
```diff
@@ -66,7 +66,8 @@
 		"unified": "9.1.0",
 		"unist-util-visit": "4.1.2",
 		"uvu": "0.5.2",
-		"vite": "5.4.11"
+		"vite": "5.4.11",
+		"glob": "11.0.0"
```
zachstence, line 70: Install this as a `devDependency` in `e2e/package.json` instead of the root `package.json`

pnpm-workspace.yaml
```diff
@@ -1,5 +1,6 @@
 packages:
   - 'e2e'
   - 'e2e/*'
+  - '!e2e/packaging'
```
zachstence, line 4: Add a comment why this is excluded

zachstence (changes requested):

PR #2864, opened 2024-11-28
Feature/2824 date input

kwongz:
### Description

Created DateInput component that builds off the DateRange structure. Redirects /components/date-range/ --> /components/date-input/

Allows users to access a single date or ranged date input calendar

image

image

### Checklist

- [x] For UI or styling changes, I have added a screenshot or gif showing before & after
- [x] I have added a [changeset](https://github.com/evidence-dev/evidence/blob/main/CONTRIBUTING.md#adding-a-changeset)
- [ ] I have added to the docs where applicable
- [ ] I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable

Issue #2573, opened 2024-09-23
Improve SSR reliability


csjh:
Problem is that, for the most part, the way we use and build components is incompatible with Svelte SSR. More specifically one part - reactive statements (ie `$: `).

For an ideal SSR experience, we should make it an antipattern to have any reactivity (that might happen during SSR; prop reactivity is an exception) happen through reactive statements, and instead use derived stores and event handlers.

Example from `_DataTable`:
```svelte
$: finalColumnOrder = getFinalColumnOrder(
	$props.columns.map((d) => d.id),
	$props.priorityColumns
);
$: orderedColumns = [...$props.columns].sort(
	(a, b) => finalColumnOrder.indexOf(a.id) - finalColumnOrder.indexOf(b.id)
);
```
should become
```svelte
const finalColumnOrder = derived(props, ($props) => getFinalColumnOrder(
	$props.columns.map((d) => d.id),
	$props.priorityColumns
));
const orderedColumns = derived([props, finalColumnOrder], ([$props, $finalColumnOrder]) => [...$props.columns].sort(
	(a, b) => $finalColumnOrder.indexOf(a.id) - $finalColumnOrder.indexOf(b.id)
));
```
or something to that effect

This is especially relevant in components that make use of `<slot />` for composing their behaviour (DataTable is also the best example here, should also be the starting point for changes)

General translation guide:
- `$: var = something;` should become `const var = derived(something, ([$something]) => { ... });`
- `$: if (state) { ... }` should become an event handler for whereever `state` is being changed
- chunky `$: { ... }`s should be broken down into chunks that should fall into one of the above 2 categories

can be expanded upon w/ exception cases when we start work on this

csjh:
Not necessarily directly related, but we should also do a once-over of basic prop reactivity to pick up on things like `ySet` in #2574

PR #2906, opened 2024-12-12
Fix prerendering of DateRange with defaultValue

csjh:
### Description

Conceptually related to #2573

Switches to a callback instead of binding so defaultValues are properly propagated out of `_DateRange`

needs to be refinished after #2864 merges

### Checklist


archiewood:
@csjh any reason this is still a draft?

csjh:
just so it isn't merged before #2864

csjh:
if 2864 isn't ready by wednesday I'll merge this without it though

PR #2913, opened 2024-12-13
Enable Strict Mode for Sources

ItsMeBrianD:
### Description

You can now run `evidence sources --strict` to cause the process to fail when an error is encountered.

### Checklist


package.json
```diff
@@ -86,7 +86,7 @@
 		"dev:test-env": "pnpm --filter ./sites/test-env dev",
 		"build:docs": "pnpm --filter ./sites/docs build",
 		"build:example-project": "pnpm --filter ./sites/example-project build",
-		"build:production": "VITE_BUILD_STRICT=true pnpm --filter ./sites/example-project build",
```
archiewood, line 89: is this a breaking change?

packages/lib/sdk/src/lib/strict.js
```diff
@@ -0,0 +1,19 @@
...
+	process.env.EVIDENCE_STRICT_MODE = 'true';
+};
+
+export const isStrictMode = () => {
+	if (typeof process !== 'undefined') return Boolean(process.env.EVIDENCE_STRICT_MODE);
+	if (typeof import.meta.env !== 'undefined') return Boolean(import.meta.env.EVIDENCE_STRICT_MODE);
+	throw new EvidenceError('Unable to identify if strict mode is enabled', [
+		'process is undefined',
+		'import.meta.env is undefined'
+	]);
```
zachstence, line 18: I feel like a warning log and `return false` would be better than throwing an error

ItsMeBrianD: We shouldn't ever have a situation where this error occurs, and it indicates that something is very wrong (can't access process or vite)

package.json
ItsMeBrianD: Only within the monorepo, everything that is end-user facing lives behind the `build:strict` cli command, which is still supported for now

packages/evidence/cli.js
```diff
@@ -351,10 +351,12 @@ prog
 	.option('--sources', 'only build queries from the specified source directories')
 	.option('--queries', 'only build the specified queries')
 	.option('--debug', 'show debug output')
+	.option('--strict', 'Fail when a source query fails')
```
archiewood, line 354:
```suggestion
	.option('--strict', 'fail when a source query fails')
```

packages/lib/sdk/src/plugins/datasources/cli/sources.js
```diff
@@ -23,35 +26,52 @@ export const sources = {
...
+			},
+			isStrictMode()
+		);
 
-		console.log('  Evaluated sources, saving manifest');
+		log.info('  Evaluated sources, saving manifest');
 
 		await updateManifest(evaluatedManifest, dataDirectory);
-		console.log(chalk.bold.green('  ✅ Done!'));
+		log.info(chalk.bold.green('  ✅ Done!'));
```
archiewood, line 75:
```suggestion
		log.info(chalk.bold.green('  ✔ Sources complete!'));
```

- I feel like CLI logging with emojis is off brand
- "Done" always felt unspecific

archiewood (approved):
Overall this seems to work as intended, nice.

## Logging

I think the logging in the case of a failure could be improved

### Strict

Notice how the error is printed twice

CleanShot 2025-01-02 at 12 24 04@2x

### Without strict

Compare with the error case without strict mode, (error only seen once)

CleanShot 2025-01-02 at 12 26 11@2x

PR #2915, opened 2024-12-13
Third Level Sidebar

archiewood:
### Description

example usage: https://docs-do3q8e9re-evidence.vercel.app/

- [x] z-index seems too high for third level
- [x] alignment of border seems too far left now

https://github.com/user-attachments/assets/a8803e8b-6340-4cf7-a19b-4e9b36e97787

## Outstanding questions
- **Sidebar links** At the moment, it is possible to make the second level links clickable, or have them as unclickable use `sidebar_link=false`, but either way, due to an existing issue with breadcrumbs, those pages are generated anyway. [Example in live docs](https://docs.evidence.dev/core-concepts/) - Should we fix this behaviour in this PR?
- **Collapsable?** Should the second level headers be collapsable `sidebar_collapsable=true` [Example MaterialUI](https://mui.com/material-ui/getting-started/)
  - if so, should they be collapsable by default
  - should collapsable be open or closed by default `sidebar_collapsable_initial=open`, or should only the current page's section be expanded

### Checklist


packages/ui/core-components/src/lib/organisms/layout/sidebar/Sidebar.svelte
```diff
@@ -253,22 +288,59 @@
 								{/if}
 							</span>
 						{/if}
-						{#each file.children as file}
-							{#if file.href && (file.frontMatter?.sidebar_link !== false || file.frontMatter?.sidebar_link === undefined)}
-								{@const active = $page.url.pathname.toUpperCase() === file.href.toUpperCase() + '/'}
+						<!-- check if there is an active state for second level headers -->
```
zachstence, line 291: Is this a TODO?

zachstence (dismissed):

zachstence:
> - Sidebar links At the moment, it is possible to make the second level links clickable, or have them as unclickable use sidebar_link=false, but either way, due to an existing issue with breadcrumbs, those pages are generated anyway. [Example in live docs](https://docs.evidence.dev/core-concepts/) - Should we fix this behaviour in this PR?

No, that should be fixed in a separate PR.

> - Collapsable? Should the second level headers be collapsable sidebar_collapsable=true [Example MaterialUI](https://mui.com/material-ui/getting-started/)

In my opinion, yes they should be collapsible

> if so, should they be collapsable by default

In my opinion, yes they should be collapsible by default

# should collapsable be open or closed by default sidebar_collapsable_initial=open, or should only the current page's section be expanded

In my opinion, they should be open by default.

I view collapsing as a way to hide stuff you're not looking for in a busy sidebar. Have everything visible at the start, but allow the end user to temporarily hide things to make it easier to find what they're looking for

packages/ui/core-components/src/lib/organisms/layout/sidebar/Sidebar.svelte
archiewood: removed

archiewood:
Okay various fixes and changes in here, see docs deploy for example

Done:
- **Slide transitions** between active pages in the border: Note that these do not have a visible effect if you are navigating between pages where the load time > transition time. Try the deploy pages for some fast loading pages to see this effect
- **Removed bolding from second level headers**
- **Hide pages with no children and sidebar_link=false**
- **Add configuration option to EvidenceDefaultLayout to disable third level sidebar** completely (backwards compat)

Not done:
- **Collapsable sections**
  - Users have the flexibility to have all sidebar elements as links - which means side elements would have potentially two roles: nav to link / toggle accordion. 
  - IMO there is not a great tradeoff that has a good UI for this. cf Docusaurus which is bad
  - We can add this in the future if it becomes a request

PR #2916, opened 2024-12-13
Fix virtual source configs

csjh:
### Description

fixes datasources that use virtual/references, adds tests, logs info when source updating errors in debug mode

not totally sure this is the proper fix, still not 100% sure of what `Options.ownKeys` semantically represents (I'm understanding it as the keys of the given options that have materialized values in the schema? Or something like that)

### Checklist

### DIFF
--- a/sites/example-project/src/pages/settings/+page.server.js
+++ b/sites/example-project/src/pages/settings/+page.server.js
@@ -10,6 +10,7 @@ import {
 	getDatasourceConfigAsEnvironmentVariables
 } from '@evidence-dev/sdk/plugins';
 import { isDebug } from '@evidence-dev/sdk/utils';
+import { log } from '@evidence-dev/sdk/logger';
 
 export const load = async () => {
 	if (dev) {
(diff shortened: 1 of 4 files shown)


sites/example-project/src/pages/settings/+page.server.js
```diff
@@ -63,6 +64,10 @@ export const actions = {
 				updatedSource: await writeSourceConfig(opts, source)
 			};
 		} catch (e) {
+			if (isDebug()) {
+				console.error(e);
```
ItsMeBrianD, line 68: Use the logger instead of console here

ItsMeBrianD (changes requested):

sites/example-project/src/pages/settings/+page.server.js
```diff
@@ -63,6 +64,8 @@ export const actions = {
 				updatedSource: await writeSourceConfig(opts, source)
 			};
 		} catch (e) {
+			log.debug(`error updating sources: ${e}`);
+			if (e.stack) log.debug(e.stack);
```
ItsMeBrianD, line 68:
```suggestion
			log.debug(`error updating sources: ${e}`, {error: e});
```

archiewood (approved):

PR #2923, opened 2024-12-16
run mac and windows tests on non-main PRs

csjh:
### Description

Re-adds the windows/mac tests on e2e, mac on non-e2e tests

also removes node-version 18 exclusions because they aren't used anymore (and we also want them to run for the same reason we're re-adding the windows/mac tests)

could also remove node 20 from the next branch exclusion at this point

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -32,13 +32,7 @@ jobs:
           - ${{ github.base_ref == 'main' }}
         exclude:
           - isMain: false
-            node-version: 18
-          - isMain: false
             node-version: 20
-          - isMain: false
-            os: macOS-latest
-          - isMain: false
-            os: windows-latest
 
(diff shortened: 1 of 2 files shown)


csjh:
@zachstence do I have to change anything to make the windows/mac tests optional/fail-friendly, or is that the default when they aren't required?

archiewood:
question: why are we doing this? is it going to make our CI very long?

csjh:
The main goal is to not let stupid windows or mac fails from slipping through during releases and messing everything up

The goal of making the windows/mac tests non-required is that you'll still see broken tests while waiting for a review or something, but when you're just adding a changeset then merging right after on an existing PR you don't have to wait for all the tests to run again (just ubuntu)

Also, if non-required tests aren't fail friendly, then what's the point of them? I kinda just assumed that's what non-required meant

archiewood:
Will merge this. If the CI starts to crush development then we can revisit

archiewood:
I already regret this. Windows proving very flaky

Issue #2918, opened 2024-12-16
[Bug]: query casing is not respected for file queries and aliases


kennan-simpliphy:
### Describe the bug

```
---
queries:
  - releaseYear: release_year.sql
---

vs

---
queries:
  - my_alternative_alias: release_year.sql
---
```

If you alias release_year.sql as releaseYear, the frontmatter will run without error and a query named releaseyear (lowercase 'y') will run successfully. But there will not be a query named releaseYear (capital 'y') in the scope, and any downstream attempts to reference it as such will fail.

While the examples in the docs use snakecase, I don't believe anywhere in the docs indicates that camelCase aliases will cause this particular bug.

### Severity

annoyance

archiewood:
I've looked back into the PR where this feature was implemented.

https://github.com/evidence-dev/evidence/pull/734

We intentionally decided to lowercase all query ids because of unexpected behaviour if you used mixed cases. Solution is therefore

1. Better docs - document this behaviour
2. And ideally - throw useful warning if user tries to use capital letters in a query ID

hughess:
Closing this issue along with our release of the new Evidence Core: https://docs.evidence.dev/

Feel free to open a new issue for this if it is blocking you!

PR #2924, opened 2024-12-17
Docs: snake_case queries

archiewood:
### Checklist

### DIFF
--- a/sites/docs/pages/core-concepts/queries/index.md
+++ b/sites/docs/pages/core-concepts/queries/index.md
@@ -26,6 +26,7 @@ You include SQL queries in your page using a markdown code fence (starting and e
 <Alert status=warning>
 
 **Use Snake Case for Queries**
+
 Use `snake_case` query names, query files and query aliases. Don't use `PascalCase`, `camelCase` or `names with-other/separators` - these may break in unexpected ways.
 
 </Alert>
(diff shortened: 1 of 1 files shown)


sites/docs/pages/core-concepts/queries/index.md
```diff
@@ -23,6 +23,14 @@ When you open a page in dev mode, Evidence runs all of the queries on the page.
 
 You include SQL queries in your page using a markdown code fence (starting and ending with 3 backticks). Evidence requires a query name to be supplied directly after the first 3 backticks.
 
+<Alert status=warning>
+
+**Use Snake Case for Queries**
+Use `snake_case` query names, query files and query aliases. Don't use `PascalCase`, `camelCase` or `names with-other/separators` - these may break in unexpected ways.
```
archiewood, line 29:
```suggestion

Use `snake_case` query names, query files and query aliases. Don't use `PascalCase`, `camelCase` or `names with-other/separators` - these may break in unexpected ways.
```

zachstence:
My preference is `camelCase`, as that fits with the rest of the product (and JS as a whole) the best.

Also, I don't think the docs are correct in stating that `camelCase` or `PascalCase` would "break in unexpected ways". I'm not even sure that special characters would break anything

archiewood:
Sorry I should have linked the issue here: https://github.com/evidence-dev/evidence/issues/2918

zachstence:
> Sorry I should have linked the issue here: #2918
> 
> It specifically breaks when you alias queries with MixedCases - we then give the id mixedcases
> 
> I'm happy to:
> 
> * just add the warning about aliases which are the only part that do not allow other casings
> * not document this, but then we should change the behaviour of the aliasing, which currently lowercases any query ids you supply

If there's a good reason for forcing lowercase then I think this PR is good but we should also add a warning/error log if someone uses caps.

If there's not a good reason for forcing lowercase then I think we should undo that so any casing can be used.

@ItsMeBrianD any input here?

archiewood:
> If there's not a good reason for forcing lowercase then I think we should undo that so any casing can be used.

Let's do this

PR #2927, opened 2024-12-17
Timeseries panel update

kwongz:
### Checklist


kwongz:
@mcrascal 
- Fixed responsiveness for smaller screens
- added Theme styling for darkmode

Lightmode:

image


Dark Mode:

image


Improvements:
- Some of the font colors have changed, such as the bg for the positive and neg nums since we were using green and red 400 and 200, but I could only find the one positive and negative color. So the background red and green are a bit different then the original
image
-Not too familiar with `bg-[radial-gradient(#f9fafb_1px,transparent_1px)]` and adding out `base vars` instead of a hex code (#f9fafb), so still need to fix the radial gradient in dark mode.

kwongz:
Updated Styles

image

image

Added error message and handling + Stories
- InvalidMetrics, Null Data prop, Invalid Query String

Cleaned Store logic, to handle all relevant component data (metrics, filtered data) and errors

finished themes styling

Issue #2673, opened 2024-10-24
[Bug]: Slider lags max value > 10 000


kwongz:
### Describe the bug

Discovered some laggy behaviour w/ slider when step = 1, max >= 10 000. At first i thought it might be the inputs, but when i removed the inputs, the slider still had the same behavior. I'm suspecting it may be shadCN related

### Steps to Reproduce

<Slider
    title="slider" 
    name=slider
    max=10000
/>

also try max=100000 or max = 1000000

I think it froze the entire page for me at 1000000

### Severity

serious, but I can work around it

PR #2940, opened 2024-12-19
Component styling changes for consistency

hughess:
Styling changes to make components more consistent-looking.

## Inputs
Standardized to the following:
- Margin top: none
- Margin bottom: `mb-4`
- Margin right: `mr-2`
 
Before and afters below

### Remaining Issues
- DataTable search bar I don't think is 100% right
- Spacing between input components is tricky with titles involved and when only using bottom margin
   - One approach we could try is using some top margin and removing it when a title is used (letting the title occupy the margin space)
   - When on multiple lines, once a title is included it makes the components look too far apart
- A way to handle groups of inputs (will address in a separate PR with an InputGroup component)
- Tabs styling looks out of place with the rest of components

### Before
CleanShot 2024-12-19 at 13 50 32

### After
CleanShot 2024-12-19 at 13 51 15

### Before
CleanShot 2024-12-19 at 13 53 06@2x

### After
CleanShot 2024-12-19 at 13 53 12@2x

### Before
CleanShot 2024-12-19 at 13 54 17@2x

### After
CleanShot 2024-12-19 at 13 54 23@2x

### Before
CleanShot 2024-12-19 at 13 54 58@2x

### After
CleanShot 2024-12-19 at 13 55 07@2x

## ButtonGroup
ButtonGroup is now `inline-block` rather than `inline-flex` to align well with other input components when side by side

### Before
CleanShot 2024-12-19 at 13 55 44@2x

CleanShot 2024-12-19 at 14 05 24@2x

### After
CleanShot 2024-12-19 at 13 55 56@2x

CleanShot 2024-12-19 at 14 03 22@2x

### Before
CleanShot 2024-12-19 at 13 56 35@2x

### After
CleanShot 2024-12-19 at 13 56 46@2x

### Before
CleanShot 2024-12-19 at 13 57 20@2x

### After
CleanShot 2024-12-19 at 13 57 33@2x

### Before
CleanShot 2024-12-19 at 13 58 20@2x

### After
CleanShot 2024-12-19 at 13 58 32@2x

### Checklist

### DIFF
--- a/.changeset/wise-papayas-clap.md
+++ b/.changeset/wise-papayas-clap.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Styling changes for consistency
--- a/packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/button-group/ButtonGroup.svelte
@@ -102,15 +102,17 @@
 {:else}
 	<HiddenInPrint enabled={hideDuringPrint}>
 		<div
-			class={display === 'tabs' ? '' : 'inline-flex w-fit max-w-full flex-col mt-2 mb-4 ml-0 mr-2'}
+			class={display === 'tabs'
+				? ''
+				: `inline-block overflow-scroll no-scrollbar align-bottom w-fit max-w-full flex-col mb-3 ml-0 mr-2`}
 		>
 			{#if title}
-				<span class="text-sm block mb-1">{title}</span>
+				<span class="text-xs font-medium block mb-0.5">{title}</span>
 			{/if}
 			<div
 				class={display === 'tabs'
 					? 'my-6 flex flex-wrap gap-x-1 gap-y-1'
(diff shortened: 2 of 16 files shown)


kwongz:
I like the idea of standardizing and having a `Design System`. Overall I like all these changes!

- Keeping all the components a standard distance from each other
- Having Titles formatted and positioned in a uniform way  

Some thoughts:

1. The Titles feels off, I think it may be too close to the components? It makes the title and the text inside the component feel like wrong to me for some reason. Maybe its just congested, or just because its zoomed in.

In terms of the tricky-ness, I feel like the components have titles built in separate ways (I know for slider I may have done some funky positioning to make it work), maybe we need a `title component` that can be used to help standardize the inputs?

image

Something like this seems Cleaner

image

2. I thought Dimension Grid had a stronger color the more full the cell was (e.g California was Darker blue then Colorado)? If not maybe we can add it?

image

3. I think Slider could be improved, as it was my first PR and has some funky stuff going on that might make styling it easier/Performance [Issue with Large Scale](https://github.com/evidence-dev/evidence/issues/2673 )

mcrascal (approved):
LGTM

PR #2944, opened 2024-12-19
mute multi-select badges

mcrascal:
Noticed when reviewing #2940 brings the badge styling in multi-selects back to pre-themes. 

Before: 

CleanShot 2024-12-19 at 16 01 14@2x

After: 
CleanShot 2024-12-19 at 16 01 27@2x

### DIFF
--- a/.changeset/lemon-singers-pretend.md
+++ b/.changeset/lemon-singers-pretend.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Bug fix: mute the color of badges in multi-selects.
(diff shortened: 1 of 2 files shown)


hughess (approved):

archiewood:
I assume we did this becuase in dark mode it looks a bit too light now?

CleanShot 2024-12-20 at 12 58 10@2x

ItsMeBrianD:
@mcrascal what's the status on this one

mcrascal:
@ItsMeBrianD should be good to go - changeset incoming. 

@archiewood I think it looks nice. Regardless, if we don't like base-200 in dark mode we should change it globablly.

PR #2945, opened 2024-12-20
Restyle Tabs component

hughess:
### Description
Bring Tabs in line with style of other components

#### Before
CleanShot 2024-12-20 at 08 17 23@2x
CleanShot 2024-12-20 at 08 17 33@2x
CleanShot 2024-12-20 at 08 20 01@2x
CleanShot 2024-12-20 at 08 20 57

#### After
CleanShot 2024-12-20 at 08 17 01@2x
CleanShot 2024-12-20 at 08 17 41@2x
CleanShot 2024-12-20 at 08 20 11@2x
CleanShot 2024-12-20 at 08 21 24

### New Options

`fullWidth` (boolean - default = false)
CleanShot 2024-12-20 at 08 19 04@2x

`background`  (boolean - default = false)
CleanShot 2024-12-20 at 08 19 10@2x

### Checklist


mcrascal (approved):
Looks fantastic. 

I think the prop listing for the `color` needs to be updated to reflect the new default of `primary` rather than `blue`, and to flag that theme tokens will work. 

I think the blue looks good, but I would probably style mine `color=primary-content`, rather than `primary`. That might be a more mellow default to start people with.

Issue #2953, opened 2024-12-23
[Bug]: Cannot use spaces in markdown page names


archiewood:
### Describe the bug

Upgraded from 39.1.13 of evidence in my local development environment and recently tried to upgrade to version 40.0.1.

Markdown pages that contain special characters now break eg `white space.md`

![Image](https://github.com/user-attachments/assets/9488cc64-8fe4-4660-9ac8-428bf70f5db7)

### Steps to Reproduce

Add a page with a space eg `page two.md` and try to navigate to it from the sidebar

https://github.com/user-attachments/assets/b7b4e231-6678-48a1-8d56-efa9cbeddeef

it's detecting the pre encoded space character %20 and then encoding it again, replacing the %-> %25
so overall you get page two.md -> page%20two -> page%2520two

### Logs

```Shell

```

### System Info

```Shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 990.44 MB / 16.00 GB
    Shell: 5.1.16 - /opt/homebrew/bin/bash
  Binaries:
    Node: 22.11.0 - ~/.nvm/versions/node/v22.11.0/bin/node
    npm: 10.9.0 - ~/.nvm/versions/node/v22.11.0/bin/npm
    pnpm: 8.6.9 - /opt/homebrew/bin/pnpm
    bun: 1.0.25 - /opt/homebrew/bin/bun
  Browsers:
    Brave Browser: 123.1.64.122
    Chrome: 131.0.6778.205
    Safari: 17.2.1
  npmPackages:
    @evidence-dev/bigquery: ^2.0.9 => 2.0.9 
    @evidence-dev/core-components: ^5.0.3 => 5.0.3 
    @evidence-dev/csv: ^1.0.14 => 1.0.14 
    @evidence-dev/databricks: ^1.0.8 => 1.0.8 
    @evidence-dev/duckdb: ^1.0.13 => 1.0.13 
    @evidence-dev/evidence: ^40.0.3 => 40.0.3 
    @evidence-dev/motherduck: ^1.0.4 => 1.0.4 
    @evidence-dev/mssql: ^1.1.2 => 1.1.2 
    @evidence-dev/mysql: ^1.1.4 => 1.1.4 
    @evidence-dev/postgres: ^1.0.7 => 1.0.7 
    @evidence-dev/snowflake: ^1.2.2 => 1.2.2 
[...]

PR #2954, opened 2024-12-24
Fix double URL encoding

yecnj:
#### Changelog
- Fixed an issue where URLs were being encoded twice
- Corrected handling of the `src/pages/` path in server configuration

#### Process
In `+server.js`, the URL was being encoded twice due to the following two steps:
- Use of `new URL` to support Windows file paths
- Calculation of `node.href` in the `_buildPageManifest` function.

The following log from `Sidebar.svelte` illustrates the double encoding issue with the fileTree object:
```bash
'white%20space': {
  label: 'white%20space',
  href: '/white%2520space',
  children: {},
  frontMatter: [Object],
  isTemplated: false,
  isPage: true
}
```
image

The `_buildPageManifest` function expects an unencoded `pagePath` for the computation of variables such as `part` and `label`.

To rectify this, I replaced the use of `new URL` with `path.normalize` and `replace` to maintain Windows compatibility.

This change also led me to adjust how paths starting with `/src/pages` are handled, changing them to start with `src/pages` instead.

Here are the updated results logged from the `Sidebar.svelte` file showing the corrected handling:

```bash
'white space': {
  label: 'white space',
  href: '/white%20space',
  children: {},
  frontMatter: [Object],
  isTemplated: false,
  isPage: true
}
```

image

This resolves issue #2953 .

#### Additional Comment
When I first started reading at the code, I thought `example-project` was a test repository unrelated to deployment.

However, I noticed that it's actually a "template" being copied to `.evidence/template`, based on various PRs such as #1737 and #2529. Have I understood this correctly?

If so, to avoid confusion, how about you rename `example-project` to something like `project-template`?

### Checklist

### DIFF
--- a/sites/example-project/src/pages/api/pagesManifest.json/+server.js
+++ b/sites/example-project/src/pages/api/pagesManifest.json/+server.js
@@ -27,7 +27,7 @@ export function _buildPageManifest(pages) {
 		isPage: false
 	};
 	for (const [pagePath, pageContent] of Object.entries(pages)) {
-		const path = pagePath.replace('/src/pages/', '');
+		const path = pagePath.replace('src/pages/', '');
 		let node = fileTree;
 		for (const part of path.split('/')) {
 			if (part === '+page.md') {
@@ -71,7 +71,8 @@ export async function GET() {
 				const relative_path = path.join(dirent.parentPath ?? dirent.path ?? dir, dirent.name);
 				const content = await fs.readFile(relative_path, 'utf-8');
 				// regularize for windows
-				pages[new URL(`file:///${relative_path}`).pathname] = content;
+				const normalized_path = path.normalize(relative_path).replace('\\', '/');
+				pages[normalized_path] = content;
 			} else if (dirent.isDirectory()) {
 				await recursiveReadDir(path.join(dir, dirent.name));
 			}
--- a/sites/example-project/src/pages/api/pagesManifest.json/+server.js
+++ b/sites/example-project/src/pages/api/pagesManifest.json/+server.js
@@ -27,7 +27,7 @@ export function _buildPageManifest(pages) {
 		isPage: false
 	};
 	for (const [pagePath, pageContent] of Object.entries(pages)) {
-		const path = pagePath.replace('src/pages/', '');
+		const path = pagePath.replace('/src/pages/', '');
 		let node = fileTree;
 		for (const part of path.split('/')) {
 			if (part === '+page.md') {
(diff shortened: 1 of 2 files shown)


ItsMeBrianD (approved):

archiewood:
This appears to fail the e2e tests on windows

sites/example-project/src/pages/api/pagesManifest.json/+server.js
```diff
@@ -71,7 +71,8 @@ export async function GET() {
 				const relative_path = path.join(dirent.parentPath ?? dirent.path ?? dir, dirent.name);
 				const content = await fs.readFile(relative_path, 'utf-8');
 				// regularize for windows
-				pages[new URL(`file:///${relative_path}`).pathname] = content;
```
archiewood, line 74: Note the comment above this removed line
```
// regularize for windows
```

yecnj:
Thanks for catching this! I'll test this change to verify it resolves the issue.

yecnj:
As mentioned in the PR description, using `new URL` causes double encoding issues.
I've modified the code to achieve similar functionality using `path.sep` and `join`, and reverted the path processing method back to its original form as the current approach was prone to issues.

I'll wait for workflow approval to proceed with the deployment.

archiewood:
@ItsMeBrianD can you kick CI again please

archiewood:
@yecnj - It seems this latest is causing the docs deploy to fail. The docs use Evidence, so this makes it a decent chance that this has broken some core Evidence functionality. 

If you are developing, running the docs `pnpm run sources:docs && pnpm run dev:docs` is a decent way to check if your changes are working

Error message on netlify:

https://app.netlify.com/sites/next-docs-evidence/deploys/677b31ce721c750008d7a5f5

archiewood (changes requested):
appears to be causing docs to fail to build

yecnj:
@archiewood Thanks for the thorough review! While I couldn't reproduce any issues in local builds, I've added the forward slash prefix to ensure consistent path handling.

 I've also checked that the Netlify build is now passing successfully.

https://app.netlify.com/sites/next-docs-evidence/deploys/677c8f22f610230008e921dd

yecnj:
hmm. I will check windows test.

archiewood:
it might just be a flaky test, let me retry it

yecnj:
I think something is pending now.
I tried log-in to chromatic but didn't get much results.
can you do me a favor?

archiewood:
Looks good in CI. @ItsMeBrianD any final thought or good to merge ?

archiewood (approved):

archiewood:
@yecnj thank you for your contribution - this will go out during our next release

PR #2967, opened 2025-01-03
Component alignment and title consistency

hughess:
### Description
Updates spacing across the component library to make them more consistent and aligned when placed side by side.

Adds title and subtitle for components to match the title and subtitles available in charts.

### New Title and Subtitle Props
CleanShot 2025-01-02 at 18 49 21@2x
CleanShot 2025-01-02 at 18 48 20@2x
CleanShot 2025-01-02 at 18 47 37@2x

### Spacing and Alignment
CleanShot 2025-01-03 at 11 08 06@2x
CleanShot 2025-01-03 at 11 07 56@2x
CleanShot 2025-01-03 at 11 06 57@2x

### Checklist

### DIFF
--- a/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
+++ b/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
@@ -7,6 +7,10 @@
 	/** @type {import("@evidence-dev/sdk/usql).Query | unknown}*/
 	export let data;
 
+	export let height = 200;
+
+	export let skeletonMt = undefined;
+
 	$: if (Query.isQuery(data)) {
 		data.fetch(); // Somebody wants this to load. Without this the query builder features don't work
 		unsub();
@@ -36,8 +40,8 @@
 	{/if}
 {:else if !_data || (!_data.dataLoaded && !_data.error)}
 	<slot name="skeleton" loaded={_data}>
-		<div class="w-full h-64">
-			<Skeleton />
+		<div class="w-full" style="height: {height}px">
+			<Skeleton mt={skeletonMt} />
 		</div>
 	</slot>
 {:else if _data.error && $$slots.error}
--- a/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
+++ b/packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
@@ -9,7 +9,7 @@
 
 	export let height = 200;
 
-	export let skeletonMt = undefined;
+	export let skeletonClass = undefined;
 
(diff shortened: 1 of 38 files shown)


archiewood:
Outstanding check: why do data tables now show titles in chromatic

hughess:
@archiewood it looks like there were title props in those stories before but they had no effect since we didn't have a title prop in the DataTable

archiewood:
Okay excellent

packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
```diff
@@ -36,8 +40,8 @@
 	{/if}
 {:else if !_data || (!_data.dataLoaded && !_data.error)}
 	<slot name="skeleton" loaded={_data}>
-		<div class="w-full h-64">
-			<Skeleton />
+		<div class="w-full" style="height: {height}px">
```
archiewood, line 43: any reason we cant use tailwind here `h-[{height}px]`

packages/ui/core-components/src/lib/atoms/skeletons/Skeleton.svelte
```diff
@@ -4,9 +4,11 @@
 
 <script>
 	import { fade } from 'svelte/transition';
+
+	export let mt = 2;
```
archiewood, line 8: feels like we could create a utility prop here

```js
export let class

<div class="{class} animate-pulse h-full w-full mt-2 mb-4" in:fade|local>
```

sites/docs/pages/components/data/data-table/index.md
```diff
@@ -15,7 +15,7 @@ limit 100
 
 <DocTab>
     <div slot='preview'>
-        <DataTable data={orders_summary}/>
+        <DataTable data={orders_summary} title="Title" subtitle="Subtitle"/>
```
archiewood, line 18: Not sure the first docs example should have a title for a table

archiewood (approved):
Couple of suggestions

packages/ui/core-components/src/lib/atoms/query-load/QueryLoad.svelte
hughess: Tailwind has issues with dynamic values, I believe since the class isn't loaded ahead of time

packages/ui/core-components/src/lib/atoms/skeletons/Skeleton.svelte
hughess: good call - we can use the `cn` function and the `class` prop pattern from Accordion

PR #2981, opened 2025-01-07
DateRange last day selection fix

kwongz:
### Description

Currently, added an extra day to `max range` to account for when `start date` and `end date` are on the end date range, causing the `rangeCalendar` to deselect the date. The extra day is disabled, to the user everything looks normal, behind the scenes we are handling what is most likely a timezone issue

Before:

https://github.com/user-attachments/assets/3f7f538b-6843-48b2-9568-c073cb21291f

After:

https://github.com/user-attachments/assets/f0154a15-d0b8-449e-afbe-21975997d630

I attempted to test the time zones in London, SF and Toronto, the last date deselection issue persisted.

Attempted to add times to possible dates' start and end range, resulting in weird behavior. Would it best to add a new library to handle dates with times? currently, date formmater is only for dates without times.

Tried to look into changing selection dates, start selection at 00h:00m:00s and end selection  23h:59m:59s, but couldn't find a way in range calendar bits UI to handle that.

### Checklist


ItsMeBrianD:
Need to test this with data that is hourly instead of daily to make sure we aren't overdisplaying

kwongz:
image

Currently, the Max date is Wed Jun 19 2024 20:00:00 GMT-0400 (Eastern Daylight Time), which goes into Thursday June 20th UTC time. 

With the 'extraDay - date disabled' fix the daterange does not over display

ItsMeBrianD (approved):

PR #2998, opened 2025-01-10
Expand clickable area of level three sidebar links

mcrascal:
### Description

Make level three links in the sidebar click-able for the full width of the sidebar, rather than just the width of the text. Aligns the behaviour to level two links.


archiewood (approved):
Good catch, thank you

PR #3016, opened 2025-01-16
BigQuery: Enable Access to Google Sheets-Backed Tables

olegam:
Explicitly specify scopes for service user authentication in BigQuery datasource and include Drive scope, to fix issue when querying tables backed by a Google Sheet document

### Description

BigQuery supports tables backed by a Google Sheet. In the current version of Evidence it's not possible to query those tables using Service User authentication. 
```bash
$ npm run sources
✔ Loading plugins & sources
-----
  [Processing] mysouce
  config_orders ✖ Error: Unknown Error
  ```
  This error is not very descriptive, but looking in the Google Cloud Console logs reveal the error:
  ```
  Access Denied: BigQuery BigQuery: Permission denied while getting Drive credentials.
  ```
  
To fix this we need to include the `https://www.googleapis.com/auth/drive` scope when authenticating.

After adding these lines I'm able to load data from my Google Sheet-backed BigQuery table.


zachstence:
If I'm understanding correctly, this would add these scopes for everybody using the BigQuery datasource, correct? If that's the case, we probably want this to be configurable by the user in their `connection.yaml`.

@ItsMeBrianD thoughts?

archiewood:
@zachstence I think this is a good approach.
So we'd want a field to allow people to add scopes, perhaps?

archiewood:
Okay @olegam - my suggestion here: 
Add a boolean checkbox to the options which a user can check to add the scopes

title: "Enable Connected Sheets"
description "Allows access to connected Google Sheets. Requires additional scopes: x,y,z"

relevant section starts

`module.exports.options = {`

olegam:
Ok I added a checkmark in the configuration options to toggle the additional Drive scope as suggested by @archiewood

archiewood:
I think if you merge back in next now, CI will pass

zachstence (approved):

PR #3022, opened 2025-01-17
Add ignoreZoom option to maps

hughess:
### Description
Adds an `ignoreZoom` prop to the following map components to allow control over map layers auto-adjusting zoom and panning:

- AreaMap
- PointMap
- BubbleMap
- Areas
- Points
- Bubbles

### Before - zoom includes all layers
CleanShot 2025-01-22 at 12 37 56@2x

### After - `ignoreZoom=true` on Areas layer
CleanShot 2025-01-22 at 12 38 17@2x

### Checklist


packages/ui/core-components/src/lib/unsorted/viz/map/PointMap.svelte
```diff
@@ -67,6 +67,9 @@
 	/** @type {boolean} */
 	export let legend = true;
 
+	/** @type {boolean} */
+	export let ignoreZoom = false;
```
zachstence, line 71: Should we use `toBoolean` here too?

packages/ui/core-components/src/lib/unsorted/viz/map/AreaMap.svelte
```diff
@@ -62,6 +62,9 @@
 	/** @type {boolean} */
 	export let legend = true;
 
+	/** @type {boolean} */
+	export let ignoreZoom = false;
```
zachstence, line 66: Should we use `toBoolean` here too?

zachstence (changes requested):
Needs `ignoreZoom` prop handling added to `BubbleMap` and `Bubbles`.

Needs docs additions for `AreaMap`, `PointMap`, and `BubbleMap`.

Needs stories with a `play` function illustrate what `ignoreZoom` does and test it with Chromatic snapshots.

packages/ui/core-components/src/lib/unsorted/viz/map/PointMap.svelte
hughess: toBoolean is included in `Points` to handle it there. Does that work?

zachstence: Yes, as long as the `ignoreZoom` prop isn't used anywhere inside `PointMap`

zachstence (approved):

PR #3047, opened 2025-01-27
Update redirect file

archiewood:
### Description

Some of our redirects appear to be failing on the docs site eg

https://docs.evidence.dev/deployment/evidence-cloud/

However others are working eg

https://docs.evidence.dev/markdown

~~I assume whitespace is the cuprit~~

The alternative option is that it is trailing slashes

### Checklist


zachstence:
> Some of our redirects appear to be failing on the docs site eg
> https://docs.evidence.dev/evidence-cloud

@archiewood The whitespace shouldn't have any affect. There isn't a redirect for `/evidence-cloud`, that's why that particular one isn't working.

archiewood:
Apologies, I added the wrong link: https://docs.evidence.dev/deployment/evidence-cloud/

archiewood:
okay so @zachstence I think you are right that it is not whitespace. I think it must be trailing slash

https://docs.evidence.dev/components/area-chart/
https://docs.evidence.dev/components/area-chart

zachstence:
@archiewood Regex might be a good solution here

For example:
```json
{
  "redirects": [
    {
      "source": "^/a/?$",
      "destination": "/b",
      "permanent": true
    }
  ]
}
```

The `source` regex matches any URL that begins with (`^`) `/a` and optionally ends with (`?$)` `/`, and redirects to the destination `/b`.

So for https://docs.evidence.dev/components/area-chart it would look something like:
```json
{
  "redirects": [
    {
      "source": "^/components/area-chart/?$",
      "destination": "/components/charts/area-chart",
      "permanent": true
    }
  ]
}
```

hughess (approved):

PR #3064, opened 2025-02-02
Fix mobile usage issue for Info component

hughess:
Fixes issue with Info component where on mobile, hover to show tooltip did not work. Now clicks will work in addition to hover.

### DIFF
--- a/packages/ui/core-components/src/lib/unsorted/ui/Info.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/ui/Info.svelte
@@ -9,27 +9,33 @@
 	import chroma from 'chroma-js';
 	import HoverCard from '../../atoms/hover-card/HoverCard.svelte';
 
-	/** @type {import('@steeze-ui/svelte-icon').IconSource} */
 	let icon = InfoCircled;
-
 	export let description = '';
-
 	export let size = 4;
-
 	export let className = undefined;
-
 	const { resolveColor } = getThemeStores();
 	export let color = 'base-content-muted';
+
 	$: colorStore = resolveColor(color);
 	$: textColor = chroma($colorStore).css();
+
+	// State for manual toggle
+	let isOpen = false;
+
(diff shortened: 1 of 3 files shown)


packages/ui/core-components/src/lib/unsorted/ui/Info.svelte
```diff
@@ -9,27 +9,33 @@
...
+
+	function toggleOpen() {
+		isOpen = !isOpen;
+	}
 </script>
 
-<HoverCard align="start" side="right" alignOffset={-8} sideOffset={4}>
+<HoverCard open={isOpen} align="start" side="right" alignOffset={-8} sideOffset={4}>
+	<!-- svelte-ignore a11y-click-events-have-key-events -->
+	<!-- svelte-ignore a11y-no-noninteractive-element-interactions -->
```
ItsMeBrianD, line 32: We shouldn't ignore these accessibility issues, lets resolve them before we merge

ItsMeBrianD (changes requested):

packages/ui/core-components/src/lib/unsorted/ui/Info.svelte
```diff
@@ -9,32 +9,41 @@
...
 
-<HoverCard align="start" side="right" alignOffset={-8} sideOffset={4}>
+<HoverCard open={isOpen} align="start" side="right" alignOffset={-8} sideOffset={4}>
 	<span
 		slot="trigger"
 		style:--textColor={textColor}
-		class="inline-block align-middle pb-0.5 pr-1 leading-4 w-fit {className}"
-		role="tooltip"
+		class="inline-block align-middle pb-0.5 pr-1 leading-4 w-fit {className} cursor-pointer"
+		role="button"
```
ItsMeBrianD, line 35: This works but we could also change the span to a button tag

ItsMeBrianD (approved):

Issue #2450, opened 2024-08-17
[Bug]: Month to Date and Year to Date in Date Range Selector returns wrong range


martinmesz:
### Describe the bug

As per definition Month to Date should return a range from the first day of the month up until now, not including today. 
Similarly for Year to Date, the range should go from the first day of the year up until now, not including today. 
Instead it selects the end date as the last day of the month or year. This will return wrong data in case there is already some records for today (or the future for some reason, like predictions)

### Steps to Reproduce

1. Create a date range selector as below:
```
<DateRange
    name=date_range
/>
```
2. Write a query that returns the selected values as strings
```
sql test
  select
    '${inputs.date_range.start}' as start,
    '${inputs.date_range.end}' as end
```
3. Examine the returned values

### System Info

```Shell
System:
    OS: macOS 13.6.4
    CPU: (4) x64 Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
    Memory: 102.20 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.16.0 - /usr/local/bin/node
    npm: 10.8.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 127.0.6533.120
    Safari: 17.3
  npmPackages:
    @evidence-dev/bigquery: ^2.0.7 => 2.0.7 
    @evidence-dev/core-components: ^4.7.2 => 4.7.2 
    @evidence-dev/csv: ^1.0.12 => 1.0.12 
    @evidence-dev/databricks: ^1.0.7 => 1.0.7 
    @evidence-dev/duckdb: ^1.0.11 => 1.0.11 
    @evidence-dev/evidence: ^39.1.1 => 39.1.1 
    @evidence-dev/motherduck: ^1.0.2 => 1.0.2 
    @evidence-dev/mssql: ^1.0.9 => 1.0.9 
    @evidence-dev/mysql: ^1.1.3 => 1.1.3 
    @evidence-dev/postgres: ^1.0.6 => 1.0.6 
    @evidence-dev/snowflake: ^1.1.0 => 1.1.0 
    @evidence-dev/sqlite: ^2.0.6 => 2.0.6 
    @evidence-dev/trino: ^1.0.7 => 1.0.7
```

### Severity

annoyance

### Additional Information, or Workarounds

Garlando:
I've opened a PR with a fix: https://github.com/evidence-dev/evidence/pull/3113

PR #3113, opened 2025-03-07
Date range preset amendment

Garlando:
### Description

New options to the Date Range UI components presets:
- `Month to Today`
- `Year to Today`

The `... to Date` presets sets the end date to the end of the Month and Year, respectively
These new presets will end the date range on the supplied end date
image
image

Raised in Bug - https://github.com/evidence-dev/evidence/issues/2450  

### Checklist

- [ ] ~I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable~

### DIFF
--- a/sites/docs/queries/orders_by_day.sql
+++ b/sites/docs/queries/orders_by_day.sql
@@ -4,6 +4,7 @@ select
     sum(sales) as sales,
     sum(sales) / count(*) as aov
 from needful_things.orders
-where order_datetime > '2019-01-02'
+where order_datetime > '2019-01-02' 
+  and order_datetime <= '2021-12-20'
 group by 1
 order by 1
--- a/.changeset/rotten-wasps-poke.md
+++ b/.changeset/rotten-wasps-poke.md
@@ -1,0 +1,7 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Update Date Range presets with better defaults
+
+Month to Date & Year to Date - now both end on last available date of the data used
--- a/.changeset/rotten-wasps-poke.md
+++ b/.changeset/rotten-wasps-poke.md
@@ -2,6 +2,6 @@
 '@evidence-dev/core-components': patch
 ---
 
-Update Date Range presets with better defaults
+Update Date Range presets with new options
 
-Month to Date & Year to Date - now both end on last available date of the data used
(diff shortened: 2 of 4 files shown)


archiewood:
While I agree with the original premise of the bug, this is worth considering.

This would be a breaking change - it would change the behaviour of people already using Month to Date / Year to Date if they have newer data

Is there an alternative option where we add new presets with the desired ranges ?
eg Month To Date Today, Year To Date Today
Or some better names?

Garlando:
@archiewood 
A fair point, although I wonder how many people are expecting the proposed behaviour

Would it be overkill to have all the following options, keeping the existing but adding new '... to today' and '... to yesterday' options
image

If that sounds ok I can make some updates to the docs too.

Garlando:
As an aside, the daterange component is somewhat a victim of its own flexibility, as it uses whatever end date you give it as "today", I could try and make that a bit clearer in the docs if you like

archiewood:
Feels a bit overkill honestly! 
These options would appear for all users by default - would just the `to Today` options be enough for your use case?

Alternatively, we could add all these and change which ones were loaded by default - I think that would be the ideal outcome, but would be a slightly larger change.

Garlando:
Yeah just the to Today options would suit what I'm trying to do, aswell as I think some other mentions i've seen in Slack. there were just some conflicting definitions when i did a more general search for the terms

a larger non-default set of choices does sound good eventually, but happy to go small for now. 
And if you think this is a lot you should see the options looker studio gives 😆

sites/docs/queries/orders_by_day.sql
```diff
@@ -4,6 +4,7 @@ select
     sum(sales) as sales,
     sum(sales) / count(*) as aov
 from needful_things.orders
-where order_datetime > '2019-01-02'
+where order_datetime > '2019-01-02' 
+  and order_datetime <= '2021-12-20'
```
Garlando, line 8: I've changed this to give some extra distinction between the date range presets, with the data ending on 2021-12-31 previously it would have given the same output for the `... to Date` options as `... to Today`
It's usage in other docs pages shouldnt be affected in any material way

Garlando:
@archiewood discussed changes applied

archiewood (approved):

archiewood:
@zachstence can you kick CI

PR #3184, opened 2025-08-05
Bump duckdb to 1.3.2

archiewood:
### Description

Bumps DuckDB to 1.3.2

### Checklist


archiewood:
Seems to be a new dependency on GLIBC that is causing all these tests to fail - we may need to switch to @duckdb/node-api

PR #3244, opened 2025-11-12
Date input range fixes

arefabouhamdan:
### Description
This PR addresses several issues with the date input component, including timezone handling, default values, year selection, date range presets, and improved date boundary management. 

#### 1. **Timezone Fix** 
- **Issue**: Date conversion was using `toISOString()` which could cause timezone-related date shifts
- **Fix**: Changed `dateToYYYYMMDD()` in `helpers.js` to use local date components instead of ISO string conversion
- **Impact**: Prevents dates from shifting by one day due to timezone conversion issues

#### 2. **Default Value Fix**
- **Issue**: Default value wasn't consistently set to today
- **Fix**: Changed default value to use `today(getLocalTimeZone())` as the reference date
- **Impact**: Date inputs now default to today's date, providing a better user experience

#### 3. **Extra Day End String Fix**
- **Issue**: `extraDayEndString` was only calculated for range inputs, causing issues with single date inputs
- **Fix**: Removed the `range` condition check so `extraDayEndString` is calculated for all inputs
- **Impact**: Ensures proper date boundary handling for both single date and range inputs, and enables date disabling for dates outside the data range

#### 4. **Data-Driven Year Selection & Date Disabling**
- **Feature**: Year dropdown now only shows years that exist in the provided data, and dates outside the data range are disabled
- **Implementation**: 
  - Added `data` and `dates` props to extract available years from query data
  - Year options are filtered to only show years present in the dataset
  - Calendar uses `calendarStart` and `calendarEnd` (with `extraDayEndString` for safety) to set `minValue` and `maxValue`, which automatically disables dates outside the range
  - Falls back to default behavior if no data is provided
- **Impact**: Users can only select years and dates that have data, preventing invalid date selections and improving UX

#### 5. **New Date Range Presets**
- **Feature**: Added "Today" and "Yesterday" presets for quick date selection

#### 6. **Improved Preset Calculations**
- **Feature**: Presets now use a `referenceDate` that selects the appropriate reference point
- **Implementation**:
  - For single date inputs: uses the selected date
  - For range inputs: uses the end date of the selected range
  - "Month to Today" and "Year to Today" presets now correctly use `todayDate` instead of `calendarEnd`

#### 7. **Enhanced Year Selection Logic**
- **Feature**: Improved year selection to handle month boundaries when switching years
- **Implementation**:
  - When changing years, attempts to keep the same month if valid
  - If the same month is invalid in the new year, finds the closest valid month
- **Impact**: Smoother user experience when navigating between years with different available date ranges

### Checklist

- [X] For UI or styling changes, I have added a screenshot or gif showing before & after
- [X] I have added to the docs where applicable
- [X] I have added to the [VS Code extension](https://github.com/evidence-dev/evidence-vscode) where applicable

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211182961543171


hughess:
@arefabouhamdan do you know why the default dates of the date input have changed? Below are current live docs and the preview docs from this PR:

### Current Docs
CleanShot 2025-12-01 at 20 56 34@2x

### This PR
CleanShot 2025-12-01 at 20 56 22@2x

arefabouhamdan:
@hughess We wanted the DateInput and DateRange components default to today's date (or the nearest available date in the data range) when no explicit defaultValue is provided. This ensures the component always shows the latest available data on initial load, rather than leaving the selection empty. If today falls outside the data range, it intelligently falls back to the closest boundary date (the end or start of the dataset), so the user always sees the most recent relevant data.

arefabouhamdan:
@hughess and regarding the timezone issue it was always off by 1 day in my timezone (EET), after the changes we tested it on multiple timezones and it was accurate on all of them

PR #3245, opened 2025-11-12
Update to latest DuckDB packages

ebardelli:
### Description

The DuckDB packages have been updated to their latest versions:
 - @duckdb/node-api to version 1.4.2
 - @duckdb/duckdb-wasm to version 1.31.1-dev1.0

duckdb-async has been replaced with duckdb/node-api. This led to some changes in how queries
are run and how results are processed. The new duckdb/node-api package offers improved performance
and follows the same release schedule as the main DuckDB project. duckdb-async will be discontinued in
about 6 months.

Additional tests have been added to ensure that semicolons inside block comments do not split SQL 
statements incorrectly. 

The new packages work on my production site. There might be remaining bugs with the duckdb/node-api implementation, but I haven't run into them yet.

The duckdb-wasm update is a drop-in replacement for the old one.

### Checklist

### DIFF
--- a/.changeset/witty-bobcats-care.md
+++ b/.changeset/witty-bobcats-care.md
@@ -1,12 +1,9 @@
 ---
 '@evidence-dev/duckdb': major
 '@evidence-dev/universal-sql': major
-'@evidence-dev/mssql': minor
 '@evidence-dev/db-commons': minor
 '@evidence-dev/core-components': minor
 ---
-
-Update DuckDB to latest packages
 
 Update DuckDB to latest packages:
 
@@ -18,4 +15,3 @@ This release also has small data fixes across several packages:
 - Better handling of NULL values when discovering column types
 - Fix batch processing of parquet files
 - Fix error with temporary parquet files when reloading data in dev environment
-- Better handing of error messages for the MSSQL driver in dev environment


packages/datasources/duckdb/package.json
```diff
@@ -10,7 +10,7 @@
 	},
 	"dependencies": {
 		"@evidence-dev/db-commons": "workspace:*",
-		"duckdb-async": "1.1.3"
+		"@duckdb/node-api": "^1.4.1-r.5"
```
archiewood: is there a stable version we can pin to?

ebardelli: Yes. Let's give duckdb a couple of days to release the new packages. 1.4.2 was released yesterday (if I remember correctly).

ebardelli:
I pushed a new commit that uses the stable releases as of today.

* @duckdb/node-api uses version 1.4.2-r.1. I think they are doing <major duckdb version>.patch now?
* @duckdb/duckdb-wasm to version 1.30.0. This uses an older version of duckdb, but that's what it's tagged as current on npmjs, even if they released 1.31.0 without a dev tag. 🤷

archiewood:
looks like you need to regenerate the lockfile

> version 1.30.0. This uses an older version of duckdb

How old we talking? WASM sometimes lags duckdb

> I think that the risk is kind of flipped.

I guess I was thinking more about the risk to the product as a whole. Impacting the duckdb connector vs impacting the runtime. I take your point around the complexity

adam-homeboost:
thank you for this!

PR #3310, opened 2026-07-21
fix: preserve BarChart axis and colors when switching to stacked100

rajanpanth:
## Description

Fix BarChart updates when the `type` prop changes dynamically between `stacked` and `stacked100`.

Previously, `BarChart` derived its `stacked100` normalization flag only during component initialization. A later `type` update reached the bar series but left the core chart using the previous normalization mode. Entering `stacked100` therefore retained raw values, while leaving it retained normalized fractions, producing incorrect axis semantics and an apparent series-color collapse even though the series identities and palette remained intact.

## Changes

- derive `stacked100` reactively from `type`
- add regression coverage for forward, reverse, repeated, and initial `stacked100` rendering
- add a patch changeset for `@evidence-dev/core-components`

## Testing

- `pnpm --filter ./packages/ui/core-components test:unit --run src/lib/unsorted/viz/bar/BarChart.spec.js` — 4 passed
- `pnpm --filter ./packages/ui/core-components test:unit --run` — 121 passed, 3 pre-existing skipped
- focused ESLint and Prettier checks for all changed files — passed
- `pnpm --filter ./packages/ui/core-components build` — passed, including `svelte-package` and `publint`
- `pnpm changeset status --since=origin/main` — passed
- `git diff --check` — passed

The package-wide Svelte check remains blocked by existing diagnostics in unchanged files. The aggregate lint command is also blocked by the current Prettier/YAML baseline; every changed file passes focused formatting and ESLint validation.

## Before

Dynamic transitions could leave the chart data normalization and axis formatting in the previous stacked mode.

## After

Normalized data, raw data, axis formatting, series identity, and distinct palette colors consistently match the active chart type across forward, reverse, and repeated transitions.

## Screenshots

No screenshot is included. The existing Storybook environment fails before rendering BarChart stories with a pre-existing `fromStorybook` runtime error; the behavior is covered deterministically at the generated ECharts configuration boundary.

## Related issue

Fixes #3299

## Compatibility

No public API change. Initial rendering and unrelated chart types remain unchanged. The runtime addition is one reactive boolean comparison when `type` changes.

PR #3311, opened 2026-07-24
fix: make prop-derived chart values reactive (Area label position, Sparkline series type)

rajanpanth:
Sibling fix to #3310 — same bug class, found by sweeping the other viz components for it.

## Problem

Two chart components derive values from props with plain `let` declarations, freezing them at mount-time while their consumers are reactive:

- `Area.svelte`: `let defaultLabelPosition = swapXY ? 'right' : 'top'` — toggling `swapXY` after mount leaves the reactive `labelPosition` fallback stale (labels fall back to `top` on a swapped chart instead of `right`).
- `_Sparkline.svelte`: `let seriesType = type === 'area' ? 'line' : type` — `seriesType` is consumed inside the `$: try` config block, so mounting as `bar` and switching to `line` keeps rendering bars.

## Fix

Derive both with `$:` so the reactive consumers see current values — the same one-line pattern as the `stacked100` fix in #3310. Changeset included (patch, `@evidence-dev/core-components`).

Deliberately excluded from the sweep: the `isInitial`/`queryID` mount-time captures in `core/*` (intentionally non-reactive — they record initial state), and function-local `let`s.

Note on tests: #3310 establishes the regression-spec pattern for this class on BarChart; these two are the same one-line derivation change. Happy to add per-component specs if you want them — Area mounts inside chart context, so I kept this PR surgical.

### DIFF
--- a/.changeset/reactive-prop-derivations.md
+++ b/.changeset/reactive-prop-derivations.md
@@ -1,0 +1,5 @@
+---
+'@evidence-dev/core-components': patch
+---
+
+Make prop-derived chart values reactive: Area's default label position now follows `swapXY`, and Sparkline's series type follows `type`, instead of freezing at their mount-time values.
--- a/.changeset/reactive-prop-derivations.md
+++ b/.changeset/reactive-prop-derivations.md
@@ -2,4 +2,4 @@
 '@evidence-dev/core-components': patch
 ---
 
-Make prop-derived chart values reactive: Area's default label position now follows `swapXY`, and Sparkline's series type follows `type`, instead of freezing at their mount-time values.
+Make prop-derived chart values reactive: Area's and Line's default label positions now follow `swapXY`, and Sparkline's series type follows `type`, instead of freezing at their mount-time values.
--- a/packages/ui/core-components/src/lib/unsorted/viz/area/Area.svelte
+++ b/packages/ui/core-components/src/lib/unsorted/viz/area/Area.svelte
@@ -117,7 +117,7 @@
 		middle: 'inside'
 	};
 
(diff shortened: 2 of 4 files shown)


rajanpanth:
Pushed a follow-up commit: an untruncated re-sweep of the viz tree found the identical pattern in `Line.svelte` (`defaultLabelPosition` derived from `swapXY` with a plain `let`, consumed by the reactive `labelPosition` fallback) — now fixed the same way, and the changeset text updated. That completes the class: the only remaining prop-derived `let`s (`_Chart`'s `ySet`/`xSet`) are intentionally non-reactive per their in-code comment and were left alone.

rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to rebase, update tests, or make any requested changes.

Thanks!

rajanpanth:
The external-contributor test workflow is still waiting for maintainer approval. Could a maintainer approve the workflow run when convenient? The deploy previews and visible checks are otherwise complete.

rajanpanth:
Closing this: the restructure superseded it.

The three files this PR touched (`viz/area/Area.svelte`, `viz/line/Line.svelte`, `viz/core/_Sparkline.svelte`) were removed with the rest of `packages/ui`. The charts now live under `core/src/user-components/tags/series_charts/` and were rewritten with Svelte 5 runes; `defaultLabelPosition` does not appear anywhere under `core/`.

The bug class itself, a plain `let` derived from a prop and then read by a `$:` statement, does not survive the move to runes.

The `505885dad publish evidence core` restructure replaced `packages/` with `cli/`, `core/` and `docs/`, so this branch no longer applies and there is nothing left to rebase onto.

PR #3318, opened 2026-08-20
test(sdk): await rejects assertions in loadSourceConfig

rajanpanth:
Three tests in `loadSourceConfig.spec.js` assert on a rejected promise without awaiting it:

```js
expect(loadConnection('./sources/invalidYaml')).rejects.toThrowError(
  'Error parsing connection.yaml file'
);
```

The assertion is never awaited, so it does not run before the test ends. All three tests are named "should throw …" and none of them can currently fail. Vitest already flags it when the suite runs:

> Promise returned by `expect(actual).rejects.toThrowError(expected)` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3.

So this is also forward-breaking, not just dead coverage.

The neighbouring tests in the same file already use `await` correctly (`expect(await loadConnectionOptions(...)).toEqual(...)`), so this looks like an oversight rather than intent.

**Verification.** All 13 tests in the file pass with the awaits added, and the three Vitest warnings are gone. To confirm the assertions are now actually live, I changed one expected message to a string that cannot match and re-ran: that test failed, as it should. Before the change, the same mutation still passed. Reverted afterwards.

I scanned the rest of the repo for the same pattern and this file was the only occurrence. `prettier --check` passes on the changed file.

### DIFF
--- a/packages/lib/sdk/src/plugins/datasources/loadSourceConfig.spec.js
+++ b/packages/lib/sdk/src/plugins/datasources/loadSourceConfig.spec.js
@@ -63,7 +63,7 @@ describe('loadSourceOptions', () => {
 		expect(await loadConnectionOptions('./sources/options')).toEqual({ x: 5 });
 	});
 	it('should throw if an invalid base64 error occurs', async () => {
-		expect(loadConnectionOptions('./sources/invalidB64')).rejects.toThrowError(
+		await expect(loadConnectionOptions('./sources/invalidB64')).rejects.toThrowError(
 			'Error parsing connection.options.yaml file'
 		);
 	});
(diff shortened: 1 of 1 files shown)


rajanpanth:
Closing this: the restructure superseded it.

`packages/lib/sdk/src/plugins/datasources/loadSourceConfig.spec.js` was removed, and there is no `plugins/datasources` directory in the new layout, so the unawaited `rejects` assertions this PR fixed no longer exist.

The `505885dad publish evidence core` restructure replaced `packages/` with `cli/`, `core/` and `docs/`, so this branch no longer applies and there is nothing left to rebase onto.

PR #3319, opened 2026-08-20
fix: honor DateInput defaultValue in single-date mode

### LINKED ISSUE #3312: [Bug]: `DateInput` `defaultValue` is silently ignored in non-range mode
### Describe the bug

`<DateInput>` accepts a `defaultValue` prop (same name/shape as the working one on `<Dropdown>`), but in non-range (single-date) mode it has no effect at all - the prop is declared and passed through, but nothing in the component ever uses it to set the initial selection. Instead, the initial selected date always falls back to the earliest date in the calendar's range (`start`), never `end`, `defaultValue`, or anything resembling "today"/"most recent".

Concretely: `updateDateRange()` sets `selectedDateInput = start` unconditionally for non-range mode. The only place `defaultValue` is consumed is `applyPreset()`, which only assigns `selectedDateInput` when `range` is `true` - so in non-range mode, calling it (which does happen) has no visible effect. Silent no-op, no warning, no error.

Two related, smaller limitations that make this hard to work around (included here since they're the same root cause - `defaultValue`/initial-value handling - rather than filing three overlapping issues):

- In range mode, `defaultValue` *does* work, but only via the built-in named presets (`Last 7/30/90/365 Days`, `Last 3/6/12 Months`, `Last Month`, `Last Year`, `Month/Year to Date`, `Month/Year to Today`, `All Time`). There's no single-day preset ("Today"/"Yesterday"), and `presetRanges` can only *filter* that existing list by label - it can't define a new one.
- Passing literal `start`/`end` props does correctly set the initial selection (e.g. `start="2026-07-29" end="2026-07-29"` opens with exactly that day selected) - but `calendarStart`/`calendarEnd` (derived from those same props) are also used as the calendar's `minValue`/`maxValue`. So this "fix" disables every other day in the calendar and the prev/next month arrows - the picker stops being usable for picking any other date. There's no prop to set the initial value independently of the navigable min/max bounds.

### Steps to Reproduce

1. Non-range mode, `defaultValue` has no effect:
   ```svelte
   <DateInput data={available_days} dates="day" name="day" title="Datum" defaultValue="2026-07-29" />
   ```
   Result: opens with the *earliest* date in `available_days` selected (e.g. `2025-12-28`), never `2026-07-29`.

2. Range mode, `defaultValue` only accepts preset names, not dates:
   ```svelte
   <DateInput data={available_days} dates="day" name="day" title="Datum" range defaultValue="Yesterday" />
   ```
   Result: no effect - `"Yesterday"` doesn't match any built-in preset label.

3. Literal `start`/`end` default correctly but disable all other days:
   ```svelte
   <DateInput data={available_days} dates="day" name="day" title="Datum" range start="2026-07-29" end="2026-07-29" />
   ```
   Result: opens correctly on `Jul 29, 2026 - Jul 29, 2026`, but opening the calendar shows every other day grayed out/unclickable, and the prev/next month arrows are disabled.

### Logs

```Shell
No console errors, no warnings, no failed network requests in any of the three cases - confirmed via browser console. The prop is simply never read (case 1) or there's no matching preset (case 2), so it fails completely silently.
```

### System Info

```Shell
- `evidence-dev/evidence`: 40.1.8
- `@evidence-dev/core-components`: 5.4.2
- Node: v24.18.0
- OS (container): Alpine Linux 3.24
- Reproduced in both `evidence dev` (SvelteKit dev server) and a static `evidence build` output
```

### Severity

serious but I can work around it

Annoyance / serious but I can work around it - not blocking. We're using `<Dropdown>` instead for this exact "default to a specific day, let the user pick any other day" use case; it works correctly (its `defaultValue` does apply), just with a worse search UX (fuzzy/subsequence matching instead of prefix matching on typed dates).

Detailed workaround description: `<Dropdown data={available_days} value=day defaultValue="...">` instead of `<DateInput>`, re-stamping the literal `defaultValue` string on every rebuild (since neither component supports a query-derived default - documented separately as expected behavior, not this bug). Trade-off: no calendar-click UI, and Dropdown's search box does fuzzy matching rather than prefix matching, so typing e.g. `2026-01` doesn't reliably show only January 2026 dates first.

What would actually solve this cleanly, roughly in order of how much they'd help:
- Make `defaultValue` in non-range mode accept a literal `YYYY-MM-DD` string and actually apply it (parity with `Dropdown`).
- Let range-mode `defaultValue`/`presetRanges` accept a literal date or `{start, end}` pair, not just named presets.
- Decouple the calendar's `minValue`/`maxValue` from the initial `selectedDateInput` - e.g. a separate `initialValue` prop, so `start`/`end`/`data`+`dates` can define the full pickable range while `initialValue` controls what's selected on mount.

Henry369-0:
## Summary
- honor an ISO defaultValue for DateInput in single-date mode
- keep the initialized value from being overwritten by the fallback range initialization
- add a Storybook example covering the single-date default value

## Validation
- verified the branch against evidence-dev/evidence:main
- reviewed the GitHub compare diff for the two intended files
- local test suite was not run because the Windows Git HTTPS credential layer could not clone the repository

Closes #3312

### DIFF
--- a/packages/ui/core-components/src/lib/atoms/inputs/date-input/DateInput.stories.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/date-input/DateInput.stories.svelte
@@ -65,3 +65,6 @@
 	<h1>Data has a query error</h1>
 	<DateInput name="QueryError" data={queryErrorData} />
 </Story>
+<Story name="Single Date Default Value" let:args>
+	<DateInput {...args} name="dateInput_default" defaultValue="2024-12-29" />
+</Story>
--- a/packages/ui/core-components/src/lib/atoms/inputs/date-input/_DateInput.svelte
+++ b/packages/ui/core-components/src/lib/atoms/inputs/date-input/_DateInput.svelte
@@ -239,7 +239,12 @@
 	}
 
 	function updateDateRange(start, end) {
-		if (selectedPreset) return;
+		if (selectedPreset || selectedDateInput) return;
+
+		if (!range && typeof defaultValue === 'string' && /^\d{4}-\d{2}-\d{2}$/.test(defaultValue)) {
+			selectedDateInput = YYYYMMDDToCalendar(defaultValue);
+			return;
+		}
 
 		if (range) {
 			selectedDateInput = { start, end };


Henry369-0:
Closing this draft of the contribution while the contributor profile privacy settings are being reviewed. No maintainer action is required.

